PR #1, opened 2015-12-30
Fix capitalization and add better window manager detection.

ix:
Summary of the WM madness going on here:

If the user has not exported the environment var `windowmanager` (in which case there is no need to overwrite it temporarily, it's already defined), then the script checks if `wmctrl` is installed. If not, it checks if `.xinitrc` exists. If it does, it greps the exec line. Otherwise, it gives up.


dylanaraps:
Awesome!

Looks good to me, thanks for the pr!

PR #2, opened 2015-12-30
A few nice additions

ix:
namely:

if mpc is not installed, do not try to detect mpc

if w3mimgdisplay is not installed, do not attempt to run it

and proper crux support as /etc/os-release doesn't exist on crux (at least for me)

### DIFF
--- a/fetch.sh
+++ b/fetch.sh
@@ -114,7 +114,11 @@ title="$(whoami)@$(hostname)"
 
 # Operating System (Configurable with "-O" and "--distro" at launch)
 # You can manually set this if the command below doesn't work for you.
-os=$(awk '/^NAME=/' /etc/*ease | sed -n 's/^NAME=//p' | tr -d '"')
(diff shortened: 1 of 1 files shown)


dylanaraps:
Thanks for the pull request. 

I'll merge this once I push the changes below later today.

```
- Added flag to disable bold text

- Added flag to disable text wrapping

- Moved to using only printf instead of a mix of echo/printf

- The script now aligns the cursor to the bottom of the
  terminal instead of using a hardcoded amount of newlines

- Image size is now dynamic, the only difference on the user
  side is that instead of setting a long line of spaces as 
  padding, you set your font width. This means that you set the 
  var once and the script will just work at any window size.

- You can disable dynamic images and go back to the old
  behaviour with a flag and in script var.

- tput is now a dependency because we need (tput cols/lines/cup)
  for the dynamic images.
```

dylanaraps:
I've added the changes and modified them slightly. Thanks for the pr!

PR #3, opened 2015-12-31
Some more info lines

fullsalvo:
Added the ability to display X Terminal fonts through .XResources/.XDefaults (if they exist) and to check for screen resolution.

Things that need to be addressed are case sensitivity in file names and the possible dependency in xdpyinfo for checking screen resolution. I've done some work in anticipation of the file name issue, but I feel there's a better way than my method.

### DIFF
--- a/fetch.sh
+++ b/fetch.sh
@@ -1,31 +1,476 @@
-#!/usr/bin/env bash
+#!/bin/bash
 # Fetch info about your system
 #
 # Optional Dependencies: (You'll lose these features without them)
 #   Displaying Images: w3m
 #   Image Cropping: ImageMagick
-#   Wallpaper Display: feh Window Manager Detection: wmctrl
+#   Wallpaper Display: feh
 #   Current Song: mpc
 #   Text formatting, dynamic image size and padding: tput
 #
 # Created by Dylan Araps
 # https://github.com/dylanaraps/dotfiles
 
-
-# Info Prefixes {{{
-# The titles that come before the info (Ram:, Cpu:, Uptime:)
-# TODO: Add an easy way to specify these at launch.
+# Speed up script by not using unicode
+export LC_ALL=C
 
 
-title_os="OS"
-title_kernel="Kernel"
-title_uptime="Uptime"
-title_packages="Packages"
-title_shell="Shell"
-title_windowmanager="Window Manager"
-title_cpu="CPU"
-title_memory="Memory"
-title_song="Song"
+# Config Options {{{
+
+
+# Info Options {{{
+
+# Info
+# What to display and in what order.
+# Format is: "Subtitle: function name"
+# Additional lines you can use include:
+# "underline" "linebreak" "echo: msg here" "title: title here"
+# You can also include your own lines by using:
+# "echo: subtitlehere: $(custom cmd here)"
+info=(
+    "gettitle"
+    "underline"
+    "OS: getos"
+    "Kernel: getkernel"
+    "Uptime: getuptime"
+    "Packages: getpackages"
+    "Shell: getshell"
+    "Window Manager: getwindowmanager"
+    "CPU: getcpu"
+    "Memory: getmemory"
+    "Song: getsong"
+    "linebreak"
+    "getcols"
+    "linebreak"
+)
+
+# CPU
+
+# CPU speed type
+# --speed_type current/min/max
+speed_type="max"
+
+
+# Color Blocks
+
+# Color block range
+# --block_range start end
+start=0
+end=7
+
+# Toggle color blocks
+# --color_blocks on/off
+color_blocks="on"
+
+# Color block width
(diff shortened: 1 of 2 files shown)


dylanaraps:
Thanks for the pull request. 

I'm unsure about a few of the changes, I'll write up some comments in a sec.

fetch.sh
```diff
@@ -183,6 +185,24 @@ getwindowmanager () {
     fi
 }
 
+
+# Font (Configurable with "-F" and "--font" at launch)
```
dylanaraps: I don't know if I want to add this as it only works with terminals that use Xresources/Xdefaults. We might be able to support more terminals by detecting terminal emulator and then checking default config locations for fonts.

I'll work on it.

fetch.sh
```diff
@@ -183,6 +185,24 @@ getwindowmanager () {
...
+elif [ -f $HOME/.XDefaults ]; then
+    font=$(cat $HOME/.XDefaults | grep -v ! | awk '/font/ {print $2}' | sed 's/xft://;s/:pixelsize.*$//')
+elif [ -f $HOME/.Xdefaults ]; then
+    font=$(cat $HOME/.Xdefaults | grep -v ! | awk '/font/ {print $2}' | sed 's/xft://;s/:pixelsize.*$//')
+else
+    font="Unknown"
+fi
+
+# Screen Resolution (Configurable with "-R" and "--resolution" at launch)
+resolution=$(xdpyinfo | awk '/dimensions:/ {print $2}')
```
dylanaraps: This should be an if statement that also checks for xrandr as they can both be used to get screen resolution and the user may only have one installed.

You can check using "if type" like so:

    if type -p xrandr >/dev/null 2>&1; then

fetch.sh
```diff
@@ -200,7 +220,9 @@ cpuspeed () {
 
 # Memory (Configurable with "-M" and "--memory" at launch)
 # Print the total amount of ram and amount of ram in use
-memory=$(free -m | awk '/Mem:/ {printf $3 "MB / " $2 "MB"}')
+usedmem=$(free -m | awk '/cache:/ {print $3}')
```
dylanaraps: This calls free and awk twice, I don't think the readability benefit is worth it.

fullsalvo: This change wasn't for the readability at all. It provides entirely different information. The method the script originally used includes cached memory. It's somewhat misleading of how much memory is actually available. For me, the difference is enormous.
15-dec-30_201124

dylanaraps: My bad.

Also I just compared the output from free on my pc and it differs to yours which means we'll have to some sort of detection or get the memory a different way.

This line doesn't output anything on my pc because of above.

    usedmem=$(free -m | awk '/cache:/ {print $3}')

fetch.sh
dylanaraps: This info is also highly innacurate for multi monitor setups. (It shows every monitor stitched together).

We should also think of an easy way for the user to hide specific info that they don't want to see. Something like this could work: 

```sh
    fetch.sh --hide "Packages Memory, Song"
```
We could also have some optional info lines that need to be enabled with a ```--show``` flag similar to above. This way we can include stuff like Xresources fonts etc.

fullsalvo: If users would like these details hidden by default, it would be simpler to have a list at the top of the file listing all possible options then a variable containing what the user would like to display. For example:

    valid=( img os kernel uptime pkgs shell wm cpu mem song)
    display=( img os shell wm song )

to be used in conjunction with show or hide flags, to save the user effort overall.

dylanaraps:
I tracked down which version of free changed the layout and added version detection, it should hopefully work now.

fullsalvo:
My free version is

    free from procps-ng 3.3.9
which makes the conditional statement you have for memory still display wrong.

    if [ ${freeversion//.} -gt 330 ] || [${freeversion//.} -gt 3300 ]; then
This should be an easy fix, though. I'm just not sure of the version where it switched, so I can't fix it myself.

dylanaraps:
I see.

procps-ng changed their version numbering last year(?) and the old versions numbers may overlap with the new version numbers in the future.

Screenfetch is using /proc/meminfo to get memory info, lets have a look at how they're using it and see if we can come up with something that works for everyone.

dylanaraps:
Before I push to master, how does this look?

```sh
getmemory () {
        freemem=$(awk '/MemFree:/ {printf $2}' /proc/meminfo)
        totalmem=$(awk '/MemTotal:/ {printf $2}' /proc/meminfo)
        availmem=$(awk '/MemAvailable:/ {printf $2}' /proc/meminfo)
        usedmem=$((totalmem - availmem))
        memory="$(( ${usedmem%% *} / 1024))MB / $(( ${totalmem%% *} / 1024))MB"
}
```

fullsalvo:
It works fine on my end.

PR #4, opened 2015-12-31
Add support for Void Linux

sapake:
A simple addition to add support for counting packages under Void Linux, by using xbps, Voids package manager.


dylanaraps:
Awesome, thanks!

PR #17, opened 2016-01-17
added osx wallpaper finding

### LINKED ISSUE #16: Blank Output on OSX
Like literally nothing. 

It just gets stuck like this:

 this

No output in regular terminal.app or iterm, when run from bash or fish. This is on OSX beta, but I doubt that would make a difference.

Goerofmuns:
this fixes a bug where on OSX the script couldn't find wallpaper.

fixes #16

### DIFF
--- a/fetch
+++ b/fetch
@@ -19,6 +19,15 @@ export LC_ALL=C
 export LANG=C
 export LANGUAGE=C
 
+# osascript source for wallpaper detection on OSX
+wallsource="
+on run
+    tell application \"Finder\"
+        select window of desktop
+        return (get desktop picture as text)
+    end tell
+end run"
+
 
 # Config Options {{{
 
@@ -736,6 +745,12 @@ getwallpaper () {
     case "$os" in
         "Linux")
             img="$(awk -F\' '/feh/ {printf $2}' $HOME/.fehbg)"
+        ;;
+
+        "Mac OS X")
+            img="$(osascript -e "$wallsource" |
+                   sed 's/:/\//g' |
+                   sed 's/.*Users/\/Users/g')"
         ;;
 
         "Windows")


dylanaraps:
Thanks for the pull request!

It'd be better if we moved the wallsource variable to the wallpaper function itself. We can also use variable substitution instead of sed as it's faster since we're not spawning any externel processes.

What does this command below output? 

```sh
osascript -e 'tell app "finder" to get posix path of (get desktop picture as text)')
```

Goerofmuns:
That does give the correct output, a path to my wallpaper. I can update the fork and PR, or you can just do it yourself.

Also looking back at the commit I see what you mean, this is why I shouldn't multitask. :)

dylanaraps:
Awesome! 

I'll add it myself. One more quick question, what does the output look like? I want to see whether or not we'll need to quote the path.

Thanks again for the pr!

dylanaraps:
I've pushed it to master, let me know if it works or not.

Goerofmuns:
Just remove the double quote escaping backslashes in `\"Finder\"` so its `"Finder"`. osascript doesn't seem to like them.

dylanaraps:
Wait a sec.

dylanaraps:
I'm trying to quote the path but the command already uses both types of quotes. Are you able to paste the exact output from the command? I want to see whether or not the path output escapes spaces etc.

Goerofmuns:
>9:10: syntax error: Expected expression, property or key form, etc. but found unknown token. (-2741)

That's what `osascript` gives when I run fetch.

`osascript -e 'tell app "finder" to get posix path of (get desktop picture as text)'` this runs fine but like you said it's that extra pair of quotes.

dylanaraps:
Are you able to give me the output of the working command?

Goerofmuns:
`/Users/tom/wall/vla.jpg` aka my wallpaper path.

dylanaraps:
I've removed the quotes and pushed it to master, it should work now. I'm not sure about what will happen if the path has spaces in it though.

Thanks for the PR!

Goerofmuns:
All works well! Nice work.

dylanaraps:
I just pushed another commit, try it out again please!

Goerofmuns:
still works!

dylanaraps:
Awesome!

Thanks for all the help!

PR #24, opened 2016-01-21
Removal of gtk-info if not being used

LostInInaka:
If GTK is not installed or being used, the information should not be printed, however if it is installed then of course it should be printed.

This pull will not print gtktheme or gtkicons if no gtk theme is installed

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ about your system and prints it to the terminal next to an image of your choice!
 https://github.com/dylanaraps/fetch/wiki/Customizing-Info
 
 
-![1](http://i.imgur.com/bORHxT5.png)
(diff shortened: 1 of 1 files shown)


dylanaraps:
I'm not sure if I want this merged, the default info printed is meant to be a base. I expect you to edit it to suit your needs.

dylanaraps:
Here's another way of doing it:

```sh
getgtktheme
if [ "$gtktheme" != "None" ]; then
    prin "GTK Theme $gtktheme"
    info "Icons" gtkicons
fi
```

This doesn't require any modification to the functions below and will work if the file exists and the name field is blank inside of the file.

Issue #30, opened 2016-01-27
TODO


dylanaraps:
#### Battery
- [ ] Multi-battery support for Windows and BSD.
- [ ] State support for multiple batteries.

#### Disk
- [ ] OpenBSD and NetBSD support.
- [ ] Print disk models

#### Testing

- [ ] Test neofetch on Qubes OS.

#### OS

- [ ] [ Puppy Linux ] Package count.

kelaun:
Hi,

For the battery, the information is stored in "/sys/class/power_supply/BAT1/capacity" if you are using only one battery. Otherwise you have multiple files to query.

Hope this helps!

dylanaraps:
Thanks!

What's the output of this?:

```sh
cat /sys/class/power_supply/BAT1/capacity
```

kelaun:
It outputs the percentage of remaining battery, but only the number without the '%' (so like 89 for instance).

dylanaraps:
Awesome!

Do you have multiple batteries? If so this command should print all of the battery info at once:

```sh
cat /sys/class/power_supply/BAT*/capacity
```

kelaun:
I only have one battery but indeed it should output all the battery percentages.

dylanaraps:
Is it possible to open the file in an editor? I want to know if it's just the battery value on line 1 like so:

```sh
1 80
2
3
```

kelaun:
There is only one line in the file which contains the value.

dylanaraps:
Awesome!

I'll get to work on this now.

dylanaraps:
This is what the output will look like for a single battery:

```sh
Battery: 80%
```

How should we format the output for multiple batteries?

```sh
Battery: 1: 25%, 2: 50%, 3: 80%: 4. 20%

Battery: 25%, 50%, 80%, 20%
```

I'm not sure about how to go about this, prefixing each percentage with `BATX 20%` etc makes the line way too long.

kelaun:
Can't you just create a new line for every battery that is present?
I would think that not that many laptops use more than one battery.

tudurom:
It's not possible because of the config file structure. A solution would be enumerating them in a line.

dylanaraps:
It **is** possible to multiline them, it'll just be some extra work.

dylanaraps:
Here's the TODO for the battery function:

- Add support for all platforms
- Add an option to only print the first battery.
- Add an option to print all batteries on a single line.

[...]

PR #44, opened 2016-02-01
Implementation for battery level, working with multiple batteries

tudurom:


tudurom:
Working on a multi battery one.

tudurom:
```
^
|
```

dylanaraps:
I'm already working on it locally and I'm almost done!

tudurom:
A multi line one?

dylanaraps:
Yup!

dylanaraps:
I'm not sure if the battery path is correct as I'm working with dummy files
on my pc. This prints each battery on a seperate line and numbers the title
accordingly.

Here's what I've got so far, test it out!:

```sh
getbattery () {
    case "$os" in
        "Linux")
            batteries=($(cat /sys/class/power_supply/BAT*/capacity))

            # If there's only a single battery,
            # don't number the subtitle.
            if [ "${#batteries[@]}" == 1 ]; then
                prin "Battery: ${batteries[0]}%"
                return
            fi

            # Print each battery on a seperate line.
            index=1
            for bat in ${batteries[@]}; do
                prin "Battery${index}: ${bat}%"
                index=$((index + 1))
            done
        ;;
    esac
}
```

tudurom:
The "Battery" text should be styled as in the config file.

tudurom:
Yeah, the path seems correct. I have done my implementation on a laptop.

dylanaraps:
Yup!

I'm working on letting the user style it the same as the other titles.

tudurom:
Btw on some models the first battery is `BAT0` while on others it's called `BAT1`, but I don't think it makes a difference.

dylanaraps:
It doesn't make a difference with the way I've set it up!

dylanaraps:
Also changing the title for the batteries works now.

kelaun:
If the first battery is 0 and there are multiple batteries, should the index not be 0 instead of 1 at the starting point?

tudurom:
I have 1 battery and it's called 0. No problem.

dylanaraps:
I'll change the index to `0`, you're right.

tudurom:
I realised I just said a very dumb thing.

kelaun:
Shouldn't you first check if BAT0 exists? Otherwise the index should indeed be 1.

dylanaraps:
Yup, the function has no error handlng at the moment.

dylanaraps:
It's in master now, please test it out.

dylanaraps:
I'm closing this PR, lets move discussion to **#30**.

tudurom:
Works.

dylanaraps:
Awesome!

Thanks for your PR too!

tudurom:
No problem.

PR #54, opened 2016-02-08
Rewrite GTK theme detection to also get correct DE theme.

dylanaraps:
This rewrite of the theme detection fixes cases where the gtk theme is 
incorrectly displayed due to the DE not writing `.gtkrc` files etc.

I've opened this PR to see if anyone has any ideas/queries about this before 
it gets pushed to master.

This is pretty much done, all that's missing is KDE support  and 
some error detection which I'm working on as we speak.

### DIFF
--- a/fetch
+++ b/fetch
@@ -1144,7 +1144,7 @@ getgtk () {
 
         "Xfce"*)
             if type -p xfconf-query >/dev/null 2>&1; then
-                gtk2Theme=$(xfconf-query -c xsettings -p /Net/$xfconf)
+                gtk2theme=$(xfconf-query -c xsettings -p /Net/$xfconf)
             fi
         ;;
 
--- a/fetch
+++ b/fetch
@@ -1153,8 +1153,8 @@ getgtk () {
         ;;
     esac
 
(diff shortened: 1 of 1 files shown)


Wyn10:
> all that's missing is KDE support

Images are not showing properly in Konsole at all :cry:.  Image starts in the menubar and even with that off it won't output the image if there's text on the same line.  I can't even get a picture of the bug because the image disappears when Konsole becomes an inactive window.  Works fine in UXTerm though.

dylanaraps:
@Wyn10, do images in ranger work fine in Konsole?

dylanaraps:
@Wyn10, I just installed Konsole and I don't think images will work with it. (Images act the same in ranger)

The only way I can see us fixing this is if we add an option to draw the image in a loop. (This is how ranger manages to at least display it.)

Wyn10:
@dylanaraps 

Looks like it, it opens the image with feh.

Here's a screenshot.  Ranger through konsole on left, Ranger through Xterm on right, and what fetch shows through konsole on the bottom-right.

Alt text

dylanaraps:
You need to enable inline images inside ranger.

https://github.com/hut/ranger/wiki/Image-Previews

PR #60, opened 2016-02-09
IP Address support

dylanaraps:
IP support is done in my eyes.

I've tested this PR with these OS and it works fine without issues.

- 8 linux distros
- 3 BSD distros
- Windows 7 

Before I push this to master I'd you guys to test this out!

You can test it by swapping to the `ip` branch and adding these 
lines to your printinfo function:

```sh
info "Local IP" localip
info "Public IP" publicip
```

If you guys have any suggestions for features/changes now is 
the time to speak up.

### DIFF
--- a/README.md
+++ b/README.md
@@ -239,6 +239,7 @@ alias fetch2="fetch \
     --shell_version on/off      Enable/Disable showing \$SHELL version
     --battery_num num           Which battery to display, default value is 'all'
     --battery_shorthand on/off  Whether or not each battery gets its own line/title
+    --ip_host url               Url to ping for public IP
     --birthday_shorthand on/off Shorten the output of birthday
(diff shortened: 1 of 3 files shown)


mclado:
Works like charm on OS X!

dylanaraps:
The flag `--ip_host` lets you change the website we ping for the ip address. 
Not every website will work with the flag, only those that return the ip through 
curl as plaintext.

The point of the flag/config opyion is to allow the user to easily change the host if 
the website goes down or is slow to respond.

Here's a list of sites that work:

- http://ipinfo.io/ip
- http://ipecho.net/plain
- http://icanhazip.com
- http://ident.me

mclado:
Aaa I get it. Works like it should.

PR #63, opened 2016-02-10
Added option to dynamically title the Window Manager output.

dylanaraps:
I'm not sure about whether ot not to add this to master. This PR adds a feature that titles
the output of `windowmanager` dynamically based on whether or not the output was a
WM or DE.

The feature can be used like so:

```sh
printinfo () {

# Dynamic title enabled
info windowmanager

# Use the title specified here
info "Window Manager" windowmanager

}
```

For example:

```sh
# The output is 'openbox' so the title 'WM' is used.
WM: Openbox

# The output is 'XFCE' so the title 'DE' is used.
DE: XFCE

```

This option will be enabled by default and removes the hardcoded-ness
of the original title.

**ALSO**: We need to rename the `windowmanager` function to something broader
as it now supports outputting the `Desktop Environment` too.

What do you guys think about adding this to master?

### DIFF
--- a/README.md
+++ b/README.md
@@ -237,6 +237,7 @@ alias fetch2="fetch \
     --gtk3 on/off               Enable/Disable gtk3 theme/icons output
     --shell_path on/off         Enable/Disable showing \$SHELL path
     --shell_version on/off      Enable/Disable showing \$SHELL version
+    --wm_dynamic_title on/off   Enable/Disable dynamically titling the DE/WM info cmd
     --battery_num num           Which battery to display, default value is 'all'
     --battery_shorthand on/off  Whether or not each battery gets its own line/title
     --ip_host url               Url to ping for public IP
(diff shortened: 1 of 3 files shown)


tudurom:
Why not leave it to the user? If you, the user, want to display your DE, then you can make a change in the config file. Simple.

dylanaraps:
The user can still set the title if they like, this just provides a saner default. I'm seeing screenshots
of people using fetch and the output includes stuff like this: `Window Manager: XFCE` which is wrong.

What we could do instead is provide a seperate function that gets the DE so the default printinfo
function would look like this:

```sh
printinfo () {

# We should rename 'desktopenvironment' and 'windowmanager' to 'DE' and 'WM'
info "Desktop Environment" desktopenvironment 
info "Window Manager" windowmanager

}
```

The output would look like this:

```sh
# If the user has no DE, don't display anything.
Window Manager: Openbox

# If the user has a DE, display it with the WM.
Desktop Environment: XFCE
Window Manager: Openbox
```

What do you think about this approach?

tudurom:
Yes.

dylanaraps:
Alright, I'll implement that then.

Thanks.

PR #68, opened 2016-02-13
Fixed uptime output when the uptime is under a minute

tudurom:
Before this fix, if the uptime was under a minute, it showed as `Uptime: up`. You don't say?

### DIFF
--- a/fetch
+++ b/fetch
@@ -527,7 +527,7 @@ getuptime () {
 
                 *)
                     uptime="$(uptime -p)"
-                    [ "$uptime" == "up " ] && uptime="up $(cat /proc/uptime)"
+                    [ "$uptime" == "up " ] && uptime="up $(cat /proc/uptime | cut -d'.' -f1) seconds"
                 ;;
             esac
         ;;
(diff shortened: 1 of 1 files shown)


tudurom:
You can see that this fix is pretty partial.

dylanaraps:
Thanks!

A few notes though:

- You don't need to use cat and cut, you can use a single awk.

```sh
awk -F '.' '{print $1}' /proc/uptime
```

- Add these to `uptime_shorthand` case statement under the `minute` substitutions: 

```sh
"on")
    # ...
    uptime=${uptime/seconds/secs}
    # ...
;;

"tiny")
    # ...
    uptime=${uptime/ seconds/s}
    # ...
;;
```

tudurom:
Will do it tomorrow, it's the middle of the night here. Thanks for the
tips! I'm still learning shell scripting.

dylanaraps:
No problem! :)

tudurom:
Done. It works brilliantly. You can merge now.

dylanaraps:
Thanks! :)

PR #71, opened 2016-02-14
Check for packages based on which package manager is installed.

dylanaraps:
This PR changes the way we get package count by checking which
package manager is installed instead of hardcoding a list of distros.

This means that all we will need to do to add a new linux distro is
add the ascii art/colors.

**Cons**: 
- Slightly slower on Linux if the package manager is towards the bottom
of the if block. (All other OS had a speedup.)

Thanks @tudurom for the idea! :)

### DIFF
--- a/fetch
+++ b/fetch
@@ -638,6 +638,9 @@ getpackages () {
 
             elif type -p nix-env >/dev/null 2>&1; then
                 packages="$(ls -d -1 /nix/store/*/ | wc -l)"
+
+            elif type -p pacman-g2 >/dev/null 2>&1; then
+                packages="$(pacman-g2 -Q | wc -l)"
             fi
         ;;
 
--- a/fetch
+++ b/fetch
@@ -607,50 +607,41 @@ getuptime () {
 # Package Count {{{
 
 getpackages () {
-    case "$distro" in
-        "Arch Linux"* | "Parabola"* | "Manjaro"* | "Antergos"* | "Chakra"*)
-            packages="$(pacman -Qq --color never | wc -l)"
+    case "$os" in
+        "Linux")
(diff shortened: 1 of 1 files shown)


dylanaraps:
I've finished testing this in various Linux, BSD and Windows virtual 
machines and it works fine.

tudurom:
Reviewing it right now. On gentoo, why not check for emerge and then get the number of packages with 

```bash
packages="$(ls -d /var/db/pkg/*/* | wc -l)"
```

?

dylanaraps:
Aren't I doing this already?

```sh
# Block from this PR
elif type -p emerge >/dev/null 2>&1; then
    packages="$(ls -d /var/db/pkg/*/* | wc -l)"
```

tudurom:
Just ate 3 caramels. Can't think :)

tudurom:
On what distro is Solus based on?

dylanaraps:
It's based on Debian Stable afaik.

tudurom:
And it uses `pisi` for package management?

dylanaraps:
Yup, since Solus 2.

https://en.wikipedia.org/wiki/Pardus_%28operating_system%29#PiSi_package_management

tudurom:
Pisi means kitty in Romanian too!

This kind of discussion is for the gitter board...

dylanaraps:
If all looks good I'll merge this. (I'm in the gitter chat btw)

tudurom:
I think you can merge it, but we need Mac OS testing.

dylanaraps:
Mac OS X will work as the only real changes happened to the
Linux package managers.

dylanaraps:
It's now in master.

PR #82, opened 2016-02-18
Font size is now calculated by fetch

dylanaraps:
The user no longer has to set the `font_width` value, fetch calculates it using
terminal size and number of columns/lines.

This change also makes image sizing take terminal height into account, this
fixes issues with terminals where the height is far lower than the width. 

This seems to work in all terminal emulators and font sizes I've tried but I'd
still like some people to test it out!

Thanks

### DIFF
--- a/1.3.md
+++ b/1.3.md
@@ -1,5 +1,1 @@
 # Fetch 1.3
-
-- You no longer need to set the font_width value, your font size<br \>
-is now calculated by fetch.
-- Image sizing now takes terminal height into account.
--- a/1.3.md
+++ b/1.3.md
@@ -1,1 +1,5 @@
 # Fetch 1.3
+
+- You no longer need to set the font_width value, your font size<br \>
+is now calculated by fetch.
+- Image sizing now takes terminal height into account.
--- a/README.md
+++ b/README.md
@@ -63,18 +63,17 @@ your distro's logo or any ascii art of your choice!
 
 ## Dependencies
 
-
 ### Required dependencies:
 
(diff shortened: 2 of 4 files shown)


tudurom:
It looks the same, though.

tudurom:
Now I see. Looks great!

tudurom:
It hangs in tmux.

dylanaraps:
It works just fine in tmux for me.

dylanaraps:
Ok, it seems like this solution only works in terminals that support `xterm`
escape sequences.

**Works in:**
- URxvt
- Xterm
- Vte based terminals (Termite, Xfce4-terminal, gnome-terminal)
- Terminator

**Doesn't work in:**
- st
- konsole

dylanaraps:
I've done some testing with `st` and `konsole` in virtual machines and it
seems that images with fetch didn't work properly in them anyway.

dylanaraps:
I've updated this PR to disable image support in terminal emulators that 
don't support  the xterm escape sequences we're using to size the images.

This means that `w3m` images won't work in `st` and `konsole`. Images
were already buggy and didn't work properly in these terminal emulators
so it's not a big deal.

This is a worthy compromise for making fetch easier to use out of the box. 
With this PR setting up image support is plug and play now, the user doesn't 
have to figure out their font size in pixels or anything.

tudurom:
Now it's working with tmux, for some reason. Thanks!

dylanaraps:
I'll go ahead and merge this then. :)

PR #90, opened 2016-02-20
Image size multiplier.

Wyn10:
After halving the image I still have plenty of room to work with.  Added a multiply so I can increase the base image size before halving without wrecking the final size.

### DIFF
--- a/config/config
+++ b/config/config
@@ -272,6 +272,11 @@ crop_offset="center"
 # --size auto, px
 image_size="auto"
 
+# Image size multiplier
+# Multiply to increase the image by px before halfing.
+# --size_multiplier px
+image_size_multiplier=0 # Recommended start 150
+
 # Right gap between image and text
 # --gap num
 gap=4
--- a/fetch
+++ b/fetch
@@ -290,6 +290,11 @@ crop_offset="center"
 # --size auto, px
 image_size="auto"
 
+# Image size multiplier
+# Multiply to increase the image by px before halfing.
+# --size_multiplier px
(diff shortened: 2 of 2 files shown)


dylanaraps:
I don't quite understand what this new feature does or why we need it, care to elaborate?

Wyn10:
After halving the image to fit into the console I still have room for the image to be bigger without squishing the text.  So I added a multiplier so I can make the image bigger and use the extra space.

dylanaraps:
I still don't understand what you mean by "bigger", can you show me an example of with/without the setting enabled?

Wyn10:
Sure.  I'm better at pictures.

This is with image size multiplier at 0, so half the terminal width as normal:

Alt text

This is with image size multiplier at 300, the image is still half to fit the console:

Alt text

The image gets bigger to fill the unneeded space.  Sorry for the bad explanation.

dylanaraps:
I think I understand now.

It's a good idea, I just think it should be implemented differently. What I've done
in a seperate branch called `size` is make the launch flag `--size` also take a
percentage as a value.

For example:

```sh
# Image takes up 70% of terminal size
fetch --size 70%

# Image takes up 20% of terminal size
fetch --size 20%
```

This way it scales with the window size and also works on smaller terminal sizes.
This also takes terminal height into account when the image won't fit the window
size.

Wyn10:
Perfect :grinning:.  I am using 53% which does the same as I was doing.

Sorry again for the bad explanation.

dylanaraps:
All good! 

Thanks for the original PR. :)

PR #95, opened 2016-02-21
Add Battery Support for FreeBSD

tudurom:
Fix it.

### DIFF
--- a/fetch
+++ b/fetch
@@ -1398,10 +1398,19 @@ getbattery () {
                 battery="None"
             fi
         ;;
+        
+        "BSD")
+          case "$distro" in
+            "FreeBSD"*)
+              battery=$(acpiconf -i 0 | awk -F ':\t' '/Remaining capacity/ {print $2}')
+            ;;
+          esac
+        ;;
 
         "Mac OS X")
             battery="$(pmset -g batt | grep -o '[0-9]*%')"
         ;;
+
 
         "Windows")
             battery="$(wmic Path Win32_Battery get EstimatedChargeRemaining /value)"
@@ -2081,7 +2090,7 @@ stdout () {
             *)
                 "get$func" 2>/dev/null
                 eval output="\$$func"
-                printf "$output  "
+                printf "%s" "$output  "
             ;;
         esac
     done


fetch
```diff
@@ -1389,6 +1389,11 @@ getbattery () {
             fi
         ;;
 
+        "BSD")
+          battery=$(acpiconf -i 0 | awk -F ':\t' '/Remaining capacity/ {print $2}')
+          battery=${battery//%}
```
dylanaraps: Doesn't this line remove all instances of `%`?

fetch
```diff
@@ -1389,6 +1389,11 @@ getbattery () {
             fi
         ;;
 
+        "BSD")
```
dylanaraps: If this only works on `FreeBSD` then you should do it like this:

```sh
case "$os" in
  "BSD")
      case "$distro" in
          "FreeBSD"*)
              # FreeBSD battery stuff
          ;;
      esac
  ;;
esac
```

dylanaraps:
Other than those two comments LGTM.

tudurom:
There's only one occurence of '%'. Btw can't fix it now.

dylanaraps:
Why are you removing the `%` though? Looking at the PR it seems like the output will just be missing the `%` character.

fetch
dylanaraps: The reason `OpenBSD` has a seperate line in `case "$(uname)"` part of the script
is that the output of `uname` om OpenBSD is entirely different to all other BSDs for 
some reason

tudurom:
Trust me, there is a % and it needs to be removed. On Linux there's no %.

dylanaraps:
What I'm trying to say is that the output needs to include the `%` sign. It should
be appearing on Linux too, have a look through the battery function and you'll
see these three lines which add the `%` sign to the output:

```sh
battery="${battery}%"
battery="${batteries[0]}%"
battery="${batteries[0]}%"
```

tudurom:
I will do it later. I'm busy now.

dylanaraps:
I can do it myself if you'd like.

tudurom:
Stay chill...

tudurom:
Done. You can see I fixed a huge bug that occurs when using `fetch --stdout battery`.

tudurom:
Tested, you can merge it.

fetch
```diff
@@ -2083,7 +2091,7 @@ stdout () {
             *)
                 "get$func" 2>/dev/null
                 eval output="\$$func"
-                printf "$output  "
+                echo -n "$output  "
```
dylanaraps: Change this back to `printf`

fetch
```diff
@@ -1401,6 +1401,14 @@ getbattery () {
             fi
         ;;
 
+        "BSD")
+          case "$distro" in
+            "FreeBSD"*)
+          battery=$(acpiconf -i 0 | awk -F ':\t' '/Remaining capacity/ {print $2}')
```
dylanaraps: Fix the indentation here.

dylanaraps:
I left some more comment, also squash the commits into a single one.

tudurom:
1. How do I squash them
2. Why printf? It breaks everything when printing the battery info because
the '%' at the end will be interpreted as a format specifier.

dylanaraps:
Change the printf line to this:

```sh
printf "%s" "$output  "
```

tudurom:
But why not echo?

dylanaraps:
Echo isn't portable.

http://unix.stackexchange.com/questions/65803/why-is-printf-better-than-echo

dylanaraps:
The easiest way to squash commits is to do this:

```sh
# Change '3' ti the number of commits to squash
git reset --soft HEAD~3 &&
git commit -m "New commit message"
```

tudurom:
Thanks, TIL!

That commit squash is a great hack. Will do in a minute.

dylanaraps:
Just a heads up: You may need to force push the squashed commit to github.

```sh
git push -f origin master
```

tudurom:
I think I nailed it.

tudurom:
The commit got squashed with the "Added Windows support for resolution" but eh.

dylanaraps:
You'll have to fix that as I'll have to force push to master which 
will break everybody's local copy and I don't want to do that.

Also, you still need to fix the indentation.

If you have any issues I can merge this manually.

tudurom:
Fuck this, I will reset my fork to upstream and readd support.

tudurom:
Done.

dylanaraps:
Thanks. :)

tudurom:
Go write about this in the changelog and in the TODO issue. Next stop: OpenBSD!

dylanaraps:
:+1:

PR #98, opened 2016-02-23
Battery support for OpenBSD

tudurom:


dylanaraps:
What does the output look like?

Also your indenting is wrong again, you've got `5` spaces instead of `4`.

tudurom:
My vim indents with 2 spaces because that's how I like it. On fetch I
indent manually. The output looks like `97%`.

dylanaraps:
I'll add a comment so that vim changes the indent option just for fetch.

dylanaraps:
I'm on Windows atm so I'll update the changelog later.

dylanaraps:
Thanks for the help! :)

tudurom:
Why are you on windows? `∑(O_O；)`?

dylanaraps:
I'm playing CS:GO. :)

tudurom:
Well it seems you're using AMD! I have a friend which plays CS:GO on AMD with no problem...

That's off-topic.

dylanaraps:
The newest update has some performance issues, I was playing on linux full time before.

PR #109, opened 2016-02-27
Don't create a config file in stdout mode.

dylanaraps:
This PR disables the creation of the config file in stdout mode.

Please test this @tudurom.


tudurom:
Thanks! Merge it.

PR #113, opened 2016-02-28
Fix 0 package count in Slackware

### LINKED ISSUE #112: Reports 0 packages installed in Slackware
Under Slackware 14.1-current, when running as a non-root user, the script reports that there are no packages installed.

    Packages: 0

(I've already fixed it, I just wanted an issue number to reference.)

h3xx:
Fixes #112.

Rationale:
- By default in Slackware, /sbin and /usr/sbin aren't part of the user's
  PATH. (It's been this way since I started using it in 2004, version
  9.1)
- `type -p /path/to/binary` works just like it was used with a bare
  binary name.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -574,7 +574,7 @@ getpackages () {
             elif type -p pacman >/dev/null 2>&1; then
                 packages="$(pacman -Qq --color never | wc -l)"
 
-            elif type -p pkgtool >/dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


dylanaraps:
Thanks :+1:

h3xx:
Thanks for supporting Slackware. :-)

dylanaraps:
No problem!

PR #117, opened 2016-02-28
Fix help message text

h3xx:
Fixes `Which Distro\'s ascii art to print` in output when running `--help`. (Notice the backslash-quote.)

Caused by a rogue backslash in the heredoc.

Aside: I don't think Github's syntax highlighting is doing a good job here.


dylanaraps:
Thanks :+1:

PR #122, opened 2016-03-04
man pages?

konimex:
For some people (including me, I guess), `screenfetch --help` is a bit... inconvenient, because you have to scroll your mouse to find your desired option, and some of us want to be as mouse-free as possible, so I thought that man page would be beneficial, so you can find it easily, scroll-free.

I don't know how other distros/CRUX port/Gentoo ebuilds will handle this, but in AUR you'll have to add this into the PKGBUILD's `package()`:

    install -Dm644 "neofetch.1" "$pkgdir/usr/share/man/man1/neofetch.1"

### DIFF
--- a/neofetch.1
+++ b/neofetch.1
@@ -1,0 +1,232 @@
+.TH NEOFETCH "1" "March 2016" "1.3" "User Commands"
+.SH NAME
+neofetch \- simple system information script
+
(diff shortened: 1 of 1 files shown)


dylanaraps:
Wow, thanks for taking the time to do this!

PR #126, opened 2016-03-09
Added OS X install instructions

iandrewt:
obviously they aren't applicable just yet :P


dylanaraps:
:+1:

Thanks for taking the time to do this!

Issue #130, opened 2016-03-10
CPU Shorthand


SethBarberee:
So it just an idea but maybe shortening CPU output could be a good idea? Like only display the model of CPU and cut off speed or maybe only display the speed. It might be a good idea so in tiling environments, the output from CPU doesn't have to wrap to the next line

dylanaraps:
Thanks to @iandrewt, this is now in master.

PR #131, opened 2016-03-10
added ability to shorten cpu output

iandrewt:
usage: --cpu_shorthand (name, model, name_model, speed)
also edited config file to suit

addresses #130
documentation updates to come

### DIFF
--- a/1.5.md
+++ b/1.5.md
@@ -11,7 +11,7 @@
 **Disk Usage**<br \>
 - Only display usage of local disks.
 **CPU Information**<br \>
-- Option to shorten CPU name to just name, model, name & model, or speed
+- Option to shorten CPU name to just name, speed, tiny, on, or off
 
 ### Ascii
 
--- a/1.5.md
+++ b/1.5.md
@@ -10,6 +10,8 @@
 
 **Disk Usage**<br \>
 - Only display usage of local disks.
+**CPU Information**<br \>
+- Option to shorten CPU name to just name, model, name & model, or speed
 
 ### Ascii
 
--- a/README.md
+++ b/README.md
@@ -270,7 +270,7 @@ alias fetch2="fetch \
                                 scaling_current, scaling_min, scaling_max
                                 NOTE: This only support Linux with cpufreq.
     --cpu_shorthand type        Shorten the output of CPU
-                                Possible values: name, model, name_model, speed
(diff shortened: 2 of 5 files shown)


dylanaraps:
Wow, thanks for the PR.

I'll do some testing before merging this. :)

iandrewt:
I've only tested on OS X, so Intel CPUs should work fine (unless output is different on other OSes?)

iandrewt:
That's great! It just cuts the string at the dash and/or the @ depending on what option is chosen, so as long as the output is always in that format it should work fine

iandrewt:
Based on the screenshots in `README.md`, Windows might have an issue as it doesn't have a dash between the "i5" and "760".

dylanaraps:
The only issue I've found is that some CPUs don't have a `-` character in their name causing some of the shorthand options to not work. Here's an example:

```sh
# Add this just before the new shorthand case statement
cpu="Intel Core2 Duo T7500 @ 2.2GHz"

```

Here's another idea for the shorthand options which would work consistently. 

```sh
# Hide the speed
--cpu_shorthand name 

# Hide the brand 
# (Intel, AMD) 
# Output: i7 2600 @ 2.2GHz
--cpu_shorthand on

# Hide the speed and the brand name
# Output: i7 2600
--cpu_shorthand tiny
```

What do you think?

(I was typing this out before your latest comment :P)

dylanaraps:
Here's an example that works on all the cpus I've tried.

```sh
    case "$cpu_shorthand" in
        "name")
            cpu=${cpu/@*}
        ;;

        "speed")
            cpu=${cpu#*@ }
        ;;

        "on")
            cpu=${cpu/Intel }
            cpu=${cpu/Core }
            cpu=${cpu/AMD }
        ;;

        "tiny")
            cpu=${cpu/Intel }
            cpu=${cpu/Core }
            cpu=${cpu/AMD }
            cpu=${cpu/@*}
        ;;
    esac
```

dylanaraps:
I edited the case statement after I commented to remove the duplicate block between `on` and `tiny`, I also added another line to remove `CoreX Duo`. Once you make these changes I'll merge this. :)

iandrewt:
Done! Works great with the original string `Intel Core i5-3210M @ 2.5GHz` and the Core2 Duo string too

dylanaraps:
Thanks for the PR!

PR #134, opened 2016-03-11
Added installation instructions for RHEL and CentOS

konimex:
Added another repo for CentOS.

Since RHEL and CentOS do not use dnf (they're based
on Fedora 19), we need to use separate instructions
for now. Although this package is originally
intended for CentOS, but since CentOS is basically
RHEL without RH's trademark, it will be the same.
But I can't say the same for Scientific Linux or
Oracle Linux, even if both of them are based on RHEL.


dylanaraps:
There's another table of contents towards the top of the packages section, add the new line to that too and I'll merge this.

Thanks again! :D

dylanaraps:
Thanks!

PR #140, opened 2016-03-12
fixed a few issues with the progress bars

iandrewt:
- disk usage bar didn't work on drives under a terabyte
- disk usage percentage still printed when disabled
- battery info option was inversed


dylanaraps:
Thanks for the fixes!

PR #146, opened 2016-03-14
fixed output of disk bar when gigabytes are used of a terabyte disk

iandrewt:
apparently bar doesn't like it when the first variable is bigger than the second, it went really haywyre.

to see old behaviour, set `disk_used` to `63` and `disk_total` to `3.1` (that's what my disk usage and disk total is, in gigabytes and terabytes respectively)


dylanaraps:
Thanks!

PR #148, opened 2016-03-15
fixed osx_buildver option

iandrewt:
it would always print the OS X build version before because the option wasn't parsed before adding the build. Now it's removed after the fact if it needs to be.


dylanaraps:
:+1:

PR #149, opened 2016-03-15
added cpu cores so the usage bar doesn't go crazy above 100%

iandrewt:
I guess adding the CPU cores to the CPU output would be possible now too.

Still need to add NetBSD and Windows support though.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -849,11 +849,15 @@ getcpu () {
             speed=${speed:0:1}.${speed:1}
 
             cpu="$cpu @ ${speed}GHz"
+
+            # Get cpu cores
+            cores=$(nproc)
         ;;
 
         "Mac OS X")
             cpu="$(sysctl -n machdep.cpu.brand_string)"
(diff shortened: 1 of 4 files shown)


dylanaraps:
Awesome!

This is probably faster than `nproc` on systems with the `/proc/` fs.

```sh
awk -F ': ' '/siblings/ {printf $2; exit}' /proc/cpuinfo
```

neofetch
```diff
@@ -849,11 +849,15 @@ getcpu () {
             speed=${speed:0:1}.${speed:1}
 
             cpu="$cpu @ ${speed}GHz"
+
```
dylanaraps: Remove the empty line and the `# Get cpu cores` line as the cores var explains itself clear enough.
eg.
```sh
cpu="$cpu @ ${speed}GHz"
cores=$(awk -F ': ' '/siblings/ {printf $2; exit}' /proc/cpuinfo)
```

PR #153, opened 2016-03-16
Fixed wallpaper error with nitrogen

CousinMachu:
Fixed the awk string to terminate after grabbing the first wallpaper found in nitrogen instead of mixing the two strings together and using the ascii art instead. In later revisions, there should be a way to check it's using the current monitor's wallpaper (I don't believe nitrogen orders them by usage, but rather by output).


dylanaraps:
LGTM Thanks!

I don't think it's possible to grab the current monitor's wallpaper without over 
complicating the wallpaper function. I have no intention of adding this either
but I'll accept pull requests that implement it if they're reasonable.

PR #157, opened 2016-03-17
Update CRUX distro name

onodera-punpun:
CRUX has a binary called crux that print the distro name and the version (CRUX version 3.2 for example).


dylanaraps:
:+1: Thanks

PR #158, opened 2016-03-17
added quotes in curl + small cosmetical changes

dawidd6:
ZSH interprets `?` as wildcard, so i've added quotes in curl instruction. Additionally i've made some cosmetical changes


dylanaraps:
:+1:

PR #165, opened 2016-03-19
Simplified finding refresh rate

CousinMachu:


dylanaraps:
Yo, I'm still working on the refresh rate stuff in my local branch and the version in the github repo is outdated. Thanks for the PR but there's a few issues:

- This requires the user to have both `xorg-xdpyinfo` and `xrandr` installed. 
- This won't work with multiple monitors. 
- This doesn't use `xrandr`'s command line flags `--nograb --current` which speed up xrandr's output.

CousinMachu:
Sounds good, thanks for the heads up.

PR #167, opened 2016-03-20
fix for mpc detection

dawidd6:
In my case your `if pgrep "mpd" >/dev/null 2>&1` is not working because i'm using mpd on server and mpc on my laptop, so i don't have mpd on it and `pgrep` is failing. I think it is better to search for mpc than for mpd in this `if`

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -1192,7 +1192,7 @@ getmemory () {
 # Song {{{
 
 getsong () {
-    if pgrep "mpd" >/dev/null 2>&1; then
+    if hash "mpc" >/dev/null 2>&1; then
         song="$(mpc current 2>/dev/null)"
         state=$(mpc | awk -F '\\[|\\]' '/\[/ {printf $2}' 2>/dev/null)
 
--- a/neofetch
+++ b/neofetch
@@ -1192,7 +1192,7 @@ getmemory () {
 # Song {{{
 
 getsong () {
-    if hash "mpc" >/dev/null 2>&1; then
+    if mpc version >/dev/null 2>&1; then
         song="$(mpc current 2>/dev/null)"
         state=$(mpc | awk -F '\\[|\\]' '/\[/ {printf $2}' 2>/dev/null)
 


dylanaraps:
Hm, `mpc` isn't always running though is it?

dylanaraps:
We could do `if pgrep "mpd|mpc" >/dev/null 2>&1` which should work right?

dawidd6:
Unfortunelly pgrep is not working for mpc, so i've done this with `hash`

dawidd6:
But mpd is not optional dependency for neofetch and mpc is

dylanaraps:
Well we're using `mpc` to talk to `mpd` so it's implied that you need both installed. Is there a way to check that you're connected to mpd?

dawidd6:
Yes, simply type `mpc` and output is saying under condition if you have MPD_HOST variable exported

dylanaraps:
Awesome, I'll add this tomorrow! (I'm not on my linux pc atm.)

dawidd6:
Damn, i see that my next commit has automatic apperared here.

okay, so i've added detection of charging status for single battery

zrzut ekranu z 2016-03-20 13-01-18

dylanaraps:
Create a new branch for the battery stuff and make it so that it only shows status when the value is `Charging`. 

We also need to get it working when using multiple batteries!

dawidd6:
Man, i have no idea how can i remove commit from this PR

dylanaraps:
You can start over and instead work in two seperate branches.

dawidd6:
OK, done. Created new branch and deleted commit from master :+1:

dylanaraps:
I'll take a look at this tonight, I'm swamped with homework.

dylanaraps:
I did some testing and changing the line to this should work and is much simpler!:

```sh
if mpc version >/dev/null 2>&1; then
```

dylanaraps:
Thanks mate!

PR #173, opened 2016-03-25
Added Google Play Music Desktop Player as a song output source

iandrewt:
Adds optional dependency `gpmdp`

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -1228,9 +1228,9 @@ getsong () {
             ;;
         esac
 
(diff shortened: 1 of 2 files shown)


neofetch
```diff
@@ -1228,6 +1228,10 @@ getsong () {
             ;;
         esac
 
+    elif ([ -n "$(ps aux | awk '!(/awk/ || /Helper/) && /Google Play Music Desktop Player/')" ] && [ -n "$(which gpmdp)" ]); then
```
dylanaraps: Are the `()` brackets needed around the start and end of the elif?

Also, change the `[ -n "$(which gpmdp)" ]` part to:

```sh
type -p gpmdp >/dev/null 2>&1
```

dylanaraps:
:+1:

Issue #169, opened 2016-03-20
Cannot get desired image to print


misterch0c:
Hi there,
I've been struggling to get a chosen image to output. if I run neofetch without any config I get my wallpaper showing as an image, but using --shuffle-dir or just giving an image as parameter I can't seem to change the image...
I've tried using iterm2 & then I get this:
http://i.imgur.com/cT24SeP.png

dylanaraps:
Have you set `image_backend` to `iterm2`?

misterch0c:
yes, and the result is the image I posted. So looks like raw data is outputed instead of the image

dylanaraps:
@iandrewt, can you take a look at this? I don't have an OS X system to test on.

iandrewt:
@misterch0c what does your config file look like in this section?

```sh
225 # Image Options {{{
226
227
228 # Image Source
229 # --image wall, shuffle, ascii, /path/to/img, off
230 image="wall"
231
232 # Thumbnail directory
233 thumbnail_dir="$HOME/.cache/thumbnails/neofetch"
234
235 # Image Backend
236 # Which program to draw images with
237 # --image_backend w3m, iterm2
238 image_backend="w3m"
239
240 # W3m-img path
241 # Some systems have this in another location
242 w3m_img_path="/usr/lib/w3m/w3mimgdisplay"
243
244 # Image position
245 # Only works with the w3m backend
246 # --image_position left/right
247 image_position="left"
248
249 # Shuffle dir
250 shuffle_dir="$HOME/Pictures/wallpapers/wash"
251
252 # Crop mode
253 # --crop_mode normal/fit/fill
254 crop_mode="normal"
255
256 # Crop offset
257 # Only affects normal mode.
258 # --crop_offset northwest/north/northeast/west/center
259 #               east/southwest/south/southeast
260 crop_offset="center"
261
262 # Image size
263 # The image is half the terminal width by default.
264 # --size auto, 00px, 00%
265 image_size="auto"
266
267 # Right gap between image and text
268 # --gap num
269 gap=4
270
271 # Image offsets
272 # --xoffset px
273 # --yoffset px
274 yoffset=0
275 xoffset=0
276
277
278 # }}}
```

This is the default config. I can't seem to reproduce this issue with the default config on version `1.5` when running `neofetch --image_backend iterm2 --shuffle_dir ~/Pictures`

misterch0c:
[...]

PR #174, opened 2016-03-25
Remove shuffledir in favor of '--image path/to/dir/'

dylanaraps:
This PR removes the shuffle mode and all the config options and flags that came with it and replaces it with `--image path/to/dir/` and `image="path/to/dir/"` instead. 

- This will hopefully clear up the confusion in issue **#169**.
- You can set the default shuffle dir in the config by changing `$image` to a directory path.
- The value **must** end in a `/`.

TODO

- [x] ~~Update the Man page~~
- [x] ~~Update the Readme~~

I'd like to get some confirmation that this works on other platforms before I merge this to master.

cc @misterch0c @iandrewt

### DIFF
--- a/README.md
+++ b/README.md
@@ -284,13 +284,15 @@ alias fetch2="fetch \
 ## Usage
 
 
-    usage: fetch --option "value" --option "value"
+    usage: neofetch --option "value" --option "value"
+
+    NOTE: There's also a config option for each flag below.
 
     Info:
     --disable infoname          Allows you to disable an info line from appearing
                                 in the output.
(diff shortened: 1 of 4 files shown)


iandrewt:
`shuf` doesn't work on OS X

dylanaraps:
Oh damn, I'm looking up alternatives to shuf now.

dylanaraps:
Try out this PR now, I've swapped shuf for an alternative using printf + an array.

iandrewt:
Works now!

PR #178, opened 2016-03-25
Fix colors in older versions of Mac OS X, OpenBSD etc

dylanaraps:
This PR changes the escape sequence we're using for coloring to something that works on a lot of older operating systems too. This also clears the coloring beforehand which fixes some other bugs
related to color I've found.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -2339,30 +2339,30 @@ colors () {
 }
 
 setcolors () {
-    c1="\033[38;5;${1}m"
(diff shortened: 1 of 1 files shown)


dylanaraps:
@iandrewt, This should fix those color issues in the old Mac OS X versions you tried about a week ago.

iandrewt:
Works in Snow Leopard's Terminal!

dylanaraps:
Awesome!

PR #181, opened 2016-03-26
fixed if statement for linux resolution

iandrewt:
the `&& \` method doesn't work with `elif`s it seems. breaks the script on OS X even though that block of code never fires.


dylanaraps:
Thanks, I didn't catch this when I changed to the `if` syntax.

PR #185, opened 2016-03-26
Use $GTK2_RC_FILES for GTK2 config if set

onespaceman:
I export GTK2_RC_FILES to $XDG_CONFIG_HOME/gtk-2.0/gtkrc (as specified in the [Arch wiki](https://wiki.archlinux.org/index.php/XDG_Base_Directory_support)) to help keep my home folder clean.

This way neofetch uses GTK2_RC_FILES if it is set, and the default location if it is not.


dylanaraps:
Awesome, thanks!

PR #192, opened 2016-03-27
Added installation instructions for Void

konimex:


dylanaraps:
Oh neat, we're in the official repos!

dylanaraps:
Thanks!

PR #202, opened 2016-03-30
made OS X packages SUPER fast

iandrewt:
could probably be made faster still, but the main culprit for speed reduction is gone

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -695,8 +695,8 @@ getpackages () {
             fi
 
             if type -p brew >/dev/null 2>&1; then
-                brew_packages=$(brew list -1 2>/dev/null | wc -l)
-                packages=$((packages + brew_packages))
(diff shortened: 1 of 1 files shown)


neofetch
```diff
@@ -695,8 +695,8 @@ getpackages () {
             fi
 
             if type -p brew >/dev/null 2>&1; then
-                brew_packages=$(brew list -1 2>/dev/null | wc -l)
-                packages=$((packages + brew_packages))
+                brew_packages=$(find /usr/local/Cellar -maxdepth 1 2>/dev/null | wc -l)
+                packages=$((packages + $((brew_packages - 1))))
```
dylanaraps: This line could be:

```sh
packages=$((packages + brew_packages - 1))
```

PR #233, opened 2016-04-30
fixed value of used memory

dawidd6:
zrzut ekranu z 2016-04-30 16-01-40

I think that now the memory usage is showing more accurate value

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -1207,10 +1207,9 @@ getgpu () {
 getmemory () {
     case "$os" in
         "Linux" | "Windows")
-            mem=($(awk -F ':| kB' '/MemTotal|MemFree|Buffers|Cached/ {printf $2}' /proc/meminfo) 0 0)
-            memused=$((mem[0] - mem[1] - mem[2] - mem[3]))
-            memused=$((memused / 1024))
-            memtotal=$((mem[0] / 1024))
+            memtotal=$(($(awk '/MemTotal/ {print $2}' /proc/meminfo) / 1024))
+            memfree=$(($(awk '/MemAvail/ {print $2}' /proc/meminfo) / 1024))
+            memused=$(($memtotal - $memfree))
         ;;
 
         "Mac OS X")


dylanaraps:
Hello, thanks for the PR.

`MemAvail` isn't available on older linux kernels so we have to calculate it manually. This PR won't work on older systems sadly.

TheDoctorsLife:
couldn't you check for older kernels and the current behavior, then if its a new kernel with ``MemAvail``?

dylanaraps:
That's a good idea, I'll dig up the kernel version `MemAvail` was added to.

dylanaraps:
`MemAvail` was added in linux `3.2`

dylanaraps:
It's going to be difficult to check for version numbers when people have custom kernel names and etc. We should look at how `MemAvail` is calculated and do it ourselves with the existing `/proc/meminfo` data.

dylanaraps:
This is the commit that added `MemAvail`.

https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/commit/?id=34e431b0ae398fc54ea69ff85ec700722c9da773

dylanaraps:
I don't think that we can calculate manually very accurately, here's a snippet from the docs for `MemAvail`:

> Calculated from MemFree, SReclaimable, the size of the file LRU lists, and the low watermarks in each zone.

dylanaraps:
I've modified the current memory function and I almost have the right output.

```sh
mem=($(awk -F ':| kB' '/MemTotal|MemFree|Buffers|Cached|SReclaimable|PageTables/ {printf $2}' /proc/meminfo) 0 0)
memused=$((mem[0] - mem[1] - mem[2] - mem[3] + mem[4] + mem[5] + mem[6]))
memused=$((memused / 1024))
memtotal=$((mem[0] / 1024))
```

On my system the output between `MemAvail` and this new function only differs by 10MB. Can you guys confirm that this works better than the current implementation we have?

TheDoctorsLife:
Sure, give me a few minutes, and I'll output what I notice.

dylanaraps:
This version and `MemAvail` are pretty much identical on my PC, test both and let me know how close they are to the `MemAvail` version.

```sh
mem=($(awk -F ':| kB' '/MemTotal|MemFree|Buffers|Cached|SReclaimable|SUnreclaim|KernelStack/ {printf $2}' /proc/meminfo) 0 0)
memused=$((mem[0] - mem[1] - mem[2] - mem[3] + mem[4] + mem[5] + mem[6] - mem[7]))
memused=$((memused / 1024))
memtotal=$((mem[0] / 1024))
```

TheDoctorsLife:
Where can I find this change?

dylanaraps:
https://github.com/dylanaraps/neofetch/pull/234

TheDoctorsLife:
Last Release
original1
MemAvail
original2
ScreenFetch
original3
cat /proc/meminfo
original4

dylanaraps:
Which one is #234?

TheDoctorsLife:
The second, sorry for not labeling it to well.

dylanaraps:
No problem, hiow does #234 compare to #233?

TheDoctorsLife:
let me see

TheDoctorsLife:
#233 
NeoFetch1
#234
NeoFetch2

dylanaraps:
Did you capture the output of both at the exact same time? The amount differs fairly quickly.

Even with a 10-15MB difference this is still better than master.

TheDoctorsLife:
I did, I captured them in about 5 seconds in-between.

dylanaraps:
5 seconds may have been too long between them, for me the output changes within a second.

TheDoctorsLife:
I'll close some programs and take new pictures.

TheDoctorsLife:
These were taken after closing programs that tend to change the amount of ram they use frequently. I got it to .1-4 fluctuation of the ram usage, but besides that everything should be accurate.

#233  
---
screenshot_2016-05-01_08-19-08
#234 
---
screenshot_2016-05-01_08-19-45
Last NeoFetch Release
---
screenshot_2016-05-01_08-30-52
ScreenFetch
---
screenshot_2016-05-01_08-20-07
cat /proc/memavail
---
screenshot_2016-05-01_08-20-36
xfce4 task manager
---
screenshot_2016-05-01_08-20-55

dylanaraps:
@TheDoctorsLife, #234 worth merging then?

TheDoctorsLife:
In my testing, yes.

dylanaraps:
Can anyone else compare the output between #233 and #234? 

Note: Try to run both neofetch versions at the exact time as the memory output will fluctuate.

iandrewt:
This is Linux only, right? I'll test it when I get home (6 hours)

dylanaraps:
It's Linux / Windows only.

dylanaraps:
#236 solves this, thanks for the suggestions.

PR #234, opened 2016-05-01
[Linux / Windows] Memory function: Mimic MemAvail

dylanaraps:
This PR makes the memory function mimic `memavail` to get a more accurate free memory output.

TODO:

- Testing


dylanaraps:
#236 solves this.

PR #238, opened 2016-05-08
Add support for Trinity DE and Twin WM.

dylanaraps:
Add support for Trinity DE and Twin WM.

TODO: 

- [ ] Testing

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -908,14 +908,14 @@ getwmtheme () {
             fi
         ;;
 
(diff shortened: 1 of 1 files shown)


dylanaraps:
The build of `trinity` I used still identified the wm as kwin instead of twin, I need to do some more investigation.

dylanaraps:
I'm going to revisit this at a lter time.

PR #240, opened 2016-05-08
Add Android Support

dylanaraps:
This PR aims to get neofetch working on Android devices.

Issues:

- [x] The `awk` command to get the length of the longest line in the ascii art doesn't work with busybox's `awk`.
- [x] There's no `tput` on android so we need to suppress the command not found errors.
    - `tput` is only needed for image mode which won't work on android anyway. 
- [ ] `./neofetch` doesn't work on android for some reason.
    - `bash ./neofetch` works though.
- [x] Fix command not found errors.
    - [x] `stty`

Apart from these issues above, all of the info displays correctly.

TODO

- [ ] Test on various android devices.
    - I can only test on an android 4.4.4 device.


dylanaraps:
Possible solutions to the awk problem:

- `awk '{ if ( length > L ) { L=length} }END{ print L}'`

konimex:
Tested on an Android 6.0, here's the output.

result

It works.

dylanaraps:
Thanks!

It looks like the android OS detection failed. (Which then caused some of the other functions to not use android specific blocks)

Does this file exist in this directory on your device? `system/app/SystemUI.apk`

If not, then we'll need to find a more reliable way of detecting Android. (`uname` prints `Linux` so we have to find a command or file that is android specific for detection.)

konimex:
On Wed, May 11, 2016 at 10:55:25PM -0700, Dylan Araps wrote:
> Does this file exist in this directory on your device? `system/app/SystemUI.apk`

No, but there is `/system/prtv-app/SystemUI/SystemUI.apk`.

konimex:
No, but there is `/system/priv-app/SystemUI/SystemUI.apk`

dylanaraps:
Yeah, I'll have to find a more reliable way of detecting android.

dylanaraps:
@konimex, can you test the latest version of this PR?

All of the issues above should be fixed.

konimex:
result

Yep. It works. However, the uptime shows users instead of the actual uptime.

I changed the awk line to `awk -F ', | ' '{printf $5}'` to show the actual uptime. It works, but I don't know if it's technically correct since I don't know a single thing about awk.

dylanaraps:
Thanks for testing!

- What's the output of `uptime` on it's own?
- It also seems like CPU name is broken, what do these commands output on your system?
    - `awk -F ': | @' '/model name/ {printf $2; exit}' /proc/cpuinfo`
    - `cat /proc/cpuinfo`

konimex:
>What's the output of `uptime` on it's own?

Not much difference from the usual Linux.
`16:30:44 up  6:13,  0 users,  load average: 5.47, 5.77, 5.73`

Output of `cat /proc/cpuinfo`:

    Processor	: AArch64 Processor rev 1 (aarch64)
    processor	: 0
    processor	: 1
    processor	: 2
    processor	: 3
    processor	: 4
    processor	: 5
    processor	: 6
    processor	: 7
    Features	: fp asimd aes pmull sha1 sha2 crc32 
    CPU implementer	: 0x41
    CPU architecture: 8
    CPU variant	: 0x0
    CPU part	: 0xd03
    CPU revision	: 1
    
    Hardware	: Qualcomm Technologies, Inc MSM8939

The output of the `awk` command is empty.

dylanaraps:
For the CPU does this command work?

```sh
awk -F ': | @' '/model name|Processor/ {printf $2; exit}' /proc/cpuinfo 
```

konimex:
Yes. Output: `AArch64 Processor rev 1 (aarch64)`

dylanaraps:
I'm going to leave Android support for next release, thanks for testing.

dawidd6:
screenshot_20160523-092517

dylanaraps:
Due to the large amount of differences between android devices and packages installed, I think it's easier to force `--image off` by default on android. (This will fix padding issues).

konimex:
Btw @dawidd6, it seems on your output it can't detect CPU, RAM, and battery. Not sure about this but I think `/proc/meminfo` and `/proc/cpuinfo` didn't exist on your device?

dawidd6:
Actually they are in right place. I've changed rom to 5.1.1 and now ram usage is shown, but cpu remains the same

tudurom:
screenie

Tested with [Termux](https://termux.com/). The reported CPU frequency is wrong, mine has 1.2 GHz.

dylanaraps:
Thanks for testing :+1:

PR #255, opened 2016-05-17
Image directory randomization can go out of bounds

gabe565:
Right now it does not subtract 1 from the total number of files in a directory, so that sometimes it will not find a file and have to fallback to ascii art.
This *should* always fix this.


gabe565:
Awesome, thank you!

dylanaraps:
Thanks, works great!

Issue #258, opened 2016-05-20
CPU Usage is 404% on FreeBSD


tudurom:
## System Information

FreeBSD 10.3-RELEASE

#### Neofetch version

HEAD

## Screenshot

![Screenshot](http://pub.iotek.org/p/9mIE1NK.png)

## Verbose log

https://p.iotek.org/ha5

I should come with a fix later.

tudurom:
**News:** `ps` aux reports 399.5% idle. I think the output should be divided by the number of cores.

tudurom:
See #259. (yes, it's a recursive `see`)

dylanaraps:
iostat doesn't come preinstalled on linux systems afaik. Is it preinstalled on BSD?

tudurom:
It is.
On 21 May 2016 02:36, Dylan Araps <notifications@github.com> wrote:iostat doesn't come preinstalled on linux systems afaik. Is it preinstalled on BSD?

PR #259, opened 2016-05-20
Fix CPU usage on BSDs

tudurom:
This should work on all BSDs.

See #258.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -1173,7 +1173,7 @@ getcpu () {
     [ ! -z "$cpu" ] && prin "$subtitle" "$cpu"
 
     if [ "$cpu_display" != "off" ]; then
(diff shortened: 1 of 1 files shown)


dylanaraps:
I'm going to look into adding a more OS specific version of the CPU Usage.

tudurom:
what's wrong with this one?

tudurom:
oh i put the line in the wrong spot.

dylanaraps:
I'm not happy with using iostat to get the cpu usage.

It isn't installed by default in a lot of distros and iirc there's no version of iostat for CYGWIN.

I want to find a more universal solution if possible.

tudurom:
i wanted initially to add that only for the BSDs. iostat is installed on all BSD operating systems

dylanaraps:
The problem is that this PR adds it for all operating systems.

PR #260, opened 2016-05-21
Add Fish version and fix BC scale.

maddcoder:
I added support for displaying the fish shell version.

I also have a `.bcrc` file that set the `scale` option, so all divides using `bc` have decimal places displayed. That makes for some ugly output. I added "scale=0" before doing the `bc` memory calculation, and it's much cleaner now.


dylanaraps:
LGTM

Thanks!

PR #262, opened 2016-05-23
fixed song output for cmus

iandrewt:
Previously would print ` - "song name" "song artist"`, now prints `"song artist" - "song name"`


dylanaraps:
+1

LGTM

PR #264, opened 2016-05-24
Unified installation instructions for Red Hat distros

konimex:
Since I found out that Fedora, RHEL and CentOS have a pretty much same installation instructions (the difference being `yum` and `dnf` only), I decided to unify the instructions.


dylanaraps:
:+1:

PR #269, opened 2016-06-01
migrate from gpmdp to gpmdp-remote

iandrewt:
I've had to change the name of gpmdp-bash to gpmdp-remote due to Google Play Music Desktop Player adding `gpmdp` as an alias to open their app. This fixes neofetch to reflect this change


dylanaraps:
:+1:

PR #271, opened 2016-06-04
fixed cmus properly, now works regardless of output order

iandrewt:
Got it working in one call of `cmus-remote -Q` too :+1:

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -1452,10 +1452,16 @@ getsong () {
         state=$(mpc | awk -F '\\[|\\]' '/\[/ {printf $2}' 2>/dev/null)
 
     elif [ -n "$(ps x | awk '!(/awk/) && /cmus/')" ]; then
-        song="$(cmus-remote -Q | grep "tag artist \|title" 2>/dev/null)"
-        artist="${song##*tag artist }"
-        title="${song##*tag title }"
-        title="${title%%tag artist*}"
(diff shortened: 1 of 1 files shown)


dylanaraps:
I think this approach is better:

```sh
        song=$(cmus-remote -Q | grep "tag artist \|tag title" 2>/dev/null | sort)
        artist="${song##tag artist }"
        artist="${artist%%tag title*}"
        title="${song##*tag title }"
        song="${artist//$'\n'} - $title"
        state=$(cmus-remote -Q | awk -F ' ' '/status/ {printf $2}' 2>/dev/null)
```

- We call `sort` to sort `tag title` and `tag artist` alphabetically.
- We remove the newline using `${artist//$'\n'}` instead of `tr`.

Note: I haven't tested this on OS X.

dylanaraps:
Ok, this is an even better way.

```sh
IFS=$'\n'
song=($(cmus-remote -Q | grep "tag artist \|tag title \|status" 2>/dev/null | sort))
artist=${song[1]/tag artist }
title=${song[2]/tag title }
state=${song[0]/status }

song="$artist - $title"
```

- Uses `sort` and an array.
- Reduce the `cmus-remote` call to once for the entire block.

iandrewt:
Fixed. That's a lot cleaner too!

dylanaraps:
:+1:

PR #272, opened 2016-06-10
Add terminology support

aranega:
Terminology (https://github.com/billiob/terminology) is a modern terminal emulator which provides a lot of features. Currently, terminology is not supported by neofetch because : 

* it does not support the "\033[14t"
* w3m-img does not seems to work properly

However, terminology understands an alternative sequence to "\033[14t" and is shipped with a tool "tycat" which allows you to display an image in your terminal.

Using both, the window size query alternative and tycat as new image_backend, neofetch is now able to deal with terminology.

I tried to match the coding style as much as possible (indentation, ifs and naming rule). I hope this will fit your requirements.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -2277,23 +2277,39 @@ getimage () {
     stty -echo
     if [ -n "$TMUX" ]; then
         printf "%b%s" "\033Ptmux;\033\033[14t\033\033[c\033\\"
+    elif [ "$term" = "terminology" ]; then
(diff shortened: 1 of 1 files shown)


dylanaraps:
This is really cool, thanks for working on this!

I've made some changes and opened a pr here: #273.

dylanaraps:
Closed in favor of #273

PR #284, opened 2016-06-13
Add support for macOS Sierra

iandrewt:
I can't test this yet, but unless the output of `sw_vers -productName` has changed then this will work. Can't test until mid July though... unless someone else has a Apple Developer Program subscription and can test it for me.


dylanaraps:
:+1:

PR #287, opened 2016-06-15
improved refresh_rate fix, now actually works on OS X...

iandrewt:
So the last refresh rate fix didn't really fix anything, in fact it broke resolution without refresh rate all together when using multiple screens... Turns out the refresh rate was never removed at all when you have multiple screens. Now it works perfectly, by looping over the individual refresh rates and removing them there (all still in pure bash)

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -1538,15 +1538,13 @@ getresolution () {
                     awk '/Resolution:/ {printf $2"x"$4" @ "$6"Hz, "}')"
             fi
 
-            if [ "$refresh_rate" == "off" ] || [ "${resolution// * @ }" == "0Hz" ]; then 
-                IFS=","
-                for res in $resolution; do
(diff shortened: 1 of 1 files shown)


dylanaraps:
This seems a little too complicated.

What's the resolution output normally? `1920x1080 @ 60hz, 1280x720 @ 30hz` ?

iandrewt:
I'm all for a simpler solution. The output I get normally in my setup is `1920x1080 @ 60Hz, 1440x900 @ 60Hz, 1440x900 @ 60Hz`

dylanaraps:
This way works with 2 lines:

```sh
resolution=${resolution// @ [0-9][0-9]Hz}
resolution=${resolution// @ [0-9][0-9][0-9]Hz}
```

I'm trying to reduce this to one.

PR #300, opened 2016-08-04
Condense in-script config

dylanaraps:
Removes all comments and formatting from the default config inside of the neofetch
main script. This removes 300~ lines from the script.

- No one should be changing these values so comments aren't necessary.
- The separate config file contains the comments/docs about what each
option does and is how you should be changing the values.


dylanaraps:
#306

PR #302, opened 2016-08-07
Added detection for Unknown OS

konimex:
This commit adds "support" (well, not exactly support, but a fallback detection method if no other OS than what's supported is detected instead of displaying just `Unknown OS detected: $(uname)`)

Some of the features will immediately work (GPU detection, DE/WM, detection, kernel detection etc.), but other features like CPU detection and RAM detection will not appear, as there'll be differences in how other *NIXes handle this.

Tested on: OpenIndiana Hipster

### DIFF
--- a/ascii/distro/others
+++ b/ascii/distro/others
@@ -1,0 +1,9 @@
+"${c1}UUU     UUU NNN      NNN IIIII XXX     XXXX
+UUU     UUU NNNN     NNN  III    XX   xXX
+UUU     UUU NNNNN    NNN  III     XX xXX
+UUU     UUU NNN NN   NNN  III      XXXX
+UUU     UUU NNN  NN  NNN  III      xXX
+UUU     UUU NNN   NN NNN  III     xXXXX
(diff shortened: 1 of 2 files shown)


dylanaraps:
I'd rather keep the current implementation (Exiting with a message for unknown OS) instead of an incomplete general detection method. 

Is it not possible for us to add full support for OpenIndiana Hipster and other *NIXes? As you said a lot of the functions already work, so we'd just have to update the broken ones.

konimex:
While it is possible, many other *NIXes have different implementations of how they detect CPUs, GPU (Xorg can handle it so we have at least one universal solution) and even birthdays (Oracle's Solaris and OpenIndiana have different implentations, and unlike OpenIndiana, Oracle's `date` does not support `-d`.) And that doesn't include other obscure *NIX like HP-UX (and that is if we will ever encounter their users)

If possible, I'll make another PR for Solaris (for now, OpenIndiana only), so I'll close this one.

dylanaraps:
:+1:

PR #303, opened 2016-08-07
Added Solaris support

konimex:
This adds Solaris detection to neofetch.

The only function I can't get is battery detection. This is caused by:
1. No laptop to test it on
2. Neither Oracle nor OpenIndiana has any documentation about it, the latest information I can get is from [2006](https://blogs.oracle.com/DanX/entry/solaris_laptop_battery_status)

virtualbox_openindiana_07_08_2016_15_21_28

(Yes, I used Arch logo because I don't know how to make ASCII logos, especially OpenIndiana.)

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -442,6 +442,7 @@ case "$(uname)" in
     "Darwin")  os="$(sw_vers -productName)" ;;
     *"BSD" | "DragonFly") os="BSD" ;;
     "CYGWIN"*) os="Windows" ;;
+    "SunOS") os="Solaris" ;;
     *) printf "%s\n" "Unknown OS detected: $(uname)"; exit 1 ;;
 esac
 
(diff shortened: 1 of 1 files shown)


neofetch
```diff
@@ -511,6 +512,10 @@ getdistro () {
             distro="${distro/Caption'='}"
             distro="${distro/Microsoft }"
         ;;
+
+        "Solaris")
+            distro="$(cat /etc/release | head -1 | awk ' {print $1 " " $2} ')"
```
dylanaraps: You can probably remove both pipes here with a line similar to linux's /etc/release part.

```sh
awk -F 'NAME=' '/^NAME=/ {printf $2}' /etc/*ease
```

dylanaraps:
Apart from my comment, LGTM :+1:

I'll start working on some logos, which Solaris distros should I make logos for?

konimex:
There are only two active Solaris distributions as of now, Oracle and OpenIndiana. Unless you want to include the old Sun OpenSolaris' logo.

dylanaraps:
Awesome, I'll make ascii art for those two distros.

dylanaraps:
LGTM

PR #306, opened 2016-08-11
Remove default config from inside script and instead source it from the user config file.

dylanaraps:
This PR removes the in-script config and instead sources the user config file. Both configs are identical, this  allows us to remove 400~ lines from neofetch and will make it easier to edit the config later on.

The config sourcing works the same way the user config sourcing works and I haven't heard about any issues with that method.

This PR also includes:

- Travis changes and fixes
- Added missing `ascii_distro` config option
- Fixed `err()` bugs
- Config file verbose output
- Use shorthand ifs near the bottom of the script
- Move the commands near the bottom of the script to a function called `main`
- Cleanup

TODO:

- [ ] Testing

Closes #300


dylanaraps:
Can anyone who stumbles across this PR test it out for me?  Just running neofetch
from this branch and letting me know if it works is enough. :+1:

konimex:
Tested on FreeBSD and OpenIndiana. Both works fine.

PR #308, opened 2016-08-14
Add 'model' function to print the device name

dylanaraps:
This PR adds a new function which will be enabled by default. The function displays the product/model name of your device if it exists. For example on my laptop it shows `Lenovo YOGA 900-13ISK` and my iPhone it shows `iPhone 5s`.

Note: This feature won't work on custom built PCs or devices without OEM info.

**TODO:**

- [ ] Add support
    - [x] Add support for Linux
        - Not sure if this works on every machine yet.
    - [ ] Add support for OS X
    - [x] Add support for iOS
    - [ ] Add support for BSD
    - [ ] Add support for Solaris
    - [ ] Add support for Windows
- [ ] Testing
    - [ ] Test the current implementation on various Linux systems
    - [x] Test the feature on iOS

**Testing:**

If you'd like to help me test this is what you need to do:

1. Git clone neofetch.
    - `git clone https://github.com/dylanaraps/neofetch`
2. cd to the new directory.
    - `cd neofetch`
2. Swap to the `model` branch. 
    - `git checkout model`
3. Run neofetch without a config from the directory.
    - `./neofetch --config off`
4. Let me know if the feature works/doesn't work on your machine.
5. Screenshots are nice too. :)

### DIFF
--- a/config/config
+++ b/config/config
@@ -18,6 +18,7 @@ printinfo () {
     info title
     info underline
 
+    info "Model" model
     info "OS" distro
     info "Kernel" kernel
     info "Uptime" uptime
--- a/neofetch
+++ b/neofetch
@@ -36,6 +36,57 @@ getos() {
 
 # }}}
 
+# Model {{{
+
+getmodel() {
+    case "$os" in
+        "Linux")
+            model="$(< /sys/devices/virtual/dmi/id/product_version)"
+            model="${model/To Be Filled*}"
+        ;;
+
+        "iPhone OS")
+            case "${ios_model:-$(uname -m)}" in
+                "iPad1,1") model="iPad" ;;
+                "iPad2,"[1-4]) model="iPad2" ;;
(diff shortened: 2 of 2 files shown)


iandrewt:
Adding Mac support for this, should I have the model print as `iMac13,1` or `iMac (21.5 inch, Late 2012)`? Both are technically correct, and I could even add an flag for choosing which to print (such as `--mac-model [model|name|modelname]`)

dylanaraps:
I'd go with the second one, it looks nicer imo.

iandrewt:
After doing some more research, it looks like I'll have to do a `curl` to http://support-sp.apple.com to get the prettier name as OS X finds it by running the last 3 characters of the serial number to that address and returns an XML file. Every single time you go to System Profiler to see it...
I can cache the result as it will never change. Are you still okay with doing the pretty method?

dylanaraps:
Lets go with the ugly method.

I've fixed the config issues and opened a new PR here: #310

PR #310, opened 2016-08-14
Add new `model` function to print device model name

dylanaraps:
This PR adds a new function which will be enabled by default. The function displays the product/model name of your device if it exists. For example on my laptop it shows `Lenovo YOGA 900-13ISK` and my iPhone it shows `iPhone 5s`.

Note: This feature won't work on custom built PCs or devices without OEM info.

**TODO:**

- [x] Add support
    - [x] Add support for Linux
        - ~~Not sure if this works on every machine yet.~~
        - The file we check for the info was added in `linux 2.6.26 ~2009`<br \>
           so this should be fine.
    - [x] Add support for MacOS (OS X)
    - [x] Add support for iOS
    - [x] Add support for BSD
    - [x] Add support for Solaris
    - [x] Add support for Windows
- [x] Testing
    - [x] Test the feature on Linux
    - [x] Test the feature on MacOS (OS X)
    - [x] Test the feature on iOS
    - [x] Test the feature on BSD
    - [x] Test the feature on Solaris
    - [x] Test the feature on Windows

**Testing:**

If you'd like to help me test this is what you need to do:

1. Git clone neofetch.
    - `git clone https://github.com/dylanaraps/neofetch`
2. cd to the new directory.
    - `cd neofetch`
2. Swap to the `model` branch. 
    - `git checkout model`
3. Run neofetch without a config from the directory.
    - `./neofetch --config off`
4. Let me know if the feature works/doesn't work on your machine.
5. Screenshots are nice too. :)

Old PR #308

### DIFF
--- a/config/config
+++ b/config/config
@@ -18,6 +18,7 @@ printinfo () {
     info title
     info underline
 
+    info "Model" model
     info "OS" distro
     info "Kernel" kernel
     info "Uptime" uptime
--- a/neofetch
+++ b/neofetch
@@ -36,6 +36,57 @@ getos() {
 
 # }}}
 
+# Model {{{
+
+getmodel() {
+    case "$os" in
+        "Linux")
+            model="$(< /sys/devices/virtual/dmi/id/product_version)"
+            model="${model/To Be Filled*}"
+        ;;
+
+        "iPhone OS")
+            case "${ios_model:-$(uname -m)}" in
+                "iPad1,1") model="iPad" ;;
+                "iPad2,"[1-4]) model="iPad2" ;;
+                "iPad3,"[1-3]) model="iPad3" ;;
+                "iPad3,"[4-6]) model="iPad4" ;;
+                "iPad4,"[1-3]) model="iPad Air" ;;
+                "iPad5,"[3-4]) model="iPad Air 2" ;;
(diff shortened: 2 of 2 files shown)


dylanaraps:
I've added support for Windows. I don't have any prebuilt PCs at home so I'm
not 100% sure that this works without issues. If someone could test this for me
I'd be very thankful. :)

dylanaraps:
On BSD systems we can use `dmidecode` to get the information, the only issue
is that the command requires root privileges. I'm trying to find an alternative
command/program to use.

iandrewt:
Works as expected in my Arch VM, model prints `Model: VirtualBox 1.2`

dylanaraps:
This is ready to merge,  I would've loved to do some more testing but I don't own anymore OEM devices.

dawidd6:
Okay, i've done some research on RPi2 and my router. Both seems to not have these directories which neofetch is using.

1. RPi2 with OSMC (debian jessie)
`/sys/firmware/devicetree/base/model`
>correctly showing Raspberry Pi 2B...

2. TP-LINK with Gargoyle (openwrt aa)
`/tmp/sysinfo/model`
>correct TP-Link TL-WR740N/ND v4

dylanaraps:
Thanks for testing!

What I'll do is add fallbacks to these directories in another PR. :+1:

PR #311, opened 2016-08-14
added macOS support for model name

iandrewt:
This PR adds support for macOS in the new model function


dylanaraps:
You can use `sysctl -n` and then get rid of the 2nd line iirc.

dylanaraps:
Thanks :+1:

PR #312, opened 2016-08-14
fixed macOS codename function

iandrewt:
Fixes the codename when there is no point-release, and fixes Sierra support (ascii was broken)


dylanaraps:
:+1:

PR #313, opened 2016-08-14
Added Solaris support for model

konimex:
This PR adds Solaris support for model function.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -89,6 +89,10 @@ getmodel() {
             model="${model/Manufacturer'='}"
             model="${model/Model'='}"
         ;;
+
+        "Solaris")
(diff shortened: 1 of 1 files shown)


dylanaraps:
Thanks! :+1:

Can I see a screenshot?

konimex:
It seems VirtualBox isn't recognized as a "prebuilt computer" so it won't generate any output. But looking from [Oracle's documentation](https://docs.oracle.com/cd/E23824_01/html/821-1451/gkkwk.html#gaefk), I guess the output will be pretty similar if I had a testbed laptop.

dylanaraps:
Yeah, I tried it out in a VM and the output was blank. I'm going to reinstall linux on my laptop soon so I'll install a Solaris distro to test for a little bit.

PR #321, opened 2016-08-27
Added support for GuixSD

konimex:
Added support for GuixSD, this includes:

- Added detection for Guix package manager
- New ASCII for GuixSD (I'm sure this can be reworked.)


konimex:
Actually, don't merge it just yet -- I'll reopen this when I figure out how to detect GuixSD version (they don't conform to the usual standards)

konimex:
Since there's no other way to detect if a distro is GuixSD other than detecting if `guix` command exists, I think this will do.

Unless you have anything more to say, then I'll squash this up before ready to merge.

PR #324, opened 2016-08-29
Added and changed some Android commands

konimex:
Included:
- Added model support for Android
- Changed CPU command to use native Android ones

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -41,7 +41,10 @@ getos() {
 getmodel() {
     case "$os" in
         "Linux")
-            if [ -f /sys/devices/virtual/dmi/id/product_name ] ||\
+            if [ -d "/system/app/" ] && [ -d "/system/priv-app" ]; then
+                model="$(getprop ro.product.brand) $(getprop ro.product.model)"
+
+            elif [ -f /sys/devices/virtual/dmi/id/product_name ] ||\
                [ -f /sys/devices/virtual/dmi/id/product_version ]; then
                 model="$(cat /sys/devices/virtual/dmi/id/product_{name,version} 2>/dev/null)"
                 model="${model/To Be Filled*}"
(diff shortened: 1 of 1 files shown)


neofetch
```diff
@@ -41,7 +41,10 @@ getos() {
 getmodel() {
     case "$os" in
         "Linux")
-            if [ -f /sys/devices/virtual/dmi/id/product_name ] ||\
+            if [ -d "/system/app/" ] && [ -d "/system/priv-app" ]; then
+                model="$(getprop ro.product.brand) $(getprop ro.product.model)"
```
dylanaraps: Is it possible to combine these two commands?

```sh
$(getprop ro.product.brand ro.product.model)
```

neofetch
```diff
@@ -640,7 +643,12 @@ getcpu() {
     case "$os" in
         "Linux" | "Windows")
             # Get cpu name
-            cpu="$(awk -F ': | @' '/model name|Processor/ {printf $2; exit}' /proc/cpuinfo)"
+            case "$distro" in
```
dylanaraps: Change the case to this:

```sh
                "Android"*) cpu="$(getprop ro.product.board)" ;;
                *) cpu="$(awk -F ': | @' '/model name|Processor/ {printf $2; exit}' /proc/cpuinfo)" ;;
```

or

```sh
                "Android"*)
                    cpu="$(getprop ro.product.board)" 
                ;;

                *)
                    cpu="$(awk -F ': | @' '/model name|Processor/ {printf $2; exit}' /proc/cpuinfo)"
                ;;
```

dylanaraps:
Apart from the comments, LGTM.

neofetch
konimex: Unfortunately, no. `getprop` doesn't support it. We can read those two directly from `/system/build.prop`, though.

dylanaraps: Alright, we'll just leave it like this then. :+1:

PR #328, opened 2016-09-02
OpenBSD's battery status get back into the BSD case

konimex:
Maybe the original committer forgot the `*` so s/he decided to get OpenBSD's case nest into the `$os` instead of `$distro`.

I suspect the original reason that because OpenBSD's `$distro` output is `OpenBSD $(uname -m)` instead of just `OpenBSD` so the results didn't appear when s/he first tested it out.

Anyway, for consistency, I decided to bring OpenBSD's battery case back to the `$distro` case and add a very little fix (which is the `*`), the problem should be fixed.

(I can't test this personally, any help would be appreciated.)


dylanaraps:
LGTM :+1:

PR #329, opened 2016-09-03
Fixed percent not displaying in cpu_display

konimex:
`cpu_display` won't appear with `%` if `cpu_display="off"` is set. This PR fixes it.

In README.md (and in the man pages), the 4 accepted modes are `off`, `barinfo`, `infobar` and `bar`. Though I'm sure we can have a fall-back mode (which is `off`) if users don't set `cpu_display`to one of those 4 values.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -919,10 +919,10 @@ getcpu_usage() {
 
     # Print the bar
     case "$cpu_display" in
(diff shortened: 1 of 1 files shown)


dylanaraps:
:+1:

PR #334, opened 2016-09-18
Initial support for Windows 10 Linux subsystem

konimex:
I actually tried this back in August, but somehow I dropped it.

Anyway, I picked this up again and decided to test around.

Bugs:
- [x] ~~`dynamicprompts` does not work (only occurred on Windows 10's terminal, in a X terminal inside Windows 10, this doesn't happen)~~ (Fixed with #339 )
- [x] ~~There are some.. artifacts after the output (only occurred on Windows 10's terminal, in a X terminal inside Windows 10, this doesn't happen). I suspect these came from the earlier `info_heights`.~~ (Fixed with #339 )
- [ ] CPU ~~Usage~~ clock speed doesn't have any output (0.0GHz) by default
 - Windows 10's `cpufreq` only has `scaling_max_freq` and `cpuinfo_max_freq`. 
 - You cannot read `scaling_max_freq` (insufficient permissions), only `cpuinfo_max_freq`.
   - Possible workaround: Change the default in config to `cpuinfo_max_freq`.

Other than those, it seems to work perfectly. Even in an X session.

virtualbox_windows 10_18_09_2016_17_30_04
virtualbox_windows 10_18_09_2016_17_30_31

I can only test this from a VM, so if anyone wants to test it, be my guest.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -118,7 +118,16 @@ getdistro() {
 
     case "$os" in
         "Linux" )
-            if type -p lsb_release >/dev/null 2>&1; then
+            if grep -q 'Microsoft' /proc/version >/dev/null 2>&1 || \
+               grep -q 'Microsoft' /proc/sys/kernel/osrelease >/dev/null 2>&1; then
+                case "$distro_shorthand" in
+                    "on")   distro="$(lsb_release -sir 2>/dev/null) [Windows 10]" ;;
+                    "tiny") distro="Windows 10" ;;
+                    *)      distro="$(lsb_release -sd 2>/dev/null) on Windows 10" ;;
+                esac
+                ascii_distro="windows10"
+
+            elif type -p lsb_release >/dev/null 2>&1; then
                 case "$distro_shorthand" in
                     "on")   distro="$(lsb_release -sir 2>/dev/null)" ;;
                     "tiny") distro="$(lsb_release -si 2>/dev/null)" ;;


neofetch
```diff
@@ -2772,6 +2781,14 @@ dynamicprompt() {
     # Get cursor position
     info_height="$(IFS=';' builtin read -srdR -t 1 -d c -p $'\033[6n\033[c' ROW COL; printf "%s" "${ROW#*[}")"
 
+    # On Windows 10 subsystem for Linux, the default terminal does not
+    # produce any integer output, so a workaround is used.
+    if [[ $info_height =~ ^-?[0-9]+$ ]]; then
+        :
+    else
+        info_height="0"
+    fi
```
konimex: For some reason I really don't like the workaround I made. I guess a more effective way can be made.

dylanaraps:
Awesome!

I'll take a closer look at this when I get back.

neofetch
```diff
@@ -2772,6 +2781,14 @@ dynamicprompt() {
     # Get cursor position
     info_height="$(IFS=';' builtin read -srdR -t 1 -d c -p $'\033[6n\033[c' ROW COL; printf "%s" "${ROW#*[}")"
 
+    # On Windows 10 subsystem for Linux, the default terminal does not
+    # produce any integer output, so a workaround is used.
+    if [[ $info_height =~ ^-?[0-9]+$ ]]; then
```
konimex: Personally, I don't like this approach. A better solution like `[ -z $info_height ]` is considered but Windows' output of `$info_height` is not empty either so it can't be done. So for now I'll just use the one that works.

dylanaraps: What's the output? What you could do is use a case statement instead, so something like this:

```sh
case "$info_height" in
    "-"?[0-9]*) ;;
    *) info_height="0" ;;
esac
```

konimex: Something like this

virtualbox_windows 10_21_09_2016_10_09_15

dylanaraps:
I'm setting up Windows 10 to take a closer look, hopefully we can get this fully working and merge it today. :)

dylanaraps:
I'm going to hardcode the cpu speed to `cpuinfo_max_freq` on Windows 10 (Bash). I'll merge this as is and open a separate PR.

dylanaraps:
We need to also fix ascii color.

dylanaraps:
Fixed CPU speed and ascii color.

PR #337, opened 2016-09-30
Fixed os-release for Arch (#336)

### LINKED ISSUE #336: neofetch cannot find arch linux os-release
Ever since [this](https://git.archlinux.org/svntogit/packages.git/commit/trunk/PKGBUILD?h=packages/filesystem&id=66553654434fbe2c3c30c0fa3f96e87bba79e3f4) commit the Arch Linux filesystem has installed os-release to /usr/lib/ instead of /etc/ (where neofetch looks for the release file).

While script will function for users who installed the Arch before 2014-07, any later installations will cause the script to fail to find the file.

konimex:
This should fix #336.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -140,17 +140,20 @@ getdistro() {
                 case "$distro_shorthand" in
                     "on")
                         distro="$(awk -F'=' '/^NAME|VERSION_ID=/ {print $2}' /etc/*ease)"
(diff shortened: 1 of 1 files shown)


dylanaraps:
Does this work?

```
awk -F'=' '/^NAME|VERSION_ID=/ {print $2}' /etc/*ease /usr/lib/*ease
```

dylanaraps:
Merged #338, thanks for the PR!

PR #341, opened 2016-10-01
Update Solus ASCII logo to a more recent one

undrskr:
neofetch just got shipped to Solus, and because it changed the logo not a long time ago, it would be good to update it for the relevancy.

### DIFF
--- a/ascii/distro/solus
+++ b/ascii/distro/solus
@@ -1,20 +1,23 @@
 "\
-${c1}              e         e
-            eee       ee
-           eeee     eee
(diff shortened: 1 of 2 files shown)


dylanaraps:
Thanks! :+1:

Is there anything we should add to the readme regarding installation on Solus?

undrskr:
@dylanaraps I don't think there is much to say, just the basic "Use the Software Center or type `sudo eopkg it neofetch`."

Also, I screwed up with the size a little, making logo a bit smaller now.

ikeydoherty:
Thanks for this :)

PR #343, opened 2016-10-02
Fixed various indentation

konimex:
A pretty trivial one. Just fixed some indentations (3 spaces instead of 4, something like that).


dylanaraps:
:+1:

PR #349, opened 2016-10-03
Add a tiny sleep to fix image flashing and not drawing properly

### LINKED ISSUE #346: neofetch doesn't display wallpaper
## Description

neofetch doesn't get my wallpaper correctly from gsettings and displays the ascii art. I don't have feh nor nitrogen installed and the `gsettings get org.gnome.desktop.background picture-uri` command returns the path to the wallpaper correctly.

#### Neofetch version
1.8 (but had the same problem with the previous one)
#### Operating system
Arch Linux x86_64
## Screenshot
[http://i.imgur.com/dJK2jV2.png](url)
## Verbose log
http://pastebin.com/KbFvsveq

dylanaraps:
This fixes the image not appearing when the thumbnail already exists, I'm not sure as to *why* this works but it seems to be related to timing between the thumbnail being created and w3m-img displaying it.

I've only tested this on my laptop and the only terminals I found to be affected were `Gnome-terminal` and `Xfce4-terminal`. Please test this out and let me know if it works or at least reduces the frequency of the bug.

This isn't a very pretty fix but if it works it'll have to do until we can find a better way to solve this problem.

@killmaster

Closes #346


killmaster:
This fixes it on gnome-terminal and termite.

zar0us:
paste to? @dylanaraps @killmaster

PR #353, opened 2016-10-05
Fix TravisCI for macOS build

iandrewt:
Travis updated the macOS build server to El Capitan, which means it has SIP now. Prefix changed to /usr/local to mitigate this


dylanaraps:
:+1:

PR #365, opened 2016-10-15
change all references of Mac OS X to macOS

iandrewt:
This does not change any programatical references as `sw_vers` still prints `Mac OS X` in Sierra


dylanaraps:
:+1:

PR #370, opened 2016-10-16
netbsd specific stuff for grabbing memory size

coypoop:


dylanaraps:
If `sysctl -n hw.physmem` is broken on 64bit systems then doesn't `sysctl -n hw.physmem64` break the output on 32bit systems? Or am I totally wrong here? 

Some more info would be nice. :)

coypoop:
sorry for not explaining more,
with 4GB RAM, the hw.physmem signed integer (32bit, not 64bit) is too small:

% sysctl hw.physmem
hw.physmem = -143978496
% sysctl hw.physmem64
hw.physmem64 = 4150988800

from sysctl(7) for netbsd:
     hw.physmem (HW_PHYSMEM)
             The bytes of physical memory as a 32-bit integer.
     hw.physmem64 (HW_PHYSMEM64)
             The bytes of physical memory as a 64-bit integer.

dylanaraps:
Thanks for clearing that up.

LGTM :+1:

PR #372, opened 2016-10-17
macOS: fix iTunes automatically opening

iandrewt:
Fixes iTunes opening if it's not running on macOS. I believe this is because of a new cache extension in Sierra, because I'm sure I fixed this bug before...


dylanaraps:
:+1:

PR #376, opened 2016-10-17
Condense osx_version

dylanaraps:
This reduces the size of the case statement.


iandrewt:
Not working for me unless I change the number in the second block to be `[8-10]`, prints `OS X` for me on `macOS Sierra`

dylanaraps:
Fixed.

iandrewt:
👍

PR #380, opened 2016-10-20
Rewrite uptime for Linux/Windows/Solaris to match macOS/BSD

dylanaraps:
This PR aims to remove issues on Linux based systems where there are an endless amount of different `uptime` commands all with a different output format. Instead of using the `uptime` command we now parse `/proc/uptime` which provides us with the uptime in seconds. We then convert those seconds into days, hours and minutes.

This is how neofetch calculates uptime on macOS/BSD and this allows us to merge both functions removing ~20 lines from the script. 

Notes:

- This removes the `procps`/`procps-ng` dependency from Windows systems.
- This fixes output errors on Android.
- This allows us to add an option to enable seconds in the output. I don't know if I'll add this or not but the possibility is there.

Testing:

- [x] Linux
- [x] Windows
- [x] Solaris
- [x] BSD
- [ ] macOS / iOS 

@iandrewt, can you test this on macOS?

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -288,9 +288,9 @@ getuptime() {
                 ;;
             esac
 
-            days="$((seconds / 86400)) days"
-            hours="$((seconds / 3600)) hours"
-            minutes="$((seconds % 3600 / 60)) minutes"
+            days="$((seconds / 60 / 60 / 24)) days"
+            hours="$((seconds / 60 / 60 % 24)) hours"
+            minutes="$((seconds / 60 % 60)) minutes"
 
             case "$days" in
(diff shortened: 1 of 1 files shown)


dylanaraps:
Bug(?) on Windows

```sh
Uptime: up 15 days, 361 hours, 37 minutes
```

dylanaraps:
This seems to be broken on macOS travis:

https://travis-ci.org/dylanaraps/neofetch/jobs/169390092

dylanaraps:
It works on BSD so I'm going to assume it works on macOS. 

If any issues arise we can fix them afterwards.

PR #393, opened 2016-10-23
Solaris: Changed Solaris' getpackage behaviour and getdistro output

konimex:
Some Solaris/illumos-based distribution used various Linux package managers such as dpkg, rpm, and portage (see more [here](http://wiki.illumos.org/display/illumos/Distributions)). It would make sense for Solaris' case to be unified with Linux.

However, Linux's `pkg` detection (I don't know which distro is that) are vastly different from Oracle Solaris/OpenIndiana's `pkg`, so a workaround (if acceptable) is used.


dylanaraps:
:+1:

PR #394, opened 2016-10-24
*SUSE: Changed ASCII behavior

konimex:
In *SUSE's `lsb_release`, they have different behavior for `-ir` and
`-d`, one output will be `openSUSE Leap 42.1` and the other will be
`SUSE LINUX 42.1`. This new behavior would also extend compatibility
with SUSE Linux Enterprise.

virtualbox_opensuse_24_10_2016_21_58_21

My apologies for the weird resolution. VirtualBox's Guest Addition is a weird thing.


dylanaraps:
:+1:

PR #401, opened 2016-10-30
Kernel: Removed `-m` from `uname` in kernel_shorthand's off option

konimex:
### Features
The reason that I removed it from the beacuse we already have `os_arch` option in `getdistro`. So I don't see the point in adding another architecture option in `getkernel`.


dylanaraps:
Merged manually

dylanaraps:
Thanks

PR #402, opened 2016-10-30
GPU: Used FreeBSD's method for DragonFlyBSD and PacBSD

konimex:
### Features

Since DragonFlyBSD has `pciconf`, and PacBSD is basically FreeBSD + `pacman`, we can use `pciconf` for those two distros instead of relying on Xorg.


dylanaraps:
:+1:

PR #403, opened 2016-10-31
Expanded distro_shorthand to *BSDs

konimex:
### Features
See title.

### Issues
Due to the monolithic nature of BSDs (Up until now, I still have not heard of a FreeBSD distro uses an OpenBSD/NetBSD kernel), the `uname -r` output is effectively the OS's version. Because of that, how `getkernel` works in BSDs will have to be altered significantly (meaning `kernel_shorthand` won't work in BSDs). Since the code is self-explanatory enough I won't dig deeper into this.

PacBSD and PC-BSD are exceptions. Since PacBSD and PC-BSD's `uname -s` output is `FreeBSD`, `kernel_shorthand` may be used normally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -448,8 +448,9 @@ alias neofetch2="neofetch \
     --cpu_temp on/off           Hide/Show cpu temperature.
                                 NOTE This only works on linux.
     --distro_shorthand on/off   Shorten the output of distro (tiny, on, off)
-                                NOTE: This is only possible on Linux, macOS, and Solaris
+                                NOTE: This option won't work in Windows (Cygwin)
     --kernel_shorthand on/off   Shorten the output of kernel
+                                NOTE: This option won't work in BSDs (except PacBSD and PC-BSD)
     --uptime_shorthand on/off   Shorten the output of uptime (tiny, on, off)
     --refresh_rate on/off       Whether to display the refresh rate of each monitor
                                 Unsupported on Windows
--- a/neofetch
+++ b/neofetch
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 # set -x
-# vim: fdm=marker:noai:ts=4:sw=4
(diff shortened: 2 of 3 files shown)


dylanaraps:
The indentation was a bit off:

```sh
    # Hardcode kernel settings in BSDs
    case "$os" in
        "BSD")
            case "$distro" in
                "PacBSD"* | "PCBSD"*) kernel="$(uname $kernel_flags)" ;;
                *)
                    case "$distro_shorthand" in
                        "on" | "tiny") kernel=$(uname -r) ;;
                        *) unset kernel ;;
                    esac
                ;;
            esac
        ;;
        *) kernel="$(uname $kernel_flags)" ;;
    esac
```

dylanaraps:
I'm pretty sure that this will also work:

 ```sh
getkernel() {
    case "$kernel_shorthand" in
         "on")   kernel_flags="-r" ;;
         "off")  kernel_flags="-sr" ;;
    esac
    
    # Hardcode kernel_shorthand on BSD
    case "$distro" in
        "PacBSD"* | "PCBSD"*) kernel="$(uname $kernel_flags)" ;;
        *"BSD") kernel="$(uname -r)" ;;
        *) kernel="$(uname $kernel_flags)" ;;
    esac
 }
```

konimex:
>`*"BSD") kernel="$(uname -r)" ;;`

The problem with this one is since in pretty much every *BSD (except the two I mentioned earlier) kernel version == OS version, I thought it would be overkill to add another `uname -r` where `uname -r` is printed twice since the default behaviour of `distro_shorthand` is off. But as usual, your call.

Also, I've added one unrelated command (line 3, about vim) to add `expandtab` to force one ~~glorious~~ tab as 4 spaces.

dylanaraps:
Alright, understood.

LGTM :+1:

PR #412, opened 2016-11-02
Android (Termux): Fix default config location

fornwall:
Without this change I get the following error when running `neofetch` after a clean install:

    neofetch: line 3261: printinfo: command not found

After applying this change it works beautifully!


dylanaraps:
:+1:

PR #413, opened 2016-11-02
Fixed used memory total for OpenBSD

Head-on-a-Stick:


dylanaraps:
I'm a little confused about why you made this change. 

```sh
sysctl -n hw.physmem
```

The command above prints the total amount of memory in the system and is what neofetch currently uses. I just tested this command and the new `vmstat` command in an OpenBSD VM. The `sysctl` command prints the correct memory and the `vmstat` command doesn't work at all (?).

dylanaraps:
I did some googling and found that the `sysctl` command above displays the wrong value on some systems. I also saw suggestions to use `vmstat` instead like you've done in this PR. I'll mess around with `vmstat` and see if I can get your command to work.

dylanaraps:
Your command doesn't work on my system, does this work on yours:

```sh
echo "$(($(vmstat | awk 'END {print $4}') / 1024))"
```

This is what neofetch currently uses:

```sh
echo "$(($(vmstat | awk 'END {print $5}') / 1024))"
```

dylanaraps:
Made some changes and merged this.

Thanks!

PR #414, opened 2016-11-03
Added initial detection for GNU (Hurd) systems

konimex:
### Features

See title.

### Issues

- `getcpu` can't be done in GNU Hurd because they have no way of detecting CPU names (no `/proc/cpuinfo`.
- There are currently no ways to fetch `getmodel` and `getbattery`, because of different kernel systems, and there are currently no documentations about both of them.

P.S.: Should you decide to merge this branch, I advise to wait until you released 1.9.x, as an 1.9.x should be for bugfixes only, IMO.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -30,6 +30,7 @@ getos() {
         *"BSD" | "DragonFly" | "Bitrig") os="BSD" ;;
         "CYGWIN"*) os="Windows" ;;
         "SunOS") os="Solaris" ;;
+        "GNU"*) os="GNU" ;;
         *) printf "%s\n" "Unknown OS detected: $(uname)"; exit 1 ;;
     esac
 }
@@ -122,7 +123,7 @@ getdistro() {
     [ "$distro" ] && return
 
     case "$os" in
-        "Linux" )
+        "Linux" | "GNU")
             if grep -q 'Microsoft' /proc/version >/dev/null 2>&1 || \
                grep -q 'Microsoft' /proc/sys/kernel/osrelease >/dev/null 2>&1; then
                 case "$distro_shorthand" in
@@ -305,7 +306,7 @@ getkernel() {
 getuptime() {
     # Get uptime in seconds
     case "$os" in
-        "Linux" | "Windows")
+        "Linux" | "Windows" | "GNU")
             seconds="$(< /proc/uptime)"
             seconds="${seconds/.*}"
         ;;
@@ -376,7 +377,7 @@ getuptime() {
 
 getpackages() {
     case "$os" in
-        "Linux" | "iPhone OS" | "Solaris")
+        "Linux" | "iPhone OS" | "Solaris" | "GNU")
             type -p pacman >/dev/null 2>&1 && \
                 packages="$(pacman -Qq --color never | wc -l)"
 
@@ -899,7 +900,7 @@ getcpu_usage() {
             cpu_usage="${cpu_usage//[[:space:]]}"
         ;;
 
-        "Linux" | "Mac OS X" | "iPhone OS" | "BSD" | "Solaris")
+        "Linux" | "Mac OS X" | "iPhone OS" | "BSD" | "Solaris" | "GNU")
             # Get cores if unset
             if [ -z "$cores" ]; then
                 case "$os" in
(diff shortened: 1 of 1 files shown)


dylanaraps:
Looking good so far :+1: 

> P.S.: Should you decide to merge this branch, I advise to wait until you released 1.9.x, as an 1.9.x should be for bugfixes only, IMO.

Yup, that's what I'm doing with the other open PR.

konimex:
Anyway, since `getcpu` is effectively empty on GNU Hurd systems, should we add a placeholder for it (e.g. screenfetch uses "Unknown" when dealing with GNU Hurd's `getcpu` function)?

dylanaraps:
The only function which has a placeholder right now is `getsong` which shows `Not Playing`. The rest of the functions just don't print anything if the info isn't found.

I think it's better this way, what do you think?

dylanaraps:
We can easily make every function print `Unknown` if we decide to go down that route. 

We just have to change this line (2312):

```sh
    [ -z "$output" ] && return
```

to this:

```sh
    [ -z "$output" ] && output="Unknown"
```

konimex:
Nope. Some functions better be empty if there's no output.

dylanaraps:
Sorry, what do you mean?

konimex:
Uh.. Please disregard my previous comment. I'll give my thought after I get home.

konimex:
>I think it's better this way, what do you think?
Actually, I have to say yes. It would be awkward if `Unknown` appears in `getde` while you are not using a DE if we decided to go with the "Unknown" route. (I did some preliminary test using your method in your second comment and using the default config).

dylanaraps:
We *can* add add `Unknown` to the output on a per function basis by adding this line to the bottom of specific functions.

```sh
# Substitute memory for function name - get
[ -z "$memory" ] memory="Unknown"
```



Another option is adding a message to `neofetch -v` letting the user know that we failed to detect info. 

We just have to change this line in `info()`

```sh
    [ -z "$output" ] && return
```

to this:

```sh
    [ -z "$output" ] && err "Info: Couldn't detect $subtitle"
```

dylanaraps:
> EDIT: Yes, I do realize that people can change their configs. But from /r/unixporn, some (if not most) people don't use it, iirc.

This is why it's essential to have the best default config possible. It's also why it's nicer to hide errors from the user where possible.

konimex:
>`[ -z "$output" ] && err "Info: Couldn't detect $subtitle"`

I think this one is fine.

dylanaraps:
I created a new branch called `2.0` in case we need to release a 1.9.2.

Also feel free to open any WIP PRs here. :)

konimex:
So.. Should I close this PR and reopen this to the `2.0` branch?

dylanaraps:
I can change it from here.

dylanaraps:
Done.

dylanaraps:
Here's an ascii logo I made for hurd:

```sh
"\
${c1}            -////////.
            yh::::::ds-.
        \`\`-:ds      yNN+::-\`\`
     \`+yso//ds      yNNo//+osyyo-
    \`d+     yh::::::ds..       -oh+
  ..oh-.    .:::yd:::.            sd\`
-/oNNNNo/-      /d      -////////- +h
yh::::::hs      \`d+   ..sh::::::ds  m:
ys      ys       \`sy+:+NNy      ys  yo
ys      ys         \`-:oNNy      ys  yo
yh//////hs            ..sd//////ds \`m:
.:+mNNm+:.              .:::dy:::. sy
  .-sy..    -::::::::.      d+   .hy\`
    +h      yh::::::hs     \`m-./yy-
    .N.     ys      yh//++shNyo:\`
     sy     ys      yh:--.-m-
      yy\`   yh//////ds   -d/
       +d+. -::::::::. -yy.
         :syso+/::/+oyyo.
             .--::-.\`
"
```

Source: https://www.gnu.org/software/hurd/logo/boxes-redrawn.svg

konimex:
I thought ASCII logos are part of the distros, and so far the only existing Hurd system is Debian (and to some extent, GuixSD), might be good for fallback ASCII though.

Speaking of fallback ASCII, why don't we add a fallback ASCII for Linux (showing Tux, of course) in case no distro ASCII is found?

dylanaraps:
Oh, you're right. My bad.

I'll look into adding a tux ascii art. Good idea.

dylanaraps:
Merged, thanks.

https://github.com/dylanaraps/neofetch/commit/03aebe277c8ba2081f648c50dcfcbde7d3a9b822

Issue #416, opened 2016-11-04
Idea: Remove info() in favor of prin()


dylanaraps:
I had the idea of removing the `info()` function entirely and instead utilizing the `prin()` function. This would simplify the `printinfo()` function in the config file and also remove the `eval` command we call every time we run `info`. 

It would work by adding a new line to the bottom of every info function.

```sh
# $1 will be the subtitle from the user config
# If $1 is empty we can fallback to the default title.
prin "${1:-Memory}" "$memory"
```

The `printinfo()` function would then look like this:

```sh
# First column is function name.
# Second column is the subtitle.
# If the second column is empty we can use a default title as seen above.
printinfo() {
    gettitle
    getunderline
    getmodel  "Model"
    getdistro "OS"
    getkernel "Kernel"
    # ...
}
```

I also had another idea, we could remove the `get` from the function names entirely, it would then look like this:

```sh
printinfo() {
    title
    underline
    model  "Model"
    distro "OS"
    kernel "Kernel"
    # ...
}
```

I think that this looks a lot cleaner and will make it easier for users to modify.

TL;DR

- Remove `info()`
- Add a new line to every info function to call `prin()`
- Remove `get` from all function names

Let me know what you guys think about this!

iandrewt:
👍 Sounds good to me! Would this also make it faster, too? Seeing as we wouldn't be double calling and using `eval`

dylanaraps:
I've finished implementing this in a separate branch and there are a few issues we have to resolve. I just got home so I'll open the PR now.

dylanaraps:
See #417

dylanaraps:
I decide against merging this since it will break every neofetch config file next release. What I'll do instead is have a look at how we can speedup `info()`.

PR #417, opened 2016-11-04
Remove info() in favor of prin()

dylanaraps:
## Description

See #416 

## Features

- Remove `info()`.
- Remove `get` from all function names.
- Fixed images in `st`.
- We no longer need to litter the script with `2>/dev/null` and `2>&1`.
- Neofetch is now 50-100ms faster in all modes.

## Issues

- [x] `term` is broken.
    - This will need a rewrite to use a loop instead.
- [x] `theme` is broken.
- [x] `--test` is broken.
- [x] Battery: [Linux] Subtitle doesn't display.
- [x] Distro: Distro doesn't display.
- [x] We can no longer log if a function failed to detect info without adding 3 more lines to each function.

## Testing

Run neofetch without a user config.

```sh
neofetch --config off
```


dylanaraps:
I'm not going to merge the commits that remove info(). I'm going to cherrypick some of the bug fixes / smaller changes from this. (Removing `get` and etc)

PR #419, opened 2016-11-04
Added initial support for Haiku

konimex:
### Features
See title.

### Issues
For `getcpu_usage` and `getterm`, we can't really rely on Haiku's tools.

This is Haiku's `ps`.
virtualbox_haiku os_12_10_2016_17_16_50

I have no idea what to do about it, and it doesn't even show anything about CPU, using `ps -i` shows memory info instead, unlike a sane `ps` from `procps-ng`.

A better alternative is using `top`, but we all know that `top` does not close immediately after fetching the CPU usage.. So... yeah.

### TODO
- [ ] getmodel
~~may be possible with `listdev`~~, ~~but I need a testing device to back this up. I'll test on my laptop.~~ Nope, `listdev` didn't show anything.
- [x] getuptime
Since we changed `getuptime` behaviour, we need another method for Haiku.
- [ ] getde
- [ ] getwm
Since Haiku doesn't use Xorg, I don't really know how to fetch it, or we can fill it manually.
- [ ] getdisk
~~I hate Haiku's `df`, it effectively makes `getdisk` impossible, _for now_~~ I need to know how Haiku handles when the case is two/more partitions mounted inside a filesystem.
- [ ] getbattery
~~No laptop to fetch `getbattery` and test it on. But~~ some Haiku users said that it is possible to get battery information, but so far I haven't got any more documentation about it.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -316,53 +316,54 @@ getuptime() {
             uptime="$(uptime -u)"
             uptime="${uptime/up }"
         ;;
-    *)
-        # Get uptime in seconds
-        case "$os" in
-            "Linux" | "Windows")
-                seconds="$(< /proc/uptime)"
-                seconds="${seconds/.*}"
-            ;;
 
-            "Mac OS X" | "iPhone OS" | "BSD")
-                boot="$(sysctl -n kern.boottime)"
-                boot="${boot/'{ sec = '}"
-                boot="${boot/,*}"
+        *)
+            # Get uptime in seconds
+            case "$os" in
+                "Linux" | "Windows")
+                    seconds="$(< /proc/uptime)"
+                    seconds="${seconds/.*}"
+                ;;
 
-                # Get current date in seconds
-                now="$(date +%s)"
-                seconds="$((now - boot))"
-            ;;
+                "Mac OS X" | "iPhone OS" | "BSD")
+                    boot="$(sysctl -n kern.boottime)"
+                    boot="${boot/'{ sec = '}"
+                    boot="${boot/,*}"
 
-            "Solaris")
-                seconds="$(kstat -p unix:0:system_misc:snaptime | awk '{print $2}')"
-                seconds="${seconds/.*}"
-            ;;
-        esac
+                    # Get current date in seconds
+                    now="$(date +%s)"
+                    seconds="$((now - boot))"
+                ;;
 
-        days="$((seconds / 60 / 60 / 24)) days"
-        hours="$((seconds / 60 / 60 % 24)) hours"
-        minutes="$((seconds / 60 % 60)) minutes"
+                "Solaris")
+                    seconds="$(kstat -p unix:0:system_misc:snaptime | awk '{print $2}')"
+                    seconds="${seconds/.*}"
+                ;;
+            esac
 
-        case "$days" in
-            "0 days") unset days ;;
-            "1 days") days="${days/s}" ;;
-        esac
+            days="$((seconds / 60 / 60 / 24)) days"
+            hours="$((seconds / 60 / 60 % 24)) hours"
+            minutes="$((seconds / 60 % 60)) minutes"
 
-        case "$hours" in
-            "0 hours") unset hours ;;
-            "1 hours") hours="${hours/s}" ;;
-        esac
+            case "$days" in
+                "0 days") unset days ;;
+                "1 days") days="${days/s}" ;;
+            esac
 
-        case "$minutes" in
-            "0 minutes") unset minutes ;;
-            "1 minutes") minutes="${minutes/s}" ;;
-        esac
+            case "$hours" in
+                "0 hours") unset hours ;;
+                "1 hours") hours="${hours/s}" ;;
+            esac
 
-        uptime="${days:+$days, }${hours:+$hours, }${minutes}"
-        uptime="${uptime%', '}"
-        uptime="${uptime:-${seconds} seconds}"
-    ;;
+            case "$minutes" in
+                "0 minutes") unset minutes ;;
+                "1 minutes") minutes="${minutes/s}" ;;
+            esac
+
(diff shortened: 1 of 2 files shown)


dylanaraps:
I'll install Haiku on my laptop tomorrow to help out.

dylanaraps:
This is a simpler CPU command:

```sh
cpu="$(sysinfo -cpu | awk -F '\\"' '/CPU #0/ {print $2}')"
cpu="${cpu/@*}"
```

dylanaraps:
This is a simpler speed command:

```sh
speed="$(sysinfo -cpu | awk '/running at/ {print $NF; exit}')"
speed="${speed/MHz}"
```

dylanaraps:
This is a simpler memory command:

```sh
memtotal="$(($(sysinfo -mem | awk -F '\\/ |)' '{print $2; exit}') / 1024 / 1024))"
memused="$(sysinfo -mem | awk -F '\\/|)' '{print $2; exit}')"
memused="$((${memused/max} / 1024 / 1024))"
```

dylanaraps:
Simpler resolution command: 

```sh
            resolution="$(screenmode | awk -F ' |, ' '{printf $2 "x" $3 " @ " $6 $7}')"

            case "$refresh_rate" in
                "off") resolution="${resolution/ @*}"
            esac
```

dylanaraps:
Simpler local IP command:

```sh
localip="$(ifconfig | awk -F ': ' '/Bcast/ {print $2}')"
localip="${localip/', Bcast'}"
```

neofetch
```diff
@@ -1715,6 +1759,10 @@ getlocalip() {
         "Windows")
             localip="$(ipconfig | awk -F ': ' '/IPv4 Address/ {printf $2}')"
         ;;
+
+        "Haiku")
+            localip="$(ifconfig | awk -F ': ' '/Bcast/ {print $2}')"
+            localip="${localip/', Bcast'}"
```
dylanaraps: Missing `;;`

neofetch
```diff
@@ -1268,6 +1305,13 @@ getresolution() {
             [ "$width" ] && \
                 resolution="${width}x${height}"
         ;;
+
+        "Haiku")
+            resolution="$(screenmode | awk -F ' |, ' '{printf $2 "x" $3 " @ " $6 $7}')"
+
+            case "$refresh_rate" in
+                "off") resolution="${resolution/ @*}"
+            esac
```
dylanaraps: Missing `;;`

neofetch
```diff
@@ -1025,6 +1053,9 @@ getgpu() {
             gpu="$(wmic path Win32_VideoController get caption /value)"
             gpu="${gpu/Caption'='}"
         ;;
+
+        "Haiku")
+            gpu="$(listdev | grep -A2 -e 'device Display controller' | awk -F':' '/device beef/ {print $2}')"
```
dylanaraps: Missing `;;`

konimex:
Since Haiku's `uptime` can't be fetched in seconds only, should we make a separate case for Haiku and the rest or find another way?

dylanaraps:
Use `uptime -u` and add another case for Haiku

dylanaraps:
Change distro to:

```sh
distro="$(uname -sv | awk '{print $1 " " $2}')"
```

dylanaraps:
Can you change the DE/WM on Haiku? If not then we can just hardcode it.

konimex:
Since Haiku's WM and DE system "[are an integral part of the system and doesn't have separate WM](https://www.haiku-os.org/docs/userguide/en/gui.html)", I think hardcoding it is the best option.

return:
Would reading from /dev/power/acpi_battery/0 in Haiku be useful for getbattery?
Here is the output:

`cat /dev/power/acpi_battery/0`

`Battery Status:`
 `State 0, Current Rate 0, Capacity 50000, Voltage 10000`

 `Battery Info:`
 `Power Unit 0, Design Capacity 50000, Last Full Charge 50000, Technology 1`
 `Design Voltage 10000, Design Capacity Warning 100, Design Capacity Low 50, Capacity      Granularity 1     1, Capacity Granularity1 1`
 `Model Number 1, Serial Number 0, Type VBOX, OEM Info innotek`

konimex:
Does the capacity changes like it does with Linux when it is discharging? We'll need to change the capacity to a 0-100 format, though.

Oh, and @dylanaraps, you may (and please) push directly to this branch if you got something new for the Haiku commands, as I'll cease doing the unorthodox rebase-and-force-push method for this branch.

return:
The capacity does indeed change in Haiku with laptop users but since I am using a desktop, It will remain constant. There is a app called PowerStatus that also reads from acpi_battery.

dylanaraps:
I've implemented battery as @return has shown above and it prints 0% on my system. I'm running Haiku inside a VM so I'm not sure if that's the reason for the wrong output.

I've also disabled `getdisk` on Haiku to suppress errors.

dylanaraps:
I can't find anything online about the names of the DE and WM. Any ideas?

konimex:
I'd say leave them empty since Haiku's windowing system are knitted tight unlike Unix-like system. We may need to suppress the "cannot detect DE/WM" errors, though.

dylanaraps:
I think it's better to leave the errors there since we did fail to detect them.

dylanaraps:
This looks ready to me now, there's no way we can add detection for the other functions. We can always add support later on if someone figures out how.

:+1:

konimex:
Alright, one last check seems to indicate that it's clear. You may merge this. Along with the Hurd one.

dylanaraps:
:+1:

dylanaraps:
There are merge conflicts, I'll have to add this manually probably.

dylanaraps:
Merged, thanis.

PR #421, opened 2016-11-04
Fixed gpu detection on linux

TonCherAmi:
## The issue

In case of several GPUs present, current version does not detect which one is active at the moment.

For example: 
`lspci -mm | awk -F '\\"|\\" \\"' '/3D|VGA/ {print $3 " " $4}'`
on my system returns: 
`Intel Corporation Xeon E3-1200 v3/4th Gen Core Processor Integrated Graphics Controller
NVIDIA Corporation GK106 [GeForce GTX 660]`

Which is afterwards immediately grabbed by:
```bash
case "$gpu" in
     intel*) gpu="Intel Integrated Graphics" ;;
```
Whether the GPU that goes first in the returned string is currently being used is disregarded.

## The fix

I believe that this issue can be fixed by checking whether there is a kernel driver in use for a given GPU. 
The check is performed by the regex:
`'VGA(?=.*\n.*\n\tKernel driver in use)'`
Then we grab the BDF number of that GPU and proceed to finding the right GPU by it (would not be necessary if `lspci -kmm` worked the way it's supposed to but alas).

I also added a fallback to the old detection method in case of no drivers present (e.g. an X-less system) or some other issues.

## Issues
I haven't tested it anywhere but on my own system so I believe some additional testing is required.

### DIFF
--- a/Changelog.md
+++ b/Changelog.md
@@ -1,6 +1,22 @@
 ## Contributors
 
+- **[@JorgeGonzalez](https://github.com/JorgeGonzalez)**
+
+
 ## General
 
 - Fixed issue where info wasn't detected properly but the subtitle was still displayed.
+
+
+
+## Info
+
+**Terminal**<br \>
+
+- Added support for HyperTerm. **[@JorgeGonzalez](https://github.com/JorgeGonzalez)**
+
+
+**Terminal Font**<br \>
+
+- Added support for HyperTerm. **[@JorgeGonzalez](https://github.com/JorgeGonzalez)**
 
--- a/neofetch
+++ b/neofetch
@@ -1542,7 +1542,7 @@ gettermfont() {
             termfont="${termfont/:*}"
         ;;
 
-        "HyperTerm")
+        "Hyper"*)
             termfont="$(awk -F "," '/fontFamily/ {a=$1} END{print a}' "${HOME}/.hyper.js" | awk -F "'" '{a=$2} END{print a}')"
         ;;
     esac
--- a/neofetch
+++ b/neofetch
@@ -934,8 +934,18 @@ getgpu() {
             if [ -f "/tmp/neofetch/gpu" ]; then
                 source "/tmp/neofetch/gpu"
             else
-                gpu="$(PATH="/sbin:$PATH" lspci -mm | awk -F '\\"|\\" \\"' '/3D|VGA/ {print $3 " " $4}')"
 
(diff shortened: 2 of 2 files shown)


dylanaraps:
My only gripe with this other than testing it is `grep -P`. How portable is the `-P` flag?

dylanaraps:
I did some googling, does this command print your active GPU?

```sh
lspci -vnnn | grep "VGA controller"
```

TonCherAmi:
Yeah, I should've paid more attention, it says that it's highly experimental and I'm unsure about the portability.

About `lspci -vnnn | grep "VGA controller"` I've seen some people suggest it too, but it prints both as well:
```
00:02.0 VGA compatible controller [0300]: Intel Corporation Xeon E3-1200 v3/4th Gen Core Processor Integrated Graphics Controller [8086:0412] (rev 06) (prog-if 00 [VGA controller])
01:00.0 VGA compatible controller [0300]: NVIDIA Corporation GK106 [GeForce GTX 660] [10de:11c0] (rev a1) (prog-if 00 [VGA controller])
```

dylanaraps:
> but it prints both as well:

Damn, that's a shame.

Have you tried checking loaded kernel modules? You might be able to check to see if the NVIDIA driver is loaded and if so specifically look for NVIDIA in lspci.

TonCherAmi:
> Have you tried checking loaded kernel modules? You might be able to check to see if the NVIDIA driver is loaded and if so specifically look for NVIDIA in lspci.

But wouldn't that mean that we'd have to check every possible GPU module there is? It seems kind of bloated to check for nvidia, nouveau, amd, intel etc.

dylanaraps:
I was thinking it might look a little like this:

```sh
driver="$(command)"

case "$driver" in
    "intel_bla" | "intel_blabla") driver="intel" ;;
    "nouveau" | "nvidia") driver="nvidia" ;;
    "amd" | "radeon" | "catalyst") driver="amd" ;;
esac
```

TonCherAmi:
I kind of really dislike the idea of looking up all the `"intel_bla", "intel_blabla"` etc.

What about this?
`bdf_number="$(PATH="/sbin:$PATH" lspci -k | grep -A2 'VGA' | grep -B2 'Kernel driver in use' | awk '/^\w/ {print $1}')`

dylanaraps:
LGTM :+1:

dylanaraps:
Merged.

Check the 2.0 branch to see your changes :)

PR #438, opened 2016-11-13
Added fallback ASCII for *BSD, Solaris, and GNU (Hurd)

konimex:
## Description

Added fallback ASCII for contingency if neofetch detects a BSD/Solaris/GNU (Hurd)-based systems that doesn't yet/will never have proper ASCII.

Logos:
- BSD Beastie taken from [Wikipedia](https://en.wikipedia.org/wiki/BSD_Daemon#ASCII_image)
- GNU ASCII from [the GNU website](https://www.gnu.org/graphics/alternative-ascii.html)
- Solaris ASCII from the old Sun Solaris logo


dylanaraps:
:+1:

PR #440, opened 2016-11-15
Music: Added support for various music players

konimex:
## Description

This adds support for GNOME Music and Lollypop.

### Issues

wtf dbus

My initial understanding here is in different sessions of GNOME Music, sometimes `dbus-send` produces `xesam:artist` first before `xesam:title`, sometimes it's `xesam:title` first before `xesam:artist`, resulting in this.

I wonder why don't gdbus simplify things like `qdbus` do...

NOTE: This also happened to Lollypop. So if my hypothesis is true, Spotify should be affected as well.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -1076,7 +1076,7 @@ get_memory() {
 
 get_song() {
     # This is absurdly long.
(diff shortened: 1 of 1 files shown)


dylanaraps:
:+1:

PR #442, opened 2016-11-17
Fix connection timed out with public_ip (dig)

iandrewt:
## Description

At my school (and I presume many other places with strong proxies in place) the `public_ip` function returns `;; connection timed out; no servers could be reached`.

## Features

Adds a check to see if the returned text from `dig` is a comment (starts with `;`) and unsets `public_ip`


dylanaraps:
:+1:

PR #452, opened 2016-11-24
Music: Added support for Clementine and Pragha

konimex:
## Description

Self explanatory.

Clementine is a Qt-based music player.

Pragha is the default Music Player in openSUSE XFCE.


dylanaraps:
LGTM :+1:

PR #463, opened 2016-11-26
Config: Updated description

konimex:
Added some clarifications since more OSes are supported with those two flags.


dylanaraps:
Thanks! LGTM

PR #472, opened 2016-12-02
Public IP: Add caching

konimex:
## Description

Self explanatory. This should make the script faster when a second iteration of neofetch is being run.


dylanaraps:
Good idea! :+1:

PR #476, opened 2016-12-02
Distro: Added support for GoboLinux

konimex:
## Description

Self-explanatory.

Yes, `GoboLinuxVersion` is correctly capitalized like that. So is `Compile`. So is `/Programs`, not `/programs`.

*Why detect `/etc/GoboLinuxVersion` instead of using `/etc/os-release`?*
Because there's no `/etc/*-release` in GoboLinux systems. But at least they have something useful in place of it, unlike GuixSD.

### Issues

No logo.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -52,6 +52,12 @@ get_distro() {
                 esac
                 ascii_distro="Windows 10"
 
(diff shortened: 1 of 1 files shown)


dylanaraps:
LGTM :+1:

PR #485, opened 2016-12-03
Kernel: Modify how kernel_shorthand works in BSDs

konimex:
## Rationale

Instead of hardcoding PacBSD and PCBSD's name, detect if `$distro` is roughly the same as `$kernel_name`.


dylanaraps:
:+1: LGTM

PR #486, opened 2016-12-03
Packages: Call wc -l once at the end, instead of multiple times. 

dylanaraps:
## Description

When I was trying to implement package manager names I took a closer look at the packages function and noticed that we were calling `wc -l` to count the packages once per package manager. 

I started messing around with the function and came up with a way to call `wc -l` only once. We add each list of packages to an array, count the array using `wc -l` and then do some math to fix inaccuracy caused by array elements joining due to missing newlines. 

This should be way faster than the current method since we no longer pipe to an external program on a per package manager basis, we just call it once at the end. 

## Features

- Only call `wc -l` once. 

## Testing

I don't have any systems with more than one package manager installed so I faked it on my machine to test locally. I'd still like someone to test this on a real machine before I merge this to master.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -372,66 +372,70 @@ get_packages() {
     case "$os" in
         "Linux" | "iPhone OS" | "Solaris" | "GNU")
             type -p pacman >/dev/null && \
-                packages="$(pacman -Qq --color never | wc -l)"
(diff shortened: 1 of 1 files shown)


konimex:
Works on my end :+1:

PR #492, opened 2016-12-04
Terminal: NetBSD, use Linux's function

konimex:
## Description

Because in `get_cpu` NetBSD uses Linux function instead of the usual thing, why not use the same thing for `get_term` since they also fetch information from `/proc` ?

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -1491,6 +1491,11 @@ get_font() {
 }
 
 get_term() {
+    # Since NetBSD emulates Linux's /proc, we can take advantage of it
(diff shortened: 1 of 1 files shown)


dylanaraps:
Have you tested this on NetBSD?

I'm not sure of the extent that NetBSD emulates the /proc/ filesystem.

konimex:
Yeah, turns out it's not a full emulation.
Oops.

dylanaraps:
All good.

PR #502, opened 2016-12-05
Scrot: Automatically detect tools

dylanaraps:
## Description

This PR adds a new function called `scrot_program()` which automatically detects screenshot tools on your system. Users no longer have to specify which tool to use in their config file. 

This only supports systems with an X server, we can add support for macOS and Windows if we can figure out how to take screenshots on these machines from the commandline.

One of the tools in the `scrot_program()` function is most likely already installed so for majority of people this will be plug and play.

The `scrot_cmd` config option will still exist and still work if users want more control over which program/flags to use.

## Features

- Automatically use whatever screenshot tool is available.
- Added screenshot support to macOS
- Added screenshot support to Haiku

## Issues

- This function supports every OS but Windows. There's no builtin way on Windows to take screenshots from the command line. CYGWIN + an X server will still work on Windows.

### DIFF
--- a/config/config
+++ b/config/config
@@ -629,16 +629,20 @@ ascii_bold="on"
 #          -s
 scrot="off"
 
-# Screenshot program to launch
-# If you're not using 'scrot' change this to your screenshot
-# program.
+# Screenshot Program
+# Neofetch will automatically use whatever screenshot tool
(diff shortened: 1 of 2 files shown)


iandrewt:
Works on macOS!

PR #504, opened 2016-12-05
Wallpaper: Expanded support to Solaris and GNU Hurd

konimex:
## Description

Self-explanatory.


dylanaraps:
This could also be expanded to `GNU` too right?

konimex:
Theoretically, all systems running Xorg should be supported. So, yep.

dylanaraps:
:+1:

PR #505, opened 2016-12-05
BSD: Use $kernel_name instead of $distro in some functions

konimex:
## Rationale

In some functions (such as `get_cpu` in fetching `cpu_temp`, `get_gpu`, `get_mem`, `get_battery` and `get_birthday`) it is better to use `$kernel_name` instead of `$distro` for the following reason:

Most of projects forked from the trinity (FreeBSD, OpenBSD, and NetBSD) have the same underlying `uname`, (with the exception of Bitrig and DragonFly). As we add more distros (as of now, we have 2 distros with the same `$kernel_name` as FreeBSD: PacBSD and PCBSD. But we have more cases like that, for instance, FreeNAS and pfSense), if we keep on using `$distro`, we would have to hardcode more and more of it in those cases. In such cases, using `$kernel_name` will be more effective. So, this is more like contingency in case we would have to add more detection BSD-based distros which derived from those projects.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -815,8 +815,8 @@ get_cpu() {
 
             # Get cpu temp
             if [[ "$cpu_temp" == "on" ]]; then
-                case "$distro" in
-                    "FreeBSD"* | "PacBSD"* | "DragonFly"* | "PCBSD"*)
+                case "$kernel_name" in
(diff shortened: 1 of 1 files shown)


dylanaraps:
Nice! :+1:

Sounds good to me.

PR #516, opened 2016-12-10
Colors: use start++ instead of adding it manually after case

konimex:
## Description

Code is self-explanatory. Just a small detail.


dylanaraps:
Neato! :+1:

PR #518, opened 2016-12-12
Cache: Simplify caching

dylanaraps:
## Description

We're only caching GPU on macOS systems right now so it makes no sense for Neofetch to do a file exists check for every info function and Operating System.

I'd like to revisit caching but it's going to be very difficult right now.

## TODO

@iandrewt, can you test this on macOS? (Just check that GPU caching still works on macOS)

- [x] Testing
    - [x] macOS

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -976,7 +976,7 @@ get_gpu() {
         ;;
 
         "Mac OS X")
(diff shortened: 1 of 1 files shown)


iandrewt:
Works 👍

dylanaraps:
Thanks :+1:

PR #522, opened 2016-12-12
[macOS] Always charging bug

jorgegonzalez:
## Description

Battery would display as "charging" every time on the last few versions of Neofetch. This patch fixes that.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -1709,7 +1709,11 @@ get_battery() {
 
         "Mac OS X")
             battery="$(pmset -g batt | grep -o '[0-9]*%')"
-            battery_state="$(pmset -g batt | awk 'NR==2 {print $3}')"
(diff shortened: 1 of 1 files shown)


neofetch
```diff
@@ -1709,7 +1709,11 @@ get_battery() {
 
         "Mac OS X")
             battery="$(pmset -g batt | grep -o '[0-9]*%')"
-            battery_state="$(pmset -g batt | awk 'NR==2 {print $3}')"
+            state="$(pmset -g batt | awk '/;/ {print $4}')"
+            if [ "$state" == "charging;" ];
```
iandrewt: Could this be changed to a one-line if statement please? This would like like this:

```bash
[[ "$state" == "charging;" ]] && battery_state="charging"
```

jorgegonzalez:
Made that change.

iandrewt:
Thanks! 👍

PR #524, opened 2016-12-13
Info: Add alignment options

dylanaraps:
## Description

This PR adds options to align the info output into columns! This works by specifying a max width for the subtitle `$align_amount` and then subtracting the length of the subtitle from that `${#subtitle}`.

`$align_amount` has to be hardcoded since we won't know the length of the longest subtitle when we start displaying info. The default amount works for all of the default subtitles in the config file and if there are any issues the user just has to up the value of `$align_amount`. 

This is mentioned in the config file and docs so it shouldn't be an issue.

Examples:

```sh
# LEFT

user@hostname
-------------
OS            : Arch Linux
Kernel        : Linux 4.9-ARCH
Uptime        : 1 hour, 41 minutes

# RIGHT

user@hostname
-------------
           OS : Arch Linux
       Kernel : Linux 4.9-ARCH
       Uptime : 1 hour, 41 minutes
```

## Features

- Align the info output into columns.

## Issues

- [x] `$align_amount` must be hardcoded since we don't know the max subtitle width.
- [x] `--disable` doesn't update max subtitle length.

## Testing

```sh
neofetch --align_info left
neofetch --align_info right
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,9 +5,6 @@
 - **[@jorgegonzalez](https://github.com/jorgegonzalez)**
 
 
-## Collaborators
-
-
 ## IRC
 
 Neofetch now has an irc channel at `#neofetch` on Freenode. If you have any questions, issues or ideas feel free to join the irc channel and I'll be happy to assist you. I know that we've already got the gitter chat but hopefully this makes things easier for those without a github account. :)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@
 - **[@jorgegonzalez](https://github.com/jorgegonzalez)**
 
 
+## Collaborators
+
+
 ## IRC
 
 Neofetch now has an irc channel at `#neofetch` on Freenode. If you have any questions, issues or ideas feel free to join the irc channel and I'll be happy to assist you. I know that we've already got the gitter chat but hopefully this makes things easier for those without a github account. :)
--- a/config/config
+++ b/config/config
@@ -61,6 +61,7 @@ print_info() {
 # -------------
 # OS            : Arch Linux
 # Kernel        : Linux 4.9-ARCH
+# Terminal Font : Lemon
 # Uptime        : 1 hour, 41 minutes
 #
 # RIGHT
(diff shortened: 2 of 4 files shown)


konimex:
I had a test run, it seems battery function won't play nice with alignment. 
2016-12-13-131857_842x526_scrot

dylanaraps:
It looks like I've forgotten to update `prin()`. 

Thanks for testing.

dylanaraps:
It should be fixed now.

iandrewt:
This isn't working at all for me on macOS with iTerm2
screen shot 2016-12-13 at 7 07 31 pm

dylanaraps:
Does it work with `--config off` added as well?

I can't reproduce that at all.

dylanaraps:
I can reproduce this if I set `--align_amount` to a low value. Try using `--align_amount 20` and see if it fixes the issue.

iandrewt:
Yeah that fixes it. What's the default value?

dylanaraps:
The default is `14` which is slightly longer than `Terminal Font` which is the longest default subtitle.

iandrewt:
Well iTerm2 doesn't print the terminal font so hmm. Longest for me with default config is `8`

dylanaraps:
It's hardcoded at `14` so even if the max subtitle size is smaller (`8`) it just means that there will be a bigger gap. The breakage should only happen if the subtitle is longer than the default value.

dylanaraps:
I've figured out how to calculate the longest subtitle length. We basically run `print_info()` again and exit right after we get the subtitle value. This is the only way I can see this implemented and it works surprisingly well.

Try this revision @iandrewt and let me know if it works now.

iandrewt:
Works now!

dylanaraps:
Awesome :+1:

dylanaraps:
I've found a new bug.

For example: If `term_font` fails (ie we don't print anything) the length of the subtitle is still logged. This means that the alignment calculation is off and you end up with more padding than needed.

dylanaraps:
Closed in favor of https://github.com/dylanaraps/neofetch/pull/532

PR #528, opened 2016-12-14
Generate man page from --help and --version

dylanaraps:
## Description

This PR re-formats the `--help` and `--version` outputs to be usable with `help2man`. This allows us to automatically generate the manpage using `help2man -N neofetch -o neofetch.1`.

We could also possibly automate the man page generation, I'm looking into this. ~~A git hook might work(?)~~

One option is to implement this in the makefile but seeing as this is a GNU tool (not portable) I don't want to add a dependency for it. I'd rather we automatically do it every push to the repo somehow.

## TODO

- [ ] Automate this.
- [x] Remove old manpage.


dylanaraps:
Even if we have to manually call `help2man bla bla` it's easier than adding to the man page manually.

iandrewt:
I'd suggest making a git commit hook that checks if the relevant sections are different between `neofetch` and `neofetch.1` and adds it to the commit (or makes an extra commit with the man-page update)

dylanaraps:
Are git hooks local only or will they be added to every cloned version of Neofetch?

iandrewt:
What does your `git config --edit` look like? Apart from the `remotes` sections, mine looks like this:

```
[core]
	repositoryformatversion = 0
	filemode = true
	bare = false
	logallrefupdates = true
	ignorecase = true
	precomposeunicode = true
```

dylanaraps:
Just this:

```sh
[core]
	repositoryformatversion = 0
	filemode = true
	bare = false
	logallrefupdates = true
```

dylanaraps:
I've made a simple git hook which works fine. I haven't added any checks to see if the output of `--help` has changed since this hook doesn't add that much of a delay.

The hook will only affect the commit if the manpage differs so it's fine if it runs on every `git commit` anyway.


```sh
#!/usr/bin/env bash
#
# Neofetch Pre-Commit hook
if type -p help2man >/dev/null 2>&1; then
    help2man -N "neofetch" -o "neofetch.1"
    git add "neofetch.1"
fi
```

iandrewt:
You could make a `CONTRIBUTING.md` file with instructions for setting up a development environment too

dylanaraps:
I think it's even easier if we just run `help2man` manually ourselves whenever we remember to. No need to over complicate things.

dylanaraps:
I've added a new flag called `--gen-man` that generates the manpage.

I'm merging this now. We could still automate this in the future.

iandrewt:
You could probably get Travis to do the automation, actually

dylanaraps:
I though about that too but how would we then save it to the repo?

iandrewt:
You could make Travis make its own commit and push it, not sure how you'd configure it though

This may be of use: https://gist.github.com/domenic/ec8b0fc8ab45f39403dd

PR #532, opened 2016-12-15
Info: Add alignment v2

dylanaraps:
## Description

Same PR as before but fixed merge conflicts with the new `info()` and `prin()` functions.

Examples:

```sh
# LEFT

user@hostname
-------------
OS            : Arch Linux
Kernel        : Linux 4.9-ARCH
Uptime        : 1 hour, 41 minutes

# RIGHT

user@hostname
-------------
           OS : Arch Linux
       Kernel : Linux 4.9-ARCH
       Uptime : 1 hour, 41 minutes
```

## Features

- Align the info output into columns.

## Issues

- [x] `$align_amount` must be hardcoded since we don't know the max subtitle width.
- [x] `--disable` doesn't update max subtitle length.
- [ ] This requires a 'hack' to get working.
- [ ] Max subtitle calculation doesn't check to see if function failed.

## Testing

```sh
neofetch --align_info left
neofetch --align_info right
```

Old PR: https://github.com/dylanaraps/neofetch/issues/524

### DIFF
--- a/config/config
+++ b/config/config
@@ -47,6 +47,43 @@ print_info() {
 }
 
 
+# Info alignment
+
+# Align the info into columns.
+# Default:  'off'
+# Values:   'left', 'right', 'off'
+# Flag:     --align_info
+# Example:
+#
+# LEFT
+#
+# user@hostname
+# -------------
+# OS            : Arch Linux
+# Kernel        : Linux 4.9-ARCH
+# Terminal Font : Lemon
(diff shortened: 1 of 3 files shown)


dylanaraps:
Unless we fix the bug mentioned in the comments of the previous PR, this won't be merged. I have ideas on how to fix it and I'll implement those in this PR.

PR #534, opened 2016-12-15
Uptime: Rewrite part of uptime

dylanaraps:
## Features

- `$minutes` --> `$mins`
    - This makes the variable the same length as the others.
- Moved the 3 duplicate case statements to a function. `strip_date()`
- Changed default `uptime_shorthand` value to `on`.
    - `on` just makes `minutes` --> `mins` in the output.
    - I've seen screenshots of people with weeks of uptime and they still have this set to `off` so the line is extremely long. This is a better default imo.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -329,24 +329,22 @@ get_uptime() {
 
             days="$((seconds / 60 / 60 / 24)) days"
             hours="$((seconds / 60 / 60 % 24)) hours"
-            minutes="$((seconds / 60 % 60)) minutes"
+            mins="$((seconds / 60 % 60)) minutes"
 
-            case "$days" in
-                "0 days") unset days ;;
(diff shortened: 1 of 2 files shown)


neofetch
```diff
@@ -329,24 +329,22 @@ get_uptime() {
...
-                "1 days") days="${days/s}" ;;
-            esac
-
-            case "$hours" in
-                "0 hours") unset hours ;;
-                "1 hours") hours="${hours/s}" ;;
-            esac
+            mins="$((seconds / 60 % 60)) minutes"
+
+            # Format the days, hours and minutes.
```
konimex: I think it's better to leave them be inside `get_uptime()`.

dylanaraps: Will do.

PR #536, opened 2016-12-15
Song: Cleanup get_song()

dylanaraps:
## Description

Title.

## Features

- Remove duplicate `dbus-send` commands.
- Remove state support.

## TODO

- [x] Testing
    - [x] GNOME-Music
    - [x] Lollypop
    - [x] Clementine
    - [x] Spotify
- [x] Possibly include Spotify on Linux.
    - [x] Move the `dbus` command to a function.

@konimex, can you test this out?

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -1152,21 +1152,20 @@ get_song() {
     }
 
     case "${player/*\/}" in
-        "mpd"*)
-            song="$(mpc current)"
-            state="$(mpc | awk -F '\\[|\\]' '/\[/ {printf $2}')"
-        ;;
+        "mpd"*)         song="$(mpc current)" ;;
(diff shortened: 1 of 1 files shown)


konimex:
After the second commit it works. I can only test this with GNOME Music though.

dylanaraps:
I'll install the others. Thanks :+1:

konimex:
While we're at it, I think we can delete all `$state` functions in `get_song()` as I think it's completely useless. Users' default behaviour when a music player stops playing its music, the music player will be closed (effectively killing the process), and as a result, there will be no output in `$state`.

dylanaraps:
Sounds good :+1:.

PR #538, opened 2016-12-16
Birthday: Remove 'date' usage and calclulate date manually

dylanaraps:
## Description

This PR removes all `date` usage from Neofetch and also renames `get_birthday()` to `get_install_date()` to make things less confusing.

This adds a new function called `convert_time()` which takes the date time-stamp (eg. `2016-12-06 16:32.00000`) and converts it to (eg. `Tue 06 Dec 2016 4:58 PM`) using nothing but bash. This will allow us to use the same format for all OS since we no longer depend on `date`.

This PR also removes `$birthday_shorthand` and `$birthday_format`.

~~Note: This was more of a test to see if this was possible, it probably won't be added to master.~~

The `convert_time()` function now works with all Operating Systems and is much faster than `date`. We also don't have to deal with all of the different `date` programs and their flags. This PR will be merged once testing is finished.

## Features

- Renamed `get_birthday()` -- > `get_install_date()`
- Removed all `date` usage from `get_install_date()`. 
- Added a new function called `convert_time()` which takes the time stamped `ls` output and converts it to a pretty format. The function only uses bash so its much faster than calling `date`. This makes things simple and keeps the output consistent across all Operating Systems. Example: `2016-12-06 16:58:58.000000000` --> `Tue 06 Dec 2016 4:58 PM`

## Issues

- None yet

## TODO

- [x] Testing
    - [x] Linux | GNU | iOS
    - [x] Windows
    - [x] BSD (Testing one BSD should suffice.)
        - [x] OpenBSD
        - [x] FreeBSD
        - [ ] NetBSD
        - [ ] DragonflyBSD
    - [x] macOS
    - [x] Haiku
    - [x] Solaris

## Sources

**Calculating Leap Year**

http://stackoverflow.com/questions/725098/leap-year-calculation

```sh
function isLeapYear (year):
    if ((year modulo 4 is 0) and (year modulo 100 is not 0))
    or (year modulo 400 is 0)
        then true
    else false
```

**Calculating Day of Week**

http://blog.artofmemory.com/how-to-calculate-the-day-of-the-week-4203.html

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -317,7 +317,7 @@ get_uptime() {
                     boot="${boot/,*}"
 
                     # Get current date in seconds
-                    now="$(date +%s)"
+                    now="$(printf "%(%s)T")"
                     seconds="$((now - boot))"
                 ;;
 
@@ -1730,7 +1730,6 @@ get_birthday() {
     case "$os" in
         "Linux" | "GNU" | "iPhone OS")
             birthday="$(ls -alct --full-time / | awk '/lost\+found|private/ {printf  $6 " " $7}')"
-            date_cmd="$(date -d"$birthday" "$birthday_format")"
         ;;
 
         "Mac OS X")
@@ -1757,7 +1756,6 @@ get_birthday() {
 
                 "FreeBSD"*)
                     birthday="$(ls -alctT /etc/hostid | awk '{printf $6 " " $7 " " $9 " " $8}')"
-                    date_cmd="$(date -j -f "%b %d %Y" "$birthday" "$birthday_format")"
                 ;;
 
                 "NetBSD"* | "DragonFly"*)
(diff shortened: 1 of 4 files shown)


iandrewt:
Output on macOS: `Birthday: Sat 2016 Sep 9:14 PM`

dylanaraps:
This is probably due to the macOS substitutions. What does this output?

```sh
ls -lUT /var/log/install.log | awk '{printf $6 " " $7 " " $9 " " $8}'
```

iandrewt:
```bash
$ ls -lUT /var/log/install.log | awk '{printf $6 " " $7 " " $9 " " $8}'
21 Sep 2016 21:14:48`
```

dylanaraps:
That's already really close to what we calculate. This is the output on Linux: 

```sh
ls -alct --full-time / | awk '/lost\+found|private/ {printf  $6 " " $7}'
2016-12-06 16:58:58.000000000
```

iandrewt:
It's because of the macOS `ls` vs GNU coreutils `ls`

dylanaraps:
Yup. 

macOS and BSD output: `21 Sep 2016 21:14:48`
Linux and * output: `2016-12-06 16:58:58.000000000`

iandrewt:
So maybe just the time portion needs to be changed for macOS and BSD?

dylanaraps:
Yeah, I can probably make the function work with both.

neofetch
```diff
@@ -317,7 +317,7 @@ get_uptime() {
                     boot="${boot/,*}"
 
                     # Get current date in seconds
-                    now="$(date +%s)"
+                    now="$(printf "%(%s)T")"
```
dylanaraps: Does this line work on macOS @iandrewt? Bash has builtin support for all date's format options with the caveat that you can only output the current date. (You can't get a previous date's info)

iandrewt: I get this: `printf: %(: invalid directive`

iandrewt: Oh oops I ran that in `zsh`

```bash
$ /bin/bash --version
GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin16)
Copyright (C) 2007 Free Software Foundation, Inc.
$ /bin/bash -c 'printf "%(%s)T"'
/bin/bash: line 0: printf: `(': invalid format character
```

dylanaraps: Does it work when run inside Neofetch? 

Does uptime display correctly?

dylanaraps: NVM

> http://stackoverflow.com/a/26804450

> Since Bash 4.2, printf has a new modifier %(dateformat)T that, when used with argument -1 outputs the current date with format given by dateformat, handled by strftime(3) (man 3 strftime for informations about the formats).

iandrewt:
```bash
$ /bin/bash neofetch
---
Uptime: -17148 days, -22 hours, -43 minutes
```

PR #542, opened 2016-12-16
Disk: Rewrite get_disk()

dylanaraps:
## Description

This PR aims to make the disk function as portable as possible. To make this possible we're now only displaying the information for the `/` partition. The old behavior will be re-implemented if we can figure out how to do so without issues on all versions of `df`.

Closes #542 

## Features

- Rewrote function from scratch.
    - The function is `40` lines smaller than before and works on all versions of `df` \[1\]
- We only show the `root (/)` partition now.
- Removed all percentage calculation since `df` already provides us with the percentage.
- We're using the same `df` flags across all Operating Systems now. 
    - No more ugly case statements or per distro hardcoding of `df` flags.

\[1\] The function doesn't work on Haiku since their `df` is wildly non-standard. (The output format and flags are 100% different from all of the other `df` versions floating around.) 

## Issues

None (?)

## TODO

- [x] Testing
    - [x] Linux | GNU | iOS
    - [x] macOS
    - [x] Windows
    - [x] BSD
        - [x] OpenBSD
        - [x] FreeBSD
        - [x] NetBSD
        - [x] DragonflyBSD
    - [ ] Haiku
    - [x] Solaris

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -1552,6 +1552,8 @@ get_disk() {
     # Convert Terabytes to Gigabytes.
     if [[ "$disk_display" != "off" ]]; then
         disk_used="${disk_used/\.}"
+        disk_used="${disk_used/G}"
+        disk_total="${disk_total/G}"
         disk_total="${disk_total/\.}"
 
         [[ "${disk_used: -1}" == "T" ]] && \
--- a/neofetch
+++ b/neofetch
@@ -1542,8 +1542,8 @@ get_disk() {
     # Get the info for /
     disks=($(df -P -h /)) || { err "Disk: 'df' exited with error code 1"; return; }
 
-    disk_used="${disks[9]}"
-    disk_total="${disks[8]}"
+    disk_used="${disks[9]/i}"
+    disk_total="${disks[8]/i}"
     disk_total_per="(${disks[11]})"
 
     # Put it all together
--- a/neofetch
+++ b/neofetch
@@ -1550,10 +1550,14 @@ get_disk() {
     (( "${#disk_used}"  == 6 )) && disk_used="${disk_used:0:1}.${disk_used:1:1}TB"
     (( "${#disk_total}" == 6 )) && disk_total="${disk_total:0:1}.${disk_total:1:1}TB"
 
-    disk_total_per="${disks[11]}"
+    # Get the filesystem type
+    disk_type="[$(awk '$2 == "/" {printf $3; exit}' /etc/fstab)]"
+
+    # Get the usage percentage
+    disk_total_per="(${disks[11]})"
 
     # Put it all together
-    disk="${disk_used} / ${disk_total} (${disk_total_per})"
+    disk="${disk_used} / ${disk_total} ${disk_total_per} ${disk_type}"
 
     # Bar
     case "$disk_display" in
--- a/neofetch
+++ b/neofetch
@@ -1539,15 +1539,8 @@ get_term_font() {
 get_disk() {
     type -p df >/dev/null 2>&1 || { err "Disk requires 'df' to function. Install 'df' to get disk info."; return; }
 
-    # Split the info
-    case "$os" in
-        "Haiku") return ;;
-        "Mac OS X") df_flags=(-P -h /) ;;
-        *) df_flags=(-h /) ;;
-    esac
-
     # Get the info for /
-    disks=($(df "${df_flags[@]}")) || { err "Disk: 'df' exited with error code 1"; return; }
+    disks=($(df -P -h /)) || { err "Disk: 'df' exited with error code 1"; return; }
 
     disk_used="${disks[9]}"
     disk_total="${disks[8]}"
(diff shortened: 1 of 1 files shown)


konimex:
I suggest adding back `-h` option as usual `df` in Oracle Solaris doesn't work.

This is the output of in a xpg4 (aka. what a Unix standard (SUS) will be)-compliant `df` when `df /` is being run:

`/                  (rpool/ROOT/solaris):14816931 blocks 14816931 files`

In xpg4, though, `df -h` works.

dylanaraps:
Alright.

Also it looks like we won't be able to get this working in Haiku, the `df` output format is entirely different.

konimex:
Well, to be fair, Haiku is what I would call an "outlier"-system, all of their commands barely comply with POSIX/SUS/xpg4/GNU standards.

Should we make separate cases for separate OSes instead? Since some systems can have `--total` in their `df` and some systems don't have it due to `--total` being a GNU extension.

dylanaraps:
I wanted to keep this as simple as possible. We could keep it as is and add exceptions for the Haiku and other OS which have a non-standard `df`.

konimex:
Sounds good. However since we are only counting `/`, some users might complain of "wrong disk size". Maybe a wiki page of our design decision would suffice?

konimex:
I've tested it with Solaris, and every *BSDs except FreeBSD, and confirmed that it works.

dylanaraps:
It works on iOS and Windows.

It's broken on macOS Travis, I think macOS has a different format for `df`...

konimex:
[According to StackExchange](https://apple.stackexchange.com/questions/132303/why-do-i-get-different-available-used-spaces-with-disk-info-and-df), this is a sample MacOS`df -h` output.

```
Filesystem                          Size   Used  Avail Capacity  iused    ifree %iused  Mounted on
/dev/disk0s2                       233Gi  118Gi  115Gi    51% 31026495 30042945   51%   /
devfs                              187Ki  187Ki    0Bi   100%      646        0  100%   /dev
map -hosts                           0Bi    0Bi    0Bi   100%        0        0  100%   /net
map auto_home                        0Bi    0Bi    0Bi   100%        0        0  100%   /home
localhost:/19YSQKN76Bmz377sW5fOwp  233Gi  233Gi    0Bi   100%        0        0  100%   /Volumes/MobileBackups
/dev/disk1s2                       119Gi   36Gi   83Gi    31%  9537194 21637536   31%   /Volumes/Jason SSD
```

dylanaraps:
This output looks like the Linux one...

> http://apple.stackexchange.com/questions/50302/how-can-i-tell-which-volume-the-operating-system-is-on

```sh
% df -h /
Filesystem     Size   Used  Avail Capacity  Mounted on
/dev/disk1s2  111Gi   75Gi   36Gi    68%    /
```

dylanaraps:
Still broken on macOS. I've pushed a temp commit to get travis to call `df -h` and exit.

dylanaraps:
Travis (macOS)

```sh
Filesystem      Size   Used  Avail Capacity  iused    ifree %iused  Mounted on
/dev/disk0s2   149Gi   70Gi   79Gi    47% 18300693 20811003   47%   /
devfs          176Ki  176Ki    0Bi   100%      608        0  100%   /dev
map -hosts       0Bi    0Bi    0Bi   100%        0        0  100%   /net
map auto_home    0Bi    0Bi    0Bi   100%        0        0  100%   /home
``

Travis (Ubuntu)

```sh
Filesystem      Size  Used Avail Use% Mounted on
/dev/sda1        30G  6.6G   22G  24% /
udev            3.7G  4.0K  3.7G   1% /dev
tmpfs           749M  244K  748M   1% /run
none            5.0M     0  5.0M   0% /run/lock
none            3.7G     0  3.7G   0% /run/shm
/dev/null       768M  176M  593M  23% /var/ramfs
```

dylanaraps:
This might be the reason for the different outputs:

> https://developer.apple.com/legacy/library/documentation/Darwin/Reference/ManPages/man1/df.1.html

> -i      Include statistics on the number of free inodes. This option is now the default to conform to
             Version 3 of the Single UNIX Specification (``SUSv3'') Use -P to suppress this output.

dylanaraps:
It works! 

```sh
Disk: 70Gi / 149Gi (47%) 
```

dylanaraps:
This page has some more info on `df` portability (We're as good as can be):

> http://mywiki.wooledge.org/BashFAQ/094

This PR is ready to be merged.

dylanaraps:
@konimex and @iandrewt, does this command get the correct file system type on your machines?

```sh
mount | awk '$3 == "/" {printf $5}'
```

iandrewt:
Not on macOS, prints `local,` for me. Changing `$5` to `$4` prints `(hfs,`

dylanaraps:
Yeah, there's no standard format for the `mount` output...

PR #544, opened 2016-12-17
OS: Expand support to Debian GNU/kFreeBSD

konimex:
## Description

This PR extends our compatibility to Debian GNU/kFreeBSD.

### Rationale

1. Merging of GNU in `get_os()` into Linux.

As I first tested this, because the `uname` `GNU/kFreeBSD` contains `BSD`, the `$os` is set to `BSD` for the remainder of this script. So, some utilities doesn't work as a result.

Also, a GNU/kFreeBSD (and to some extents, GNU Hurd) works the same way with Linux, so it makes sense to merge them.

2. Merged Local IP of BSD (and Solaris) to Linux

It turns out that GNU/kFreeBSD doesn't have `ip` (and the Hurd one doesn't have `ifconfig`). But since Linux has both `ifconfig` and `ip`, and `ifconfig` can be used as fallback if `iproute2` somehow isn't installed, so I merged it.

3. The $PATH

Since when we interact with outside software we only require `/usr/sbin:/usr/bin:/sbin:/bin`, I've set this to prevent neofetch interacting with pacman (the game, not the package manager).

Another reason I did this is because in Debian GNU/kFreeBSD, they only have `ifconfig` in `/sbin`, I could set this to `/sbin/ifconfig`, but I'm not sure that other Linux/BSD/Solaris systems have `ifconfig` in their `/sbin`, and I think adding `PATH="/sbin:$PATH"` to the commands would be ugly. This would also remove the requirement to use `/sbin` in `lspci` and Slackware's `pkgtool`.

virtualbox_debian kfreebsd_17_12_2016_15_13_20

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -1638,17 +1638,14 @@ get_battery() {
 
 get_local_ip() {
     case "$os" in
-        "Linux")
+        "Linux" | "BSD" | "Solaris")
             local_ip="$(ip route get 1 | awk '{print $NF;exit}')"
+            [[ -z "$local_ip" ]] && local_ip="$(ifconfig | awk '/broadcast/ {print $2}')"
         ;;
 
         "Mac OS X" | "iPhone OS")
             local_ip="$(ipconfig getifaddr en0)"
             [[ -z "$local_ip" ]] && local_ip="$(ipconfig getifaddr en1)"
-        ;;
-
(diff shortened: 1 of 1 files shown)


dylanaraps:
LGTM :+1:

Great stuff! :D

PR #545, opened 2016-12-18
General: Add more paths to PATH

dylanaraps:
## Description

This PR adds more paths to `$PATH`.

- `/opt`: Standalone programs are installed here.
- `/usr/local/bin`: Local programs are installed here.
- `/cygdrive/c/Windows/*`: Windows `cmd.exe` and `powershell.exe` programs.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -18,7 +18,7 @@ export LANG=C
 
 # Set PATH to binary directories only
 # This solves issues with neofetch opening the pacman game.
-export PATH="/usr/sbin:/usr/bin:/sbin:/bin"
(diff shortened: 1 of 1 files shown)


dylanaraps:
On second thought, this new PATH change might not be a good idea as the user might have something non-standard on their machine. I think we should go back to the old pacman fix.

konimex:
>On second thought, this new PATH change might not be a good idea as the user might have something non-standard on their machine.

Wait, what do you mean by this? Since when Neofetch interacts with outside programs, all (or most, as I don't know about where Cygwin keeps `wmic`) of them, one way or another are inside `/usr/bin`, `/usr/sbin`, `/bin`, or `/sbin`. Well, maybe count `/usr/local`.

dylanaraps:
The user might have programs we use in Neofetch installed outside of `/usr/sbin, /usr/bin, /sbin:/bin` and etc. This would cause Neofetch to break since it can't find the programs.

PR #546, opened 2016-12-18
General: Revert PATH changes and add /usr/sbin and /sbin to PATH

dylanaraps:
## Description

This reverts the recent pacman fix since the user could have a non-standard path which would break Neofetch. 
 
Also see: #545

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -17,7 +17,7 @@ export LC_ALL=C
 export LANG=C
 
 # Add /usr/sbin and /sbin to PATH.
-export PATH="${PATH}:/usr/sbin:/sbin"
+export PATH="/usr/sbin:/sbin:${PATH}"
 
 # Set no case match.
 shopt -s nocasematch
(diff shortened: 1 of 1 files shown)


neofetch
```diff
@@ -16,9 +16,8 @@ XDG_CONFIG_HOME="${XDG_CONFIG_HOME:-${HOME}/.config}"
 export LC_ALL=C
 export LANG=C
 
-# Set PATH to binary directories only
-# This solves issues with neofetch opening the pacman game.
-export PATH="/usr/sbin:/usr/bin:/sbin:/bin"
+# Add /usr/sbin and /sbin to PATH.
+export PATH="${PATH}:/usr/sbin:/sbin"
```
konimex: I think this is better the other way around.
`PATH="/usr/sbin:/sbin:${PATH}"` since Neofetch will mostly interact with `/usr/sbin`.

dylanaraps: :+1:

PR #547, opened 2016-12-18
OS: Added initial support for MINIX

konimex:
## Description

This adds initial support for MINIX.

## Issues

This only supports [Minix R3.4rc](http://download.minix3.org/iso/snapshot/minix_R3.4.0rc2-373b793.iso.bz2) and up, because `sysctl` was ported (from NetBSD!) just this January (before R3.4.0rc was released), however I'm yet to test this in MINIX 3.3 so maybe I can find a workaround (maybe using `top -d 1`).

Some functions also didn't appear because when a system is installed, it was a real barebone system. No `dig`, no `curl`, no `wget`.

virtualbox_minix3 clone_18_12_2016_21_52_11

## TODO

TBA

PR #549, opened 2016-12-19
OS: Fix some Minix bugs.

dylanaraps:
## Description

This PR fixes some of the bugs in @konimex's PR. #547 

## Features

- Term: Fixed bug with `init` appearing in `get_term()` output.
- Shell: Fixed bug with `sh` and `--version` causing an error.
- Local IP: Added initial support for Minix.
    - The Linux `ifconfig` command doesn't work for me, this change makes it work. I'm not sure if it already worked on your system or not @konimex. 
- Disk: Fixed `df` flags on Minix. Disk now works.

**Note:** Keep your PR open @konimex and we can merge mine into yours or you can manually add the changes since they're tiny.


konimex:
Merged to `minix` branch. 👍 

I will do some more tests when I get home later.

PR #552, opened 2016-12-19
General: [Minix] Fix unsupported escape sequences in Minix.

dylanaraps:
## Description

Check if `$TERM` is `minix` before printing the escape sequences. This fixes some garbage output in minix's virtual console.

This is the only solution to this problem I've found so far.


konimex:
Until we can find a better way, I guess this will do.

PR #554, opened 2016-12-22
GPU: Print each GPU on a different line

dylanaraps:
## Description

This PR makes `get_gpu()` display each GPU on a separate line on Linux systems. 

## Features

- Each GPU is printed on a separate line.

## Issues

- None yet.

## TODO

- [x] Testing
- [x] Add option to print `all`, `dedicated` or `integrated`.

CC @konimex,, can you test this?

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -969,9 +969,14 @@ get_gpu() {
             gpus=($(lspci -mm | awk -F '\\"|\\" \\"' '/"Display|"3D|"VGA/ {print $3 " " $4}'))
 
             # Number the GPUs if more than one exists.
-            (( "${#gpus[@]}" > 1 )) && gpu_num=0
+            (( "${#gpus[@]}" > 1 )) && gpu_num=1
 
             for gpu in "${gpus[@]}"; do
+                # GPU shorthand tests.
+                [[ "$gpu_show" == "dedicated" && "$gpu" =~ (i|I)ntel ]] || \
+                [[ "$gpu_show" == "integrated" && ! "$gpu" =~ (i|I)ntel ]] && \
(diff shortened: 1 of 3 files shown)


konimex:
It works!

2016-12-22-085505_804x536_scrot

For some reason, battery subtitle seems to be broke, though.

dylanaraps:
Is the battery subtitle broken in master?

konimex:
Yep, seems to be broken in master too.

dylanaraps:
Fixed.

neofetch
```diff
@@ -965,35 +965,51 @@ get_cpu_usage() {
...
+                    ;;
+                esac
+
+                if [[ "$gpu_brand" == "off" ]]; then
+                    gpu="${gpu/AMD }"
+                    gpu="${gpu/NVIDIA }"
+                    gpu="${gpu/Intel }"
+                fi
+
+                prin "GPU${gpu_num}" "$gpu"
```
konimex: You may want to change the `GPU` one into `${subtitle}`, like the battery function.

dylanaraps: Yup, just did. :)

konimex: That was fast....

konimex:
Also, it seems we're still printing the results starting from `GPU0`, should we add the value by 1 so we can start it from `GPU1`?

dylanaraps:
Sure, easy fix. :)

dylanaraps:
I've added `--gpu_type` which lets you pick `dedicated`, `integrated` or `all`. (`all` by default). Please test it out. :)

konimex:
LGTM :+1:

PR #556, opened 2016-12-22
Add support for Exaile music player

mstraube:
This PR adds song info support for Exaile.

There are also the commandline options` --get-title` and `--get-artist`, but `dbus-send` is much faster.
850ms vs 16ms on my system.


dylanaraps:
LGTM :+1:

Thanks!

PR #562, opened 2016-12-24
Term: Move pid handling to separate functions

dylanaraps:
## Description

This PR moves the PID / PPID handling from `get_term()` to two separate functions. This change was made so that we can do the same PID handling in another function. (`get_term_font`)

## TODO

- [x] Testing.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -3088,7 +3088,6 @@ get_process_name() {
     printf "%s" "$name"
 }
 
-
 # FINISH UP
 
 usage() { printf "%s" "\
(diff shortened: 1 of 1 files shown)


dylanaraps:
@mstraube you should be able to substitute the `child=` line with this once this PR is merged.

```sh
child="$(get_ppid "$$")"
```

mstraube:
I've tested the konsole font code with this, and it works. Note that I've recreated my konsole-font branch from updated master. That was easier for me than merging, squashig and resolving the conflicts. ;)

PR #564, opened 2016-12-24
Song: enforce order artist - title in get_song_dbus()

mstraube:
Some players report title before artist. That leads to song = titel - artist.
This PR makes sure it's always artist - title


dylanaraps:
LGTM :+1:

PR #567, opened 2016-12-25
General: Changed manpage generation command

konimex:
## Description

Most of the changes are cosmetics, except for the generation part.

The `./` before `neofetch` is added because when `help2man -N neofetch -o neofetch.1` is run, it generates the manpage for `/usr/bin/neofetch` instead.


dylanaraps:
LGTM :+1:

PR #569, opened 2016-12-26
Song: Fix Exaile song query

mstraube:
Exaile does't report artist and title if playback is stopped. That leads to `song=" -"`
Added a check to avoid printing this empty song info.


dylanaraps:
LGTM :+1:

PR #571, opened 2016-12-27
Speed up song query for some players

mstraube:
This PR speeds up song query for Pragha, Amarok and Banshee. They take half the time now. 
Most noticeable for Banshee because of it's slow query command. 

I've also changed the cmus query to avoid printing of an empty song info " - " if playback is stopped.
(Though it's only one call to `cmus-remote -Q` now, there's no significant speed increase.)

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -1249,9 +1249,7 @@ get_song() {
         ;;
 
         "amarok"*)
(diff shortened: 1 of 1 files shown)


dylanaraps:
LGTM :+1:

PR #572, opened 2016-12-27
OS: Added support for Gentoo/FreeBSD

konimex:
### Features

The ultimate aim of this PR is to expand Neofetch's support to Gentoo/FreeBSD. Testing will have to be done in these distros (in an installed environment, not a LiveCD):

- [x] FreeBSD / DragonFlyBSD
- [x] OpenBSD / Bitrig / NetBSD
- [ ] PacBSD
- [x] Gentoo/FreeBSD (see screenshot)
- [x] Solaris (OpenIndiana/Oracle)
- [x] Dragora

These testings are done because we're merging the package detection for Linux/Solaris and BSD. We all want accurate results for package, don't we?

### Issues

virtualbox_freebsd gentoo_27_12_2016_20_38_17

Also, see my comments in the diff.

### TODO

- Errors need to be suppressed.
- Cleanup `get_distro`, I think.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -3472,7 +3472,7 @@ main() {
     get_os
     get_default_config 2>/dev/null
     get_args "$@"
-    get_distro
+    get_distro 2>/dev/null
     get_bold
     get_distro_colors
 
--- a/neofetch
+++ b/neofetch
@@ -54,7 +54,7 @@ get_distro() {
 
     case "$os" in
         "Linux" | "BSD" | "MINIX")
-            if [[ "$(< /proc/version)" == *"Microsoft"* || "$(< /proc/sys/kernel/osrelease)" == *"Microsoft"* ]]; then ## ERRORS NEED TO BE SUPPRESSED!
+            if [[ "$(< /proc/version)" == *"Microsoft"* || "$(< /proc/sys/kernel/osrelease)" == *"Microsoft"* ]]; then
                 case "$distro_shorthand" in
                     "on")   distro="$(lsb_release -sir) [Windows 10]" ;;
                     "tiny") distro="Windows 10" ;;
@@ -109,11 +109,13 @@ get_distro() {
                 esac
 
                 # Workarounds for distros that go against the os-release standard.
-                [[ -z "${distro// }" ]] && distro="$(awk '/BLAG/ {print $1; exit}' /etc/*ease /usr/lib/*ease)"
-                [[ -z "${distro// }" ]] && distro="$(awk -F'=' '{print $2; exit}' /etc/*ease /usr/lib/*ease)"
+                [[ -z "${distro// }" ]] && distro="$(awk '/BLAG/ {print $1; exit}' /etc/os-release /usr/lib/os-release)"
+                [[ -z "${distro// }" ]] && distro="$(awk -F'=' '{print $2; exit}' /etc/os-release /usr/lib/os-release)"
 
             else
-                distro="$(cat /etc/*-release)" # UUOC, but using usual bash $(< *-file) returns "ambiguous redirect"
+                for release_file in /etc/*-release; do
+                    distro+="$(< "$release_file")"
+                done
                 if [[ -z "$distro" ]]; then
                     case "$distro_shorthand" in
                         "on" | "tiny") distro="$kernel_name" ;;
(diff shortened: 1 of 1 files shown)


neofetch
```diff
@@ -111,6 +111,21 @@ get_distro() {
                 # Workarounds for distros that go against the os-release standard.
                 [[ -z "${distro// }" ]] && distro="$(awk '/BLAG/ {print $1; exit}' /etc/*ease /usr/lib/*ease)"
                 [[ -z "${distro// }" ]] && distro="$(awk -F'=' '{print $2; exit}' /etc/*ease /usr/lib/*ease)"
+
+            else
+                distro="$(cat /etc/*-release)" # UUOC, but using usual bash $(< *-file) returns "ambiguous redirect"
```
dylanaraps: Yup, you can't use `<` on multiple files.

Does this work? (I'm not on my Linux machine right now)

```sh
release_files=(/etc/*-release)
distro="$(< "${release_files[@]}")"
```

neofetch
```diff
@@ -111,6 +111,21 @@ get_distro() {
...
+                    case "$distro_shorthand" in
+                        "on" | "tiny") distro="$kernel_name" ;;
+                        *) distro="$kernel_name $kernel_version" ;;
+                    esac
+                    distro="${distro/DragonFly/DragonFlyBSD}"
+
+                    # Workarounds for FreeBSD based distros.
+                    [[ -f "/etc/pcbsd-lang" ]] && distro="PCBSD"
+                    [[ -f "/etc/rc.conf.trueos" ]] && distro="TrueOS"
+                    [[ -f "/etc/pacbsd-release" ]] && distro="PacBSD" # /etc/pacbsd-release is an empty file
```
konimex: How are we going to detect these 3 distros, then?

neofetch
```diff
@@ -53,7 +53,7 @@ get_distro() {
     [[ "$distro" ]] && return
 
     case "$os" in
-        "Linux")
+        "Linux" | "BSD" | "MINIX")
             if [[ "$(< /proc/version)" == *"Microsoft"* || "$(< /proc/sys/kernel/osrelease)" == *"Microsoft"* ]]; then
```
konimex: We need to suppress errors caused by this.

neofetch
dylanaraps: Good point. 

We need to find a more reliable way of detecting them.

neofetch
dylanaraps: Does this work?:

```sh
if [[ "$(< /proc/version 2>/dev/null)" == *"Microsoft"* || "$(< /proc/sys/kernel/osrelease 2>/dev/null)" == *"Microsoft"* ]]; then
```

konimex: Nope.

dylanaraps: Adding `2>/dev/null` to `get_distro` on line 3485 should work.

dylanaraps: Now that we've added `2>/dev/null` to `get_distro` we can remove all traces of `2>/dev/null` from inside the function.

konimex: Looks like there are only one `2>/dev/null` instances.

konimex:
It seems ASCII detection of Gentoo is a bit broken when we use `distro_shorthand=off`. In Getnoo FreeBSD, the `PRETTY_NAME` is `Gentoo/BSD` while on Linux it's `Gentoo/Linux`.

pb

dylanaraps:
We're going to have modify the case for Gentoo.

Not on Linux right now so here's an example:

```sh
# This is in get_distro_colors()
# Just remove Gentoo from wherever it is now and add this in.
"Gentoo"*)
    set_colors X X X
    ascii_distro="gentoo"
;;
```

konimex:
Alright. Since the problem also exist in Gentoo Linux, I'm going to push that one change to master.

konimex:
So all tests are done and I'm ready to merge. I need only one more test from PacBSD.

AFAIK, PacBSD has both `pacman` and `pkg` installed, but only `pacman` is used. If `pkg` exists, I want to know if the output of `pkg info | wc -l` == 0. If the output is not 0, then we'll need to use `if [[ "$kernel_name" == "FreeBSD" && "$distro" != "PacBSD"* ]]` instead of using case.

dylanaraps:
LGTM :+1: 

We can fix PacBSD once the servers are online.

PR #575, opened 2016-12-28
CPU: Add support for displaying temp as Fahrenheit

dylanaraps:
## Description

This works on Linux and BSD ~~but it's untested on BSD.~~

Note: This probably won't be merged.

Note: I accidentally included #574 in this PR.


konimex:
Works in BSD :+1:

PR #580, opened 2016-12-30
Term: Rewrite Konsole font detection, closes #576

### LINKED ISSUE #576: Konsole font detection may fail on Ubuntu 16.04, maybe others too
Konsole font dectection may fail because it's possible that there is only `org.kde.konsole`
instead of `org.kde.konsole-$PPID`. It's also possible that both are present. I noticed this on Kubuntu 16.04.

I'm working on a solution that checks for both.

mstraube:
I've rewritten the font detection for Konsole. It now
- works for `org.kde.konsole` and `org.kde-konsole-$PPID`
- works if the profile filename differs from the profile name
- doesn't use hardcoded config file paths
- works for KDE5 and KDE4

Tested on Arch, Kubuntu, openSUSE, Slackware, KDE neon.

Note:
With Konsole versions affected by https://bugs.kde.org/show_bug.cgi?id=358083 (I think..)
it's possible that we have two profile files containing the same profile name.

I noticed this behavior (with a new created user) on Kubuntu 16.04:
Starting Konsole -> edit current profile -> close konsole -> restart it -> edit current profile
Produced two differnt profile files with the same profile name in it.

That's the only situation I was able to produce two different profile files containing the same profile name.
But it's possible. So we have to make sure we get a valid filename from the grep command. ( head -n 1 )

Issues:
Detecting the fallback font fails if no fixed font is defined in kdeglobals.
This is the case on fresh installations or for new created users for example.

Konsole internally uses `QFontDatabase::systemFont(QFontDatabase::FixedFont)`
or `KGlobalSettings::fixedFont()` in qt4 versions.

The font settings are written to kdeglobals once the user changes fonts in the KDE System-Settings GUI.

So if the user doesn't edit a profile or the system fonts, no font will be detected.
I'm afraid there is not much we can do..

Maybe someone else knows a way to get the fixed Qt system font without writing (e.g. C++ or python) small helper programs? I couldn't find any existing commands.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -1568,25 +1568,47 @@ get_term_font() {
         ;;
 
         "konsole"*)
+            get_konsole_profile() {
+                konsole_windows=($(qdbus "${1}" | awk '/Windows\//'))
+                for window in "${konsole_windows[@]}"; do
+                    konsole_session="$(qdbus "${1}" "${window}" currentSession)"
+                    if ((child == "$(qdbus "${1}" /Sessions/"${konsole_session}" processId)")); then
+                        konsole_profile="$(qdbus "${1}" /Sessions/"${konsole_session}" environment |\
+                        awk -F '=' '/KONSOLE_PROFILE_NAME/ {print $2}')"
+                        break
+                    fi
+                done
+            }
+
+            get_konsole_font() {
+                profile_path="$("${1}" --path data | awk -F':' '{print $1}')/konsole"
+
+                # Profile filename can differ from profile name, so find the filename first, take first match
+                # It's possible that we have more than one file containing the same profile name
+                profile_filename="$(grep -l "Name=${konsole_profile}" "${profile_path}"/*.profile | head -n 1)"
+                [[ "${profile_filename}" ]] && term_font="$(awk -F '=|,' '/Font=/ {print $2 " " $3}' "${profile_filename}")"
+
+                # If there are no profiles or no font is defined in the profile, Konsole falls back to system's monospace font
+                # This fails if no fixed font is defined in kdeglobals, Konsole internally uses
+                # QFontDatabase::systemFont(QFontDatabase::FixedFont) or KGlobalSettings::fixedFont() in qt4 versions
+                [[ ! "${term_font}" ]] && term_font="$(awk -F '=|,' '/fixed=/ {print $2 " " $3}' "$("${1}" --path config --locate kdeglobals)")"
+            }
+
             # Use Process ID from get_term().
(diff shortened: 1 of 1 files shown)


dylanaraps:
I really don't like how complex this has become. I think it's better if we remove Konsole support from `term_font()` entirely, at least until we come up with a simpler solution.

@konimex, @iandrewt, what do you think about this?

iandrewt:
If theres a way to simplify this then yeah, I'd go with that. But at 33 line additions just for one terminal emulator support, that's almost like `dbus` level horribleness

mstraube:
Yes, it's a bit complex. I looked at it again and removed fallback to kdeglobals (since it's useless if the user didn't edit the system font settings), removed kde4 support and switched back to the hardcoded default path for the profile files. 

I ended up with [this](https://gist.github.com/mstraube/074b1d278268925f528928977f656db3). Once the user has edited a profile and Konsole uses the default path it should work.

Happy new year! :)

dylanaraps:
I left some comments.

Happy new year. :)

mstraube:
Pushed the simplified version.

neofetch
```diff
@@ -1568,25 +1568,26 @@ get_term_font() {
...
+                    fi
+                done
+                [[ "$profile" ]] && break
             done
-            term_font="$(awk -F '=|,' '/Font=/ {print $2}' "${HOME}/.local/share/konsole/${profile}".profile)"
+
+            # We could have two profile files for the same profile name, take first match
+            profile_filename="$(grep -l "Name=${profile}" "${HOME}"/.local/share/konsole/*.profile)"
+            profile_filename="${profile_filename/$'\n'*}"
+            [[ "$profile_filename" ]] && term_font="$(awk -F '=|,' '/Font=/ {print $2}' "${profile_filename}")"
```
dylanaraps: Remove the `{}` from `$profile_filename`.

dylanaraps:
LGTM now :+1:

Is this ready to merge @mstraube?

mstraube:
Yes I think so, just one question. Should it also print the font size?
That would only require to change the print command to `{print $2 " " $3}`. I'm undecided...

mstraube:
Ok, updated.

I also thought about kde4 support. It could be done with a function similar to kde_config_dir() for the directory with the profiles. `kde_data_dir="$(kf5-config --path data)" `for example.

pro:
- support kde4
- don't use hardcoded path

con:
- needs extra function
- slower

But I guess there are not many distros, beside Slackware, that still have kde4?
For now I would leave it as it is.

dylanaraps:
:+1:

PR #584, opened 2017-01-02
Ascii: Glob match ascii art

dylanaraps:
## Description

The goal of this PR is to make Neofetch fuzzy match ascii art.

For example:

- `Linux Mint` --> `mint`
- `LinuxMint` --> `mint`
- `Linux Mint 17.3 x86_64` --> `mint`
- `Mint` --> `mint`

## Features

- Removes a call to `tr` on bash 3!

## Issues

- [x] `distro_small` doesn't work.
- [x] `distro_old` doesn't work.
- [x] `distro-variant` doesn't work.
- [ ] Windows 10 ascii art doesn't work.

## TODO

- [ ] Testing.
- [ ] Clean up `get_distro_colors()` since we don't need as many single cases anymore.
- [x] Add back error handling.

## Note

I'm not sure if this is the best way of doing this. A safer but more hard-coded way of fixing these issues would be adding modifications to `get_distro_colors()`. For example:

```sh
case "$ascii_distro" in
    "Arch"* | "Arch Linux"*)
        set_colors 4 7 1
        distro_file="arch"
    ;;

    "Fedora"* | "RFRemix"*)
        set_colors 4 7 1
        distro_file="fedora"
    ;;

    # This would continue for every distro.
    # ...
    # ...
esac
```

This is imo the better solution but would increase the script size a ton. The benefits are that we won't have anymore detection issues and we can further simplify the `get_ascii()` function. Removing all `tr` calls and file checks.

What do you think @iandrewt and @konimex ?

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -1944,6 +1944,8 @@ get_ascii() {
         for art in "${ascii_dir}/${ascii_distro:0:1}"*; do
             [[ "$ascii_distro" =~ ${art/*\/} ]] && { ascii="$art"; break; }
         done
+
+        [[ ! -f "$ascii" ]] && { to_off "Ascii: Ascii file not found, falling back to off mode."; return; }
     fi
 
     # Set locale to get correct padding.
--- a/neofetch
+++ b/neofetch
@@ -1940,10 +1940,15 @@ get_ascii() {
         # Remove 'linux' from 'linux mint'.
         [[ "$ascii_distro" =~ ^linux ]] && ascii_distro="${ascii_distro/linux }"
 
+        # If _value or -value was used, dom't loop.
+        [[ "$ascii_distro" =~ (-|_) ]] && { ascii="${ascii_dir}/${ascii_distro}"; ascii_loop="off"; }
+
         # ascii_dir/first_letter (example: 'ascii_dir/m'*)
-        for art in "${ascii_dir}/${ascii_distro:0:1}"*; do
-            [[ "$ascii_distro" =~ ${art/*\/} ]] && { ascii="$art"; break; }
-        done
+        if [[ "${ascii_loop:-on}" == "on" ]]; then
+            for art in "${ascii_dir}/${ascii_distro:0:1}"*; do
+                [[ "$ascii_distro" =~ ${art/*\/} ]] && { ascii="$art"; break; }
+            done
+        fi
 
         [[ ! -f "$ascii" ]] && { to_off "Ascii: Ascii file not found, falling back to off mode."; return; }
     fi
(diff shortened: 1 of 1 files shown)


neofetch
```diff
@@ -1922,34 +1922,28 @@ get_ascii() {
...
         else
             get_script_dir 2>/dev/null
+            ascii_dir="${script_dir}/ascii/distro"
+        fi
 
-            # If the ascii file doesn't exist fallback to text mode.
-            if [[ -f "$script_dir/ascii/distro/${ascii/ *}" ]]; then
-                ascii="$script_dir/ascii/distro/${ascii/ *}"
+        # Remove 'linux' from 'linux mint'.
+        [[ "$ascii_distro" =~ ^linux ]] && ascii_distro="${ascii_distro/linux }"
```
dylanaraps: This is needed so `linux mint` doesn't match the `linux` ascii art.

neofetch
```diff
@@ -1922,34 +1922,28 @@ get_ascii() {
...
 
-            else
-                to_off "Ascii: Ascii file not found, falling back to text mode."
-                return
-            fi
-        fi
+        # ascii_dir/first_letter (example: 'ascii_dir/m'*)
+        for art in "${ascii_dir}/${ascii_distro:0:1}"*; do
```
dylanaraps: We loop over the entire ascii art folder here but it's optimized so that we only loop over files that match the first letter of `$ascii_distro`.

For example:

```sh
# If the distro is Arch Linux then the real command will look like this:
for art in bla/bla/ascii/distro/a*; do
```

The speed is the same as the current ascii handling this way.

dylanaraps:
@konimex, can you test and see if the distros with detection issues are fixed? (Void, Gentoo variants and etc).

neofetch
```diff
@@ -1922,34 +1922,35 @@ get_ascii() {
...
-                ascii="$script_dir/ascii/distro/${ascii/ *}"
+        # Remove 'linux' from 'linux mint'.
+        [[ "$ascii_distro" =~ ^linux ]] && ascii_distro="${ascii_distro/linux }"
 
-            else
-                to_off "Ascii: Ascii file not found, falling back to text mode."
-                return
-            fi
+        # If _value or -value was used, dom't loop.
+        [[ "$ascii_distro" =~ (-|_) ]] && { ascii="${ascii_dir}/${ascii_distro}"; ascii_loop="off"; }
```
dylanaraps: All `-value` and `_value` variants have to be manually specified so it's ok to skip the loop here.

konimex:
Works in Gentoo/BSD, I still cannot test it in Void because I'm still reinstalling Void in a VM.

However it seems the detection stopped working in my main machine.
virtualbox_freebsd gentoo_02_01_2017_11_11_15
2017-01-02-110947_617x454_scrot

dylanaraps:
Damn, I think I'll start working on what I wrote about under `Note:` at the top. The size doesn't matter if it simplifies everything.

PR #585, opened 2017-01-02
Ascii: Simplify file handling.

dylanaraps:
## Description

This PR simplifies the file handling in `get_ascii()` by making `get_distro_colors()` explicitly set the ascii file to use. This speeds up the script and removes a `tr` call on systems using `bash 3`. This also allows us to more easily fix ascii detection on Void Linux and Gentoo/variant. 

This does make the function and script much larger but this is just a side effect of simplifying the file handling in `get_ascii()`. The script is actually faster with these changes!

## Features

- Removes a call to `tr`!
- Removes the assumption that `first word in distro --> ascii file name`.
- Removes the long lists in the case statement.
- Removes Travis ascii art since it's no longer used.

## Issues

- [x] `distro_variant` doesn't work.

## TODO

- [x] Alphabetize the distros to make it easier to edit them later on.
- [x] Make some distro matching less greedy.
    - [x] Mac OS
    - [x] Linux Mint
    - [x] OpenSUSE
    - [x] Red Hat
- [x] Testing
    - [x] Test all distros.
        - I'll write a script to loop over all ascii files.

## Note

@konimex, this is the other idea I had. This might not fix the detection issues yet as we'll need to update the matching for Void and Gentoo.


konimex:
Works fine in Void here. All Gentoo variants should be fine too.

virtualbox_void linux musl_02_01_2017_13_47_22

dylanaraps:
I think that's a bug, I'll take a look.

neofetch
```diff
@@ -2938,8 +2936,7 @@ get_distro_colors() {
             ascii_file="solaris"
         ;;
 
-        # TODO (Before PR is merged): Make this glob less greedy.
-        *"SUSE"*)
+        "openSUSE"* | "open SUSE")
```
konimex: Shouldn't we add `"SUSE"*` too? To increase compatibility with SUSE Linux Enterprise?

dylanaraps: Good idea.

PR #588, opened 2017-01-03
 Fix WM Theme detection for KDE5, closes #586

### LINKED ISSUE #586: KDE/Plasma 5 WM Theme detection fails
On my Arch system WM Theme detection fails for KDE/Plasma 5 and `~/.kde4/share/config/kdeglobals` is used instead of `~/.config/kdeglobals`. I guess the `.kde4` folder exists because some older applications I've installed created it, or for backwards compatibility?

 [This](https://github.com/mstraube/neofetch/commit/e90cf76440f6b2c778012ac46597bb3ca80f9638) might be a solution.

Some notes about the changes:
-  KDE / Plasme 5 default config dir seems to be just `~/.config` not `~/<kde*>/share/config`
- `kf5-config` has no `--localprefix` option so I changed the others to `--path config`. 
- `$KDE_CONFIG_DIR` is empty on my system. I thought about to set it to `$KDE_CONFIG_DIR/share/config`, but that would lead to a wrong path if it's value is `~/.confg`. I don't know if that could happen on a Plasma 5 system, so I commented it.
- I assumed kwinrc has only one ```library=org.kde.<theme>``` line. The section of my kwinrc looks like this (library=org.kde.breeze is the last line of the file). 
```
[org.kde.kdecoration2]
BorderSize=Normal
ButtonsOnLeft=MS
ButtonsOnRight=HIAX
CloseOnDoubleClickOnMenu=false
library=org.kde.breeze
```

Edit:
Just noticed that line [688](https://github.com/mstraube/neofetch/commit/e90cf76440f6b2c778012ac46597bb3ca80f9638#diff-e863270127ca6116fd30e708cdc582fcR688) can be removed if the slash is stripped in the variable assignments `kde_config_dir="${kde_config_dir/$'/:'*}"`

After line [1353](https://github.com/mstraube/neofetch/commit/e90cf76440f6b2c778012ac46597bb3ca80f9638#diff-e863270127ca6116fd30e708cdc582fcL1353) it's used unstripped...

mstraube:
Fixes KWin Theme detection and config directory for KDE/Plasma 5

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -685,13 +685,14 @@ get_wm_theme() {
 
         "KWin"*)
             kde_config_dir
-            kde_config_dir="${kde_config_dir%/}"
 
-            if [[ -f "$kde_config_dir/share/config/kwinrc" ]]; then
-                wm_theme="$(awk '/PluginLib=kwin3_/{gsub(/PluginLib=kwin3_/,"",$0); print $0; exit}' "$kde_config_dir/share/config/kwinrc")"
+            if [[ -f "${kde_config_dir}/kwinrc" ]]; then
+                wm_theme="$(awk '/theme=kwin4/{gsub(/theme=kwin4_decoration_qml_/,"",$0); print $0; exit}' "${kde_config_dir}/kwinrc")"
+                [[ -z "$wm_theme" ]] && wm_theme="$(awk '/library=org.kde/{gsub(/library=org.kde./,"",$0); print $0; exit}' "${kde_config_dir}/kwinrc")"
+                [[ -z "$wm_theme" ]] && wm_theme="$(awk '/PluginLib=kwin3_/{gsub(/PluginLib=kwin3_/,"",$0); print $0; exit}' "${kde_config_dir}/kwinrc")"
 
-            elif [[ -f "$kde_config_dir/share/config/kdebugrc" ]]; then
-                wm_theme="$(awk '/(decoration)/ {gsub(/\[/,"",$1); print $1; exit}' "$kde_config_dir/share/config/kdebugrc")"
+            elif [[ -f "${kde_config_dir}/kdebugrc" ]]; then
+                wm_theme="$(awk '/(decoration)/ {gsub(/\[/,"",$1); print $1; exit}' "${kde_config_dir}/kdebugrc")"
             fi
         ;;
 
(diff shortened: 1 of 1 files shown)


dylanaraps:
LGTM :+1:

neofetch
```diff
@@ -2884,20 +2885,26 @@ kde_config_dir() {
     if [[ "$kde_config_dir" ]]; then
         return
 
-    elif [[ -n "$KDE_CONFIG_DIR" ]]; then
-        kde_config_dir="$KDE_CONFIG_DIR"
+#    elif [[ -n "$KDE_CONFIG_DIR" ]]; then
+#        kde_config_dir="$KDE_CONFIG_DIR"
```
dylanaraps: Remove the commented lines here.

neofetch
```diff
@@ -2884,20 +2885,26 @@ kde_config_dir() {
...
+        kde_config_dir="${kde_config_dir/$'/:'*}"
 
     elif [[ -d "${HOME}/.kde4" ]]; then
-        kde_config_dir="${HOME}/.kde4"
+        kde_config_dir="${HOME}/.kde4/share/config"
 
     elif [[ -d "${HOME}/.kde3" ]]; then
-        kde_config_dir="${HOME}/.kde3"
+        kde_config_dir="${HOME}/.kde3/share/config"
     fi
```
dylanaraps: Can we not just move `kde_config_dir="${kde_config_dir/$'/:'*}"` to the bottom of the function? This way we only have it once instead of 3 times.

neofetch
```diff
@@ -685,13 +685,14 @@ get_wm_theme() {
 
         "KWin"*)
             kde_config_dir
-            kde_config_dir="${kde_config_dir%/}"
 
-            if [[ -f "$kde_config_dir/share/config/kwinrc" ]]; then
-                wm_theme="$(awk '/PluginLib=kwin3_/{gsub(/PluginLib=kwin3_/,"",$0); print $0; exit}' "$kde_config_dir/share/config/kwinrc")"
+            if [[ -f "$kde_config_dir/kwinrc" ]]; then
+                wm_theme="$(awk '/theme=kwin4/{gsub(/theme=kwin4_decoration_qml_/,"",$0); print $0; exit}' "$kde_config_dir/kwinrc")"
+                [[ ! "$wm_theme" ]] && wm_theme="$(awk '/library=org.kde/{gsub(/library=org.kde./,"",$0); print $0; exit}' "$kde_config_dir/kwinrc")"
```
dylanaraps: Can we change these to `[[ -z "$wm_theme" ]]`?

neofetch
mstraube: good idea

mstraube:
I updated the PR. I squashed the changes into one commit.
...I thought the comments would remain. :/

dylanaraps:
There's no point in keeping the comments there, if we ever need to re-visit the KDE stuff the commit history has those 2 lines stored.

LGTM :+1:

PR #596, opened 2017-01-05
GPU: Use readarray instead of modifying IFS

dylanaraps:
## Description

`readarray` works better here since it ignores `IFS` and splits elements by newline.

@konimex, can you test this on your multi GPU system?


konimex:
LGTM :+1:

PR #597, opened 2017-01-05
Song: Add support for Yarock and Qmmp

mstraube:
- Adds support for Yarock and Qmmp.
- Generally don't print empty song info `" - "`.
This way we don't have to care about that for every single player.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -1202,7 +1202,7 @@ get_memory() {
 
 get_song() {
     # This is absurdly long.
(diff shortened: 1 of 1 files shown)


neofetch
```diff
@@ -1266,6 +1268,8 @@ get_song() {
         ;;
     esac
 
+    [[ "$song" =~ ^[\ -]+$ ]] && song=""
```
dylanaraps: I don't think we need regex matching here.

Does this work?:

```sh
[[ "$song" == " - " ]] && unset -v song

# Safer alternative
[[ "$(trim "$song")" == " - " ]] && unset -v song
```

dylanaraps:
Other than my comment, LGTM! :+!:

neofetch
mstraube: I'm not sure, I'll do some test...

mstraube:
It's `[[ "$(trim "$song")" == "-" ]] && unset -v song`

I also changed the query command for deadbeef. `--nowplaying` is deprecated and sets song to `"nothing"` if playback is stopped.

All good now.

dylanaraps:
:+1:

PR #599, opened 2017-01-06
Disk: Add option to specify which disks to show

dylanaraps:
## Description

This PR adds a new option called `disk_show` which allows you to specify which disks you want to display in the info. (one per line).

## Features

- Display multiple disks (one per line).
- Subtitle is updated to display mount point `Disk (/)`, `Disk (/mnt/Local Disk)`.

## Issues

- The subtitle currently shows mount point, should we instead display the drive name? `/dev/sdXX` 

```sh
# Example:

# Mount point (Current behavior)
Disk (/mnt/Videos): 24G / 60G (45%)

# Drive name
Disk (/dev/sdc1): 24G / 60G (45%)
```

## TODO

- [x] Testing
    - [x] Linux
    - [x] MacOS
        - Broken in travis.
    - [x] BSD
        - [x] OpenBSD
        - [x] FreeBSD
        - [x] NetBSD
        - [x] DragonflyBSD
    - [x] MINIX 
    - [x] Windows
    - [x] Solaris

## Testing

```sh
# Example command.
neofetch --disk_show '/' '/dev/sdc1' '/mnt/Videos'

# Inside your neofetch config
disk_show=('/' '/dev/sdc1' '/mnt/Videos')
```

### DIFF
--- a/config/config
+++ b/config/config
@@ -316,6 +316,21 @@ public_ip_host="http://ident.me"
 #
 disk_show=('/')
 
+# Disk subtitle.
+# What to append to the Disk subtitle.
+#
+# Default: 'mount'
+# Values:  'mount', 'name'
+# Flag:    --disk_subtitle
+#
+# Example:
+# name:   'Disk (/dev/sda1): 74G / 118G (66%)'
+#         'Disk (/dev/sdb2): 74G / 118G (66%)'
+#
+# mount:  'Disk (/): 74G / 118G (66%)'
+#         'Disk (/mnt/Local Disk): 74G / 118G (66%)'
+disk_subtitle="mount"
+
 
 # Song
 
--- a/config/config
+++ b/config/config
@@ -32,7 +32,7 @@ print_info() {
     info "Memory" memory
 
     # info "CPU Usage" cpu_usage
-    # info "Disk (root)" disk
+    # info "Disk" disk
     # info "Battery" battery
     # info "Font" font
     # info "Song" song
@@ -230,6 +230,7 @@ gpu_brand="on"
 #   GPU1: Intel Integrated Graphics
 gpu_type="all"
 
+
 # Resolution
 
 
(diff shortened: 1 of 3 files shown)


konimex:
>The subtitle currently shows mount point, should we instead display the drive name? `/dev/sdXX`

I think we should add another option on whether we should show the mount point or the partition name.

konimex:
It seems to be broken in NetBSD, since NetBSD `df` doesn't support `-P` when `-h` is invoked.

dylanaraps:
Does it work if you change the df_flags section to this?

```sh
    case "$os" in
        "Haiku") err "Disk doesn't work on Haiku due to the non-standard 'df'"; return ;;
        "Minix") df_flags=(-h) ;;
        *"BSD")
            case "$kernel_name" in
                "NetBSD"*) df_flags=(-h) ;;
                *) df_flags=(-P -h) ;;
            esac
        ;;
        *) df_flags=(-P -h) ;;
    esac

```

dylanaraps:
This might be an issue with `df` on BSD in general seeing as it's also broken in OS X travis.

konimex:
You know, maybe I think we should get rid of `-P` altogether as it's conflicting with `-h`, [OpenBSD stated that `-h` is incompatible with `-P`](http://man.openbsd.org/OpenBSD-current/man1/df.1), while Oracle stated that [`-P` will show in the same format as `-h`, just using 512-byte blocks and not really human-readable (in GNU `df`, it's equivalent as running `df -P --block-size 512`)](https://docs.oracle.com/cd/E23823_01/html/816-5166/df-1m.html).

dylanaraps:
We're using `-P` to fix the output on OS X, I guess we can remove it for other OS.

See my comment here: https://github.com/dylanaraps/neofetch/pull/542#issuecomment-267589573

dylanaraps:
`df` doesn't seem to be working at all on macOS...

@iandrewt can you test this out?

dylanaraps:
I've figured out the issue, `readarray` doesn't work on MacOS...

Even the homebrew version of `bash` doesn't include it.

dylanaraps:
It works on MacOS now!

dylanaraps:
Also, the only problem with changing the subtitle to disk name is that we lose `/`, it'll instead show `/dev/sdXX`. What do you think @konimex?

konimex:
Like I said earlier, I think we should add another option on whether we should show the mount point or the partition name. Personally, though, I prefer to show the mount point.

dylanaraps:
This PR should be ready now.

PR #600, opened 2017-01-07
OS: Added support for Windows (MSYS2)

konimex:
## Description

This adds detection for any Windows systems using [MSYS2](https://msys2.github.io/) as opposed to Cygwin.

The changes are mostly compatibility changes because apparently MSYS2/MinGW's `wmic` doesn't support `/value`.

2017-01-07-130914_1920x1080_scrot


dylanaraps:
Awesome! LGTM

dylanaraps:
We should also update the README so people know we now support MSYS2.

PR #609, opened 2017-01-09
Term: Fix termite font detection, closes #608

### LINKED ISSUE #608: Terminal font detection for termite takes it from wrong place
## Description

Using [Termite](https://github.com/thestinger/termite), the script will fetch the font name from the `font` that is at the [`END`](https://github.com/dylanaraps/neofetch/blob/2d7114e03513ba7876934ed4597fabe0fe3461f8/neofetch#L1606)(?) of the file. 

A minimal version of the [default](https://github.com/thestinger/termite/blob/master/config) config looks like this.

```cfg
[options]
font = Monospace 9
[hints]
#font = Monospace 12
```

The first `font` option specifies the font used in the terminal, this is what neofetch *should* output. In this case, neofetch *will* output `Monospace 9`. But if we uncomment the `[hints]` `font` option, neofetch will output `Monospace 12` instead, which is **wrong**.

Neofetch should always output the `font` option that is under `[options]`, which should be at the beginning of the file. My bash/awk knowledge is limited, but as I said above, in the script `END` is mentioned. I assume this looks for `font` at the END of the file, not beginning.

#### Neofetch version

`2.0.2`

mstraube:
This should fix #608 

Works for me wether the `[hints]` section is above or below the `[options]` section. With and without commented `font=` in the `[hints]` section.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -1603,7 +1603,8 @@ get_term_font() {
         ;;
 
         "termite")
-            term_font="$(awk -F '= ' '/^font/ {a=$2} END{print a}' "${XDG_CONFIG_HOME}/termite/config")"
+            [[ -f "${XDG_CONFIG_HOME}/termite/config" ]] && config_file="${XDG_CONFIG_HOME}/termite/config"
+            term_font="$(awk -F '= ' '/\[options\]/ {opt=1} /^font/ {if(opt==1) a=$2; opt=0} END{print a}' "/etc/xdg/termite/config" "$config_file")"
         ;;
 
         "urxvt" | "urxvtd" | "rxvt-unicode" | "xterm")


dylanaraps:
LGTM :+1:

dylanaraps:
On second though, this is better since we only call `awk` once:

```sh
[[ -f "${XDG_CONFIG_HOME}/termite/config" ]] && config_file="${XDG_CONFIG_HOME}/termite/config"
term_font="$(awk -F '= ' '/\[options\]/ {opt=1} /^font/ {if(opt==1) a=$2; opt=0} END{print a; exit}' "$config_file" "/etc/xdg/termite/config")"
```
You can give `awk` multiple files but if a file doesn't exist it exits on an error. This checks to see if the user config file exists and if it does it looks there first.

mstraube:
Ok, but then the user config should be after the system-wide.
```bash
[[ -f "${XDG_CONFIG_HOME}/termite/config" ]] && config_file="${XDG_CONFIG_HOME}/termite/config"
term_font="$(awk -F '= ' '/\[options\]/ {opt=1} /^font/ {if(opt==1) a=$2; opt=0} END{print a; exit}' "/etc/xdg/termite/config" "$config_file")"
```
Otherwise the system-wide font is printed.

dylanaraps:
Good catch, thanks. In that case we can remove the file check altogether:

```sh
term_font="$(awk -F '= ' '/\[options\]/ {opt=1} /^font/ {if(opt==1) a=$2; opt=0} END{print a}' "$config_file" "/etc/xdg/termite/config" "${XDG_CONFIG_HOME}/termite/config")"\
```

Note: We'll have to modify the awk command to print the last match now. :P

mstraube:
I don't get it. This way it fails if the user config file does not exist
```bash
term_font="$(awk -F '= ' '/\[options\]/ {opt=1} /^font/ {if(opt==1) a=$2; opt=0} END{print a; exit}' "/etc/xdg/termite/config" "${XDG_CONFIG_HOME}/termite/config")"
```
How must the awk command be modfied?

dylanaraps:
Yup, my bad. 

I did some testing after I commented and noticed this too, we'll have to keep the file check in there.

mstraube:
Ok, then I'll update it.

dylanaraps:
LGTM :+1:

mstraube:
Hm, makes the `exit` in `END{print a; exit}` sense?  I think it could be removed.

dylanaraps:
:+1:

PR #617, opened 2017-01-15
Info: [WIP] Add info output formats

dylanaraps:
## Description

This PR adds every info function's output to an array which allows us to then output to different formats. This PR makes use of the existing user's `print_info()` function but I'd like to eventually simplify the `print_info()` syntax.

This PR currently includes the following formats:

- ` json`
- `yaml`
- `tab`: Print the info in a tab separated format.
- `*`: This is the default format and prints the info as neofetch always has.

## Features

- Adds `--info_format` to change the output format for the info.
- Adds `--json` to print the info in `json`.
- Adds `--yaml` to print the info in `yaml`.
- Adds `--tab` to print the info in a tab separated list`.

## Testing

Here's the basic usage:

```sh
# Output in format.
neofetch --info_format json

# Output in json.
neofetch --json

# Output in yaml.
neofetch --yaml

# Output in tab format.
neofetch --tab

# Sending output to a file
neofetch --json > neofetch.json
```

## Issues

- [ ] if a function fails to get the info or the info is empty, the subtitle is still shown.
- [x] `get_cols`: The additional rows of blocks aren't aligned properly.
- [x] `get_cols`: The additional rows of blocks aren't aligned properly in image off mode.
- [ ] `json` and `yaml` key names are subtitle instead of func_name.
- [x] Functions that declare a `func_run` variable don't work properly. 
- [ ] `tab`: Disable printing of color blocks, underlines and title.
- [x] Cursor position is broken in image mode.
    - This is most likely related to `$info_height`.

## TODO

- [ ] Clean this up.
    - This is still really rough.
- [ ] Add more info formats.
    - [x] `yaml`.
        - [ ] Handle multi line outputs properly.
    - [x] `json`.
        - [ ] Handle multi line outputs properly.
- [ ] Disable image printing in some modes.
    - the image shouldn't display `yaml` and `json` modes.
- [ ] Write docs.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -2552,6 +2552,29 @@ get_info() {
             done
         ;;
 
+        "yaml")
+            for output in "${info[@]}"; do
+                [[ "$output" != *:* ]] && continue
+
+                info="\"$(trim "${output/*:}")\""
+                printf "%s\n" "${output/:*}: $info"
+            done
+        ;;
+
+        "json")
+            printf "%s\n" "{"
+            for output in "${info[@]}"; do
+                [[ "$output" != *:* ]] && continue
+
+                subtitle="${output/:*}"
+                info="$(trim "${output/*:}")"
+
+                printf "\t%s\n" "\"${subtitle}\": \"${info}\","
+            done
+            printf "\t%s\n" "\"Version\": \"2.1.0\""
+            printf "%s\n" "}"
+        ;;
+
         *)
             for output in "${info[@]}"; do
                 if [[ "${output/*:*/:}" == ":" ]]; then
@@ -2576,7 +2599,7 @@ prin() {
     string="${string/:/${reset}${colon_color}:${info_color}}"
     string="${subtitle_color}${bold}${string}"
 
-    # print the info.
+    # Print the info.
     printf "%b\n" "${text_padding:+\033[${text_padding}C}${zws}${string}${reset} "
 }
 
@@ -4029,7 +4052,7 @@ main() {
     get_distro_colors
 
     # Minix doesn't support these sequences.
-    if [[ "$TERM" != "minix" ]]; then
+    if [[ "$TERM" != "minix" && ! "$info_format" =~ (tab|yaml|json) ]]; then
         # If the script exits for any reason, unhide the cursor.
         trap 'printf "\033[?25h\033[?7h"' EXIT
 
--- a/neofetch
+++ b/neofetch
@@ -3933,8 +3933,13 @@ get_args() {
                 colors+=(7 7 7 7 7 7)
             ;;
 
-            # Text Formatting
+            # Info formats
             "--info_format") info_format="$2" ;;
+            "--json") info_format="json" ;;
+            "--yaml") info_format="yaml" ;;
+            "--tab") info_format="tab" ;;
+
+            # Text Formatting
             "--underline") underline_enabled="$2" ;;
             "--underline_char") underline_char="$2" ;;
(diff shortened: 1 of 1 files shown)


konimex:
Few comments about using YAML.

For GPU, the current behavior is:

```yaml
GPU: "Intel Integrated Graphics"
GPU: "AMD Radeon HD 8670A/8670M/8690M"
```

This can be 

```yaml
GPU:
  - "Intel Integrated Graphics"
  - "AMD Radeon HD 8670A/8670M/8690M"
```

For disk, the current behaviour is

```yaml
Disk (/): "8.8G / 267G (4%)"
Disk (/home): "8.3G / 631G (2%)"
```

This can be changed to:

```yaml
Disk:
    /: "8.8G / 267G (4%)"
    /home: "8.3G / 631G (2%)"
```

Take this with a grain of salt as I don't know if YAML accepts slashes as key and how programs read YAML.

dylanaraps:
Thanks, I forgot all about the multi line functions. :+1:

dylanaraps:
I don't like how this PR has been implemented so far. We need to think about how this will be implemented.

## print_info() and info()

The `print_info()` function in the config will need to be changed to an array. The only issue with this is that it will break existing config files. I can look into adding backwards compatibility but I think it's better if make next release a major release. 

Following semantic versioning, next release will be `3.0` since it contains breaking changes. We can then also remove all of the backwards compatibility stuff and inform users that they must recreate their config file. I don't really want to do this but it seems like it's the only way to implement this PR properly.

An array better suits the modifications in this PR since:

1. We will know how many functions will be run beforehand.
2. We can loop over the functions properly.
    - Right now we're just calling a function over and over. (`info()`)
3. We can properly implement info alignment since we can count the subtitle lengths without a hack.
4. We can simplify `--disable`.
5. We can add an `--enable` flag since we can now append functions to the user's configuration.
    - With the current `print_info()` function we can't do this.

## Info functions.

There are a few options for handling function output, we need to think about which method we want to go with and the Pros/Cons of each one.

1. Each function prints as plaintext to stdout. 
    - `get_memory` --> `250MB / 1024MB`
    - PROS
        - When splitting the functions into multiple files this will eventually allow each info function to work standalone.
    - CONS
        - Multi-line function output doesn't really work with this method.
2. Each function adds to an array.
    - `get_memory` --> `info+=("250 MB / 1024MB")` --> `250MB / 1024MB`
    - PROS
        - This works with multi-line outputs.
    - **This seems like the best option**
3. Each function declares a variable.
    - `get_memory` --> `$memory` --> `250MB / 1024MB`
    - PROS
        - No additional lines need to be added to info functions.
    - CONS
        - Multi-line function output doesn't really work with this method.

## Special functions (get_cols, get_underline, get_title)

These functions need to be modified to print without escape sequences. Right now they all print without `info()` or `prin()` and handle their own coloring, padding and etc. 
 
If we modify them to print as plain text we can also append them to the `info=()` array like the other info functions.

dylanaraps:
## Proposed PR

- The `print_info()` function will become an array.

```sh
print_info=(
    "title"
    "underline"
    
    "OS: distro"
    "Model: model"
    # etc...
)
```

- Each function will append its output to an array called `info`
    - This is what this PR does now.
- Special functions will be modified to remove all escape sequences.
    - Escape sequences will be added back later in the `info_format` section.
- ???

I'll start working on this and we'll see how it turns out. Nothing is set in stone yet.

dylanaraps:
Closed in favor of #618

PR #620, opened 2017-01-17
Makefile: Move default neofetch config to /etc/

dylanaraps:
## Description

This PR changes the install location for the config file to `/etc/neofetch` instead of `/usr/share/neofetch`, this way the default config file acts as a system-wide config for neofetch.

## Notes

- I'm unsure about the Makefile now, is it ok to hard-code `/etc/` without a `$(PREFIX)`?
    - Should we add a `/usr/` prefix variable and a `/etc/` variable?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,16 +9,18 @@ all:
 
 install:
 	$(INSTALL_DIR) $(DESTDIR)$(PREFIX)/bin
+	$(INSTALL_DIR) $(DESTDIR)/etc/neofetch
 	$(INSTALL_DIR) $(DESTDIR)$(PREFIX)/share/neofetch/ascii/distro
 	$(INSTALL_DIR) $(DESTDIR)$(PREFIX)/share/man/man1
 	$(INSTALL_PROG) neofetch $(DESTDIR)$(PREFIX)/bin/neofetch
 	$(INSTALL_FILE) neofetch.1 $(DESTDIR)$(PREFIX)/share/man/man1/neofetch.1
-	$(INSTALL_FILE) config/config $(DESTDIR)$(PREFIX)/share/neofetch/config
+	$(INSTALL_FILE) config/config $(DESTDIR)/etc/neofetch/config
 	$(INSTALL_FILE) ascii/distro/* $(DESTDIR)$(PREFIX)/share/neofetch/ascii/distro
 
 uninstall:
 	$(RM) $(DESTDIR)$(PREFIX)/bin/neofetch
 	$(RM) $(DESTDIR)$(PREFIX)/share/man/man1/neofetch.1
 	$(RM) -r $(DESTDIR)$(PREFIX)/share/neofetch
+	$(RM) -r $(DESTDIR)/etc/neofetch
 
 
--- a/neofetch
+++ b/neofetch
@@ -3262,14 +3262,11 @@ get_script_dir() {
 }
 
 get_default_config() {
-    if [[ -f "/usr/share/neofetch/config" ]]; then
-        default_config="/usr/share/neofetch/config"
+    if [[ -f "/etc/neofetch/config" ]]; then
+        default_config="/etc/neofetch/config"
 
-    elif [[ -f "/usr/local/share/neofetch/config" ]]; then
-        default_config="/usr/local/share/neofetch/config"
-
-    elif [[ -f "/data/data/com.termux/files/usr/share/neofetch/config" ]]; then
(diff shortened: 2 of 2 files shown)


konimex:
My suggestions (and questions):

1. Get rid of `$(PREFIX)`, use `/usr` manually (as we would use `/etc` manually).
2. Why are we using `?=` ?

If possible, we can make it as POSIX-compliant as possible, and that is reducing any use of LSB-specific (`install`)/GNU-specific (that `?=` in the Makefile) extensions. Although this means we will need to use another utility, `chmod` (that is, if `-p` didn't preserve the attributes, but `man 1p cp` said that file permission bits should be preserved if `-p` was specified).

This is my rough "sketch" of what the new Makefile I proposed look like.

```make
all:
	@echo Run \'make install\' to install Neofetch

install:
	mkdir -p $(DESTDIR)/usr/bin
	mkdir -p $(DESTDIR)/usr/share/man/man1
	mkdir -p $(DESTDIR)/etc/neofetch
	mkdir -p $(DESTDIR)/usr/share/neofetch/ascii/distro
	cp -p neofetch $(DESTDIR)/usr/bin/neofetch
	cp -p neofetch.1 $(DESTDIR)/usr/share/man/man1/neofetch.1
	cp -p config/config $(DESTDIR)/etc/neofetch/config
	cp -p ascii/distro/* $(DESTDIR)/usr/share/neofetch/ascii/distro

uninstall:
	rm -f $(DESTDIR)/usr/bin/neofetch
	rm -f $(DESTDIR)/usr/share/man/man1/neofetch.1
	rm -f -r $(DESTDIR)/usr/share/neofetch
	rm -f -r $(DESTDIR)/etc/neofetch
```

I'll make a new branch and test this in various package managers and OSes.

dylanaraps:
In macOS, neofetch must be installed to `/usr/local/bin/` (`/usr/bin` is readonly iirc(?)) this is the main (if only) reason we have the `$(PREFIX)` variable. 

I can't remember why we're using `?=` but I'm all for removing it if it's GNU specific. :+1:

That Makefile looks good to me so far. :+1:

dylanaraps:
(Accidentally deleted my comment...)

> Get rid of $(PREFIX), use /usr manually (as we would use /etc manually).

The original reason for using `$(PREFIX)` was that on recent versions of macOS, the root account can't access `/usr/bin` (or similar). I'm not sure about how we fix this. \[1\]

> Why are we using ?= ?

I'm all for removing it if it's GNU specific :+1:

The Makefile looks good to me so far :+1:

\[1\] http://apple.stackexchange.com/a/193379

konimex:
I'm unfamiliar with how macOS works, but do macOS homebrew has to specify `$(PREFIX)`? If so, (maybe) we can have `PREFIX` set to `/usr` (`cp -p neofetch $(DESTDIR)$(PREFIX)/bin/neofetch` and so on). Does macOS (in this case, homebrew) override `$(PREFIX)` to `/usr/local`?

dylanaraps:
Yup, the homebrew formula sets `$PREFIX` to `/usr/local/`.

konimex:
So, this is my second proposal for Makefile.

```make
PREFIX = /usr

all:
	@echo Run \'make install\' to install Neofetch

install:
	mkdir -p $(DESTDIR)$(PREFIX)/bin
	mkdir -p $(DESTDIR)$(PREFIX)/share/man/man1
	mkdir -p $(DESTDIR)/etc/neofetch
	mkdir -p $(DESTDIR)$(PREFIX)/share/neofetch/ascii/distro
	cp -p neofetch $(DESTDIR)$(PREFIX)/bin/neofetch
	cp -p neofetch.1 $(DESTDIR)$(PREFIX)/share/man/man1/neofetch.1
	cp -p config/config $(DESTDIR)/etc/neofetch/config
	cp -p ascii/distro/* $(DESTDIR)$(PREFIX)/share/neofetch/ascii/distro

uninstall:
	rm -f $(DESTDIR)$(PREFIX)/bin/neofetch
	rm -f $(DESTDIR)$(PREFIX)/share/man/man1/neofetch.1
	rm -f -r $(DESTDIR)$(PREFIX)/share/neofetch
	rm -f -r $(DESTDIR)/etc/neofetch
```

I'm testing it for Gentoo (ebuild) and Oracle Solaris, but this should be fine in any *NIX systems.

dylanaraps:
LGTM :+1:

konimex:
Should I push the change directly to this branch?

dylanaraps:
Sure :+1:

PR #622, opened 2017-01-18
Config: Move default config to /etc (#620)

konimex:
## Description

Self-titled.

If this looks good enough, please merge it (and update the changelogs accordingly) as I won't be able to open my desktop for the next 6-8 hours.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,24 +1,20 @@
-PREFIX        ?=  /usr
-RM            ?=  rm -f
-INSTALL_DIR   ?=  install -m755 -d
-INSTALL_PROG  ?=  install -m755
-INSTALL_FILE  ?=  install -m644
+PREFIX = /usr
 
 all:
 	@echo Run \'make install\' to install Neofetch
 
 install:
-	$(INSTALL_DIR) $(DESTDIR)$(PREFIX)/bin
-	$(INSTALL_DIR) $(DESTDIR)$(PREFIX)/share/neofetch/ascii/distro
-	$(INSTALL_DIR) $(DESTDIR)$(PREFIX)/share/man/man1
-	$(INSTALL_PROG) neofetch $(DESTDIR)$(PREFIX)/bin/neofetch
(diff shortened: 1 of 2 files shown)


dylanaraps:
We can remove the check for `/usr/local/share` now since we're only installing the config to `/etc/`. Other than that LGTM. I can make this change after merging this if you'd like. :+1:

konimex:
Please do. :+1:

dylanaraps:
One thing: 

```sh
PREFIX=/usr/local make install
```

This doesn't work, neofetch is still put in `/usr/bin`. I'm pretty sure this is happening because the `PREFIX = /usr` part of the Makefile overwrites the user's env. This is also probably the reason why we were using `=?`. I'll look into this.

konimex:
How about `make PREFIX=/usr/local install`?

On 18 January 2017 8:56:24 AM GMT+07:00, Dylan Araps <notifications@github.com> wrote:
>One thing: 
>
>```sh
>PREFIX=/usr/local make install
>```
>
>This doesn't work, neofetch is still put in `/usr/bin`. I'm pretty sure
>this is happening because the `PREFIX = /usr` part of the Makefile
>overwrites the user's env. This is also probably the reason why we were
>using `=?`. I'll look into this.
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/dylanaraps/neofetch/pull/622#issuecomment-273359994

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

konimex:
Ah, yes. Please do.

On 18 January 2017 8:34:08 AM GMT+07:00, Dylan Araps <notifications@github.com> wrote:
>We can remove the check for `/usr/local/share` now since we're only
>installing the config to `/etc/`. Other than that LGTM. I can make this
>change after merging this if you'd like. :+1:
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/dylanaraps/neofetch/pull/622#issuecomment-273356364

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

dylanaraps:
That worked. :+1:

PR #624, opened 2017-01-18
Term Font: Add support for deepin-terminal

mstraube:
## Description

Adds font detection for deepin-terminal.
Tested on Arch and Deepin.


dylanaraps:
:+1:

PR #625, opened 2017-01-19
OS: Add initial support for HP-UX 11.31+

dylanaraps:
## Description

Title.

## TODO

I've excluded functions that only require an X server or work on all OS already.

- [x] `get_os()`
- [x] `get_distro()`
- [x] `get_model()`
- [x] `get_kernel()`
- [x] `get_uptime()`
- [x] `get_packages()`
- [ ] `get_cpu()`
    - Partial support.
- [ ] `get_gpu()`
    - Is this necessary?
- [ ] `get_memory()`
    - Partial support.
    - `vmstat` looks like the easiest way to do this.
- [x] `get_disk()`
- [x] <s>`get_battery()`</s>
- [ ] `get_local_ip()`
    - `ifconfig` on `HP-UX` doesn't have `-a`.
- [x] `get_install_date()`

## Note

All of this is untested, I'm going off of google search results.


konimex:
So I found [this](http://ptpb.pw/FrrS) while lurking into HP forums. Might be useful (also, it is a bit painful to see different versions have different commands).

dylanaraps:
That link is helpful, thanks for finding it.

I'm not sure about the differences between each version but what has been implemented so far \*should\* work on all versions of HP-UX. If we run into issues, I think we should support only the latest version of HP-UX, what do you think?

konimex:
The latest version is `11.31`, so yeah, latest version it is.

dylanaraps:
I'm going to rent a HP-UX server next pay day, then we can merge this. :)

konimex:
There are some problems in finding some info (such as CPU and memory).

1. Different arch requires different commands.
    This can be solved by case statements.
2. Some commands require root.

Also, I can only test on IA64 and have no access to any HP-UX PA-RISC systems yet.

konimex:
I may need to find an alternative (or wait for someone with root access) to gather infromation. `machinfo` is not an optimal command since it can change in next release (or "product updates").

From machinfo(1):

>The machinfo command output is not suitable for parsing by scripts since its content and format is subject to change based on product updates, os version, hardware platform, and/or system configuration

nfinit:
I've been working with an older Visualize C3000 running 11.11 and yesterday attempted to kludge out a solution for `get_cpu()` that would work with it. Since I don't personally have any newer Itanium systems, I've left that alone for now since it seems like the `model` output needs to be parsed differently, but on PA-RISC systems it's a fixed `9000/[family]/[speed|submodel]` format that you can leverage nicely.

This is what I've managed to get going, it works correctly on my system and should work on any 7xx series PA-RISC system running 11.00 or later:
```
"HP-UX")
            # Get CPU name.
            # Approach varies between 9000 (PA-RISC) systems
	    # and ia64 (Integrity) systems.
	    machine_type="$(echo `model` | cut -c1-4)"
	    if [ "$machine_type" == "9000" ]; then
		chip_num=$(getconf CPU_CHIP_TYPE)
		chip_bin=`echo "obase=2;$chip_num" | bc | sed -e 's/.....$//'`
		chip_type=`echo "obase=16;ibase=2;$chip_bin" | bc`
		case "$chip_type" in
		    b)  cpu="HP PA-7200"   ;;
		    d)  cpu="HP PA-7100LC" ;;
		    e)  cpu="HP PA-8000"   ;;
		    f)  cpu="HP PA-7300LC" ;;
		    10) cpu="HP PA-8200"   ;;
		    11) cpu="HP PA-8500"   ;;
		    12) cpu="HP PA-8600"   ;;
		    13) cpu="HP PA-8700"   ;;
		    14) cpu="HP PA-8800"   ;;
		    15) cpu="HP PA-8750"   ;;
		    *)  cpu="HP PA-RISC"   ;;
		 esac
	    elif [ "$machine_type" == "ia64" ]; then
		# WIP 
		cpu="Intel Itanium"
	    fi

            # Get CPU speed.
            # Also varies between the two system types.

	    if [ "$machine_type" == "9000" ]; then
		# PA-RISC systems generally have one CPU to one model
		# with the exception of early (i.e. 712, 715) systems
		# and the C8000 which ships with multiple speed grades
		# of the PA-8800 or PA-8900
		machine_model="$(echo `model` | awk -F'/' '{print $2}')"
		machine_sub="$(echo `model` | awk -F'/' '{print $3}')"
		case "$machine_model" in
		    770) case "$machine_sub" in
			    J200)   speed=100 ;;
			    J210)   speed=120 ;;
			    J210XC) speed=120 ;;
			 esac ;;
		    777) case "$machine_sub" in
			    C100) speed=100 ;;
			    C110) speed=120 ;;
			 esac ;;
		    778) case "$machine_sub" in
			    B132L)  speed=132 ;;
			    B132L+) speed=132 ;;
			    B160L)  speed=160 ;;
			    B180L+) speed=180 ;;
			 esac ;;
		    779) case "$machine_sub" in
			    C132L) speed=132 ;;
			    C160L) speed=160 ;;
			 esac ;;
		    780) case "$machine_sub" in
			    C160) speed=160 ;;
			    C180) speed=180 ;;
			    J280) speed=180 ;;
			    J282) speed=180 ;;
			 esac ;;
		    782) case "$machine_sub" in
			    C200)  speed=200 ;;
			    C240)  speed=236 ;;
			    J2240) speed=236 ;;
			 esac ;;
		    785) case "$machine_sub" in
			    B1000) speed=300 ;;
			    B2000) speed=400 ;;
			    B2600) speed=500 ;;
			    C360)  speed=367 ;;
			    C3000) speed=400 ;;
			    C3600) speed=552 ;;
			    C3650) speed=625 ;;
			    C3700) speed=750 ;;
			    C3750) speed=875 ;;
			    J5000) speed=440 ;;
			    J5600) speed=552 ;;
			    J6000) speed=552 ;;
			    J6700) speed=750 ;;
			    J6750) speed=875 ;;
			    J7000) speed=440 ;;
			    J7600) speed=552 ;;
			 esac ;;
		esac
	    fi

            # Get CPU cores.
            cores="$(ioscan -k | grep -c "processor")"
        ;;
```
It's a bit more of a proof-of-concept, since I don't know if this is up to the standards of this project as it uses hard-coded values and outsources some work to `bc` which may not be on every system, but it seems like there isn't a reliable way to directly obtain the speed of the processor as a non-root user in a way that can be parsed on a wide variety of systems, so instead I figured I could take advantage of the fact that most PA-RISC systems had generally fixed configurations (CPU-wise) that were very unlikely to be subject to change. You might, say, be able to pull a 440 MHz PA-8500 out of a J5000 and install it in a C3000, but I've never personally seen it done, although I'm not going to act like my experience isn't limited in that matter.

Looking back on the `bc` dependency issue, I had initially just had the goal of [re-implementing this script I found](http://unix.derkeiler.com/Mailing-Lists/HP-UX-Admin/2003-04/0003.html) and was hoping I could find some way to get the clock frequency directly, but to my knowledge thus far it doesn't seem possible. At this point, if I'm already getting the speed using this method, it might be more effective to leverage it to set the CPU, too. Supposedly the current implementation has compatibility problems with HP-UX 10.20 as well.

If you have any suggestions on how I might be able to improve or refactor this, I'm more than happy to hear it.

dylanaraps:
You can remove the dependency on `bc` by faking floating point math in bash. We do this throughout the script and it's accurate enough to work.

You basically just strip the decimal points from the numbers and treat them as larger whole numbers. See this example script I wrote:

```sh
#!/usr/bin/env bash
#
# floating point math in pure bash.

echo "> 100 * 0.2"
echo "bc: $(echo '100 * 0.2' | bc)"
echo "bash: $((100 * 2 / 10))"

echo

echo "> 340.22 / 10"
echo "bc: $(echo '340.22 / 10' | bc)"
echo "bash: $((34022 / 1000))" # Treat the number as a whole.

echo

echo "> 2000.33 / 2.5"
echo "bc: $(echo '2000.33 / 2.35' | bc)"
echo "bash: $((200033 / 235))" # Treat the numbers as a whole.
```

You can remove the dependency on `cut` by using bash's builtin string manipulation to cut strings.

```sh
var="123456"
var="${var:0:4}"
echo "$var" # output: 1234
```

PR #627, opened 2017-01-23
Intel GPU Info

SomaUlte:
## Description

Add better support for integrated Intel GPUs

## Features

Uses glxinfo to find the OpenGL renderer when using hardware rendering to display more detailed info about integrated intel GPUs

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -1075,7 +1075,11 @@ get_gpu() {
                     ;;
 
                     *"intel"*)
-                        gpu="Intel Integrated Graphics"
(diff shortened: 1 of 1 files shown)


dylanaraps:
You can structure it like this to avoid calling `glxinfo` twice:

```sh
gpu="$(glxinfo | grep -F "Device:")"
gpu="${gpu/Device:}"
gpu="${gpu/'(R)'}"
gpu="${gpu/\(*}"

[[ -z "$gpu" ]] && gpu="Intel Integrated Graphics"
```

This also removes the `sed` commands. `bash` can do them builtin much faster. :)

PR #632, opened 2017-01-23
MPD server on another host

dawidd6:
## Description

Well, if i remember correctly this issue was here in the past and now is present again.

What if i'm using mpc on my computer, to control mopidy on rpi? This commit is fixing this issue, maybe it's not good optimized, i've done my best i think. Simply neofetch doesn't display SONG when it doesn't find any player on host.

This screenshot shows output of command used to determine which player is currently opened on host.

zrzut ekranu z 2017-01-23 22-30-39

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -1332,6 +1332,10 @@ get_song() {
             song="$(dbus-send --print-reply --dest=net.sacredchao.QuodLibet /net/sacredchao/QuodLibet net.sacredchao.QuodLibet.CurrentSong |\
                     awk -F'"' '/artist/ {getline; a=$2} /title/ {getline; t=$2} END{print a " - " t}')"
         ;;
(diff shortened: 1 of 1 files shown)


dylanaraps:
We can change the `mpc` line to this:

```sh
mpc >/dev/null 2>&1 && song="$(mpc current)"
```

This way the song is only set if `mpc` connects to `mpd` successfully.

dawidd6:
Done as suggested

>this **2**/dev/null was a mistake ofc :+1:

dylanaraps:
Thanks, I'll merge this once travis is finished. :+1:

dylanaraps:
macOS travis seems to be stuck so I'll merge this anyway. (Linux check finished with no errors)

PR #639, opened 2017-01-25
Fix paths for "proper" Haiku installation

YellowApple:
## Description

Adds support for `neofetch` when installed on Haiku with `PREFIX=/boot/home/config/non-packaged make install` (as is convention)

## Features

`neofetch` now properly loads ASCII art and default configs when installed "properly" on Haiku (without being packaged).

## Issues

Would probably be better to have a `$PREFIX` variable defined somewhere, to be pre-populated by running `PREFIX=/foo/bar/baz make install`.  Not entirely sure how feasible or desirable that would be versus this sort of one-off case.

## TODO

See "Issues".

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -2030,6 +2030,9 @@ get_ascii() {
 
         elif [[ -d "/data/data/com.termux/files/usr/share/neofetch/ascii/distro" ]]; then
             ascii_dir="/data/data/com.termux/files/usr/share/neofetch/ascii/distro"
+            
+        elif [[ -d "/boot/home/config/non-packaged/share/neofetch/ascii/distro" ]]; then
+            ascii_dir="/boot/home/config/non-packaged/share/neofetch/ascii/distro"
 
         else
             [[ -z "$script_dir" ]] && script_dir="$(get_full_path "$0")"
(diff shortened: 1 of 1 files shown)


konimex:
[We can detect OS via makefile](http://stackoverflow.com/questions/714100/os-detecting-makefile), I'll look into this.

Anyway, the proper make command will be `make PREFIX=/usr/foo/bar install`.

dylanaraps:
We could also possibly modify the Makefile to store the `$PREFIX` value in a file. Then when sourcing the default config and ascii art we simply look at the saved prefix value. This way we don't hard-code a list of install locations for everything.

There's probably a better way of doing this though.

konimex:
I made https://github.com/konimex/neofetch/commit/cb48262f25bff395291f1a920904c9db4e867d81, but this currently only works with GNU make. I hate the BSDs and Solaris sometimes.

dylanaraps:
Are there any other projects that have a similar installation process? We could have a loot at other projects to see how they've solved this.

konimex:
[cfg2html](https://github.com/cfg2html/cfg2html) uses separate makefiles for separate OS.

dylanaraps:
That seems overly complicated (at least for us), cfg2html already has separate scripts for each OS so it makes sense there.

konimex:
Due to [the lack of standardized makefile formatting](https://stackoverflow.com/questions/9096018/make-gmake-compatible-if-else-statment), the best we can do is add a note for Haiku, I've added it in the Installation Wiki.

dylanaraps:
Thanks for the PR!

PR #649, opened 2017-01-27
Fix Alacritty font detection

siiptuo:
The default configuration of Alacritty happens to include a comment, so
the output is the following:

> Terminal Font: monospace # should be "Menlo" or something on macOS.

This pull request fixes this and the output is as expected:

> Terminal Font: monospace


dylanaraps:
LGTM :+1:

PR #651, opened 2017-01-29
ascii/distro/openbsd: removed extraneous backslashes to correct align…

Head-on-a-Stick:
…ment.

## Description

The (large) OpenBSD ascii logo had some extra backslashes added that were throwing out the alignment, I have removed these and it seems to work now.

Tested on OpenBSD-current


dylanaraps:
LGTM :+1:

PR #658, opened 2017-02-22
WM: Add support for Kwm on macOS

jorgegonzalez:
## Description

This PR adds support for [Kwm](https://github.com/koekeishiya/kwm).

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -631,7 +631,7 @@ get_wm() {
         case "$os" in
             "Mac OS X")
                 wm="Quartz Compositor"
-                [[ "$(ps -e | grep '[k]wm')" ]] && wm="Kwm"
+                ps -e | grep -F -q "kwm" && wm="Kwm"
             ;;
             "Windows")
                 wm="$(tasklist | grep -m 1 -o -F -e "bugn" -e "Windawesome" -e "blackbox" -e "emerge" -e "litestep")"
--- a/neofetch
+++ b/neofetch
@@ -629,7 +629,10 @@ get_wm() {
 
     else
         case "$os" in
-            "Mac OS X") wm="Quartz Compositor" ;;
+            "Mac OS X")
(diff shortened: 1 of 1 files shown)


dylanaraps:
Change the `ps` line to this: 

```sh
ps -e | grep -F -q "kwm" && wm="Kwm"
```

Otherwise LGTM :+1:

jorgegonzalez:
Done.

dylanaraps:
Thanks!

jorgegonzalez:
They always seem to fail on my PRs. Would you happen to know why?

dylanaraps:
Travis runs shellcheck and travis errors if shellcheck doesn't like anything.

jorgegonzalez:
So I made the change you suggested without actually testing it.

`[[ "$(ps -e | grep '[k]wm')" ]] && wm="Kwm"` -> `ps -e | grep -F -q "kwm" && wm="Kwm"`

This change doesn't work as intended.

dylanaraps:
I thought you would've tested it before changing the PR.

What doesn't work exactly?

jorgegonzalez:
My apologies.

The line `ps -e | grep -F -q "kwm" && wm="Kwm"` sets the window manager to "Kwm" even if it isn't actually running as a process.

This happens because the grep process always contains the `"kwm"` string when searching for it. `grep  "[k]wm"` prevents that process from showing up in the output.

https://unix.stackexchange.com/questions/74185/how-can-i-prevent-grep-from-showing-up-in-ps-results?noredirect=1&lq=1

`ps -e | grep -q '[k]wm' && wm="Kwm"`  is the proper fix.

dylanaraps:
That was my fault, don't worry.

Thanks for finding the correct fix. :)

Feel free to open another PR and I'll merge it. :+1:

jorgegonzalez:
Done.

PR #665, opened 2017-02-28
WM: Add support for Spectacle on macOS

jorgegonzalez:
## Description

This PR adds support for the [Spectacle](https://github.com/eczarny/spectacle) window manager.

## Issues

If Spectacle and Kwm are both running, Kwm pretty much overtakes everything and Spectacle does nothing, which is why detection for it happens beforehand.

screen shot 2017-02-28 at 11 43 00 am


dylanaraps:
:+1: LGTM

PR #668, opened 2017-03-02
Add font detection for iTerm2 on macOS

jorgegonzalez:
## Description

This PR adds font detection for iTerm2.

## Issues

There is a somewhat noticeable lag, possibly due to the regex. Is there any way we could speed this up a bit more?

Also this assumes that the user only has one iTerm2 profile. Unfortunately there is no way to determine which iTerm2 profile (and therefore which font) is currently being used. If the user has more than one profile, this may output the wrong font.

screen shot 2017-03-02 at 9 06 18 am


jorgegonzalez:
Closing in favor of #670.

PR #670, opened 2017-03-02
Add iterm2 font detection

### LINKED ISSUE #664: Add font detection for iTerm2 on macOS
## Description

I think this feature is long overdue and would be appreciated by all Neofetch users that use iTerm.

The command `defaults read ~/Library/Preferences/com.googlecode.iterm2.plist | grep 'Normal Font'` will return a line of the following form: `"Normal Font" = "Lemonmedium 14";`.

I've been trying to do this for a while but I don't know Bash well enough to format the output correctly. Ideally the conversions should look like this: 

```
"Normal Font" = "RobotoMono-Regular 13";     -> Roboto Mono
"Normal Font" = "SourceCodePro-Regular 13";  -> Source Code Pro
"Normal Font" = "AndaleMono 13";             -> Andale Mono
"Normal Font" = "Lemonmedium 14";            -> Lemon
```

Of course displaying anything at all is better than nothing, so even `RobotoMono` or `SourceCodePro` is (at least for me) a good start.

dritter:
This PR outputs the font of the currently active profile of iTerm2. I have written this in the last couple of weeks for a debugging script in [powerlevel9k](https://github.com/bhilburn/powerlevel9k) (debugging script not yet online) when I discovered #664 , so I wanted to contribute it back to here..
This is a different approach than @jorgegonzales uses in #668 (sorry). And sorry @jorgegonzales for entering the discussion so late..
This fixes #664

## How it works
  1. We ask iTerm2 via `osascript` to tell us the currently active [profile name](http://www.iterm2.com/documentation-scripting.html)
  2. We loop over all profiles
  3. If there is a profile with the same name found, then we parse the relevant information via `PlistBuddy`

## Disclaimer
It does **not** work when
  - iTerm2s [dynamic configuration](https://www.iterm2.com/documentation-dynamic-profiles.html) feature is used.

  - The Profile name is empty

  - a non distinct profile name is used. Unfortunately, the only way to get the currently active profile is via its name, which is ambiguous. See my question at the [iTerm2 discussion group](https://groups.google.com/forum/#!topic/iterm2-discuss/0tO3xZ4Zlwg) and at the [gitlab bugtracker](https://gitlab.com/gnachman/iterm2/issues/5586).

## What could be done better?
  - I wanted to keep the code dry, and refactored the plist path into a variable. Unfortunately this triggered an error, so I left it like it is now.

  - IMHO it would be nicer instead of using `PlistBuddy` to rely on `defaults read`, but I cannot get it to output the value of nested keys (e.g. `defaults read -app iTerm "/New Bookmarks/0/Normal Font"`).

  - Output could be nicer, as suggested in the issue. Currently the output is raw. Reason is, I am not a big fan of recreating the fonts name via regex. This should be done by querying the O's font daemon (if this is even possible).

- performance could be better. There are a lot of queries to PlistBuddy. Maybe there is a way to query multiple keys at once..

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -1659,9 +1659,11 @@ get_term_font() {
             # Count Guids in "New Bookmarks"; they should be unique
             local profiles_count
             profiles_count="$(/usr/libexec/PlistBuddy -c "Print :New\ Bookmarks:" ~/Library/Preferences/com.googlecode.iterm2.plist 2>/dev/null | grep -c "Guid")"
+
             for ((i=0; i<=profiles_count; i++)); do
                 local profile_name
                 profile_name="$(/usr/libexec/PlistBuddy -c "Print :New\ Bookmarks:${i}:Name:" ~/Library/Preferences/com.googlecode.iterm2.plist 2>/dev/null)"
+
                 if [[ "$profile_name" == "$current_profile_name" ]]; then
                     # "Normal Font"
                     term_font="$(/usr/libexec/PlistBuddy -c "Print :New\ Bookmarks:${i}:Normal\ Font:" ~/Library/Preferences/com.googlecode.iterm2.plist)"
@@ -1670,11 +1672,13 @@ get_term_font() {
                     # the "use a different font for non-ascii text" switch.
                     local use_different_font
                     use_different_font="$(/usr/libexec/PlistBuddy -c "Print :New\ Bookmarks:${i}:Use\ Non-ASCII\ Font:" ~/Library/Preferences/com.googlecode.iterm2.plist)"
+
                     if [[ "$use_different_font" == "true" ]]; then
                         local non_ascii_font
                         non_ascii_font="$(/usr/libexec/PlistBuddy -c "Print :New\ Bookmarks:${i}:Non\ Ascii\ Font:" ~/Library/Preferences/com.googlecode.iterm2.plist)"
                         [[ "$term_font" != "$non_ascii_font" ]] && \
                             term_font="$term_font (normal) / $non_ascii_font (non-ascii)"
+
                     fi
                 fi
             done
--- a/neofetch
+++ b/neofetch
@@ -1655,6 +1655,12 @@ get_term_font() {
                     # "Normal Font"
                     term_font=$(/usr/libexec/PlistBuddy -c "Print :New\ Bookmarks:${idx}:Normal\ Font:" ~/Library/Preferences/com.googlecode.iterm2.plist)
                     # Font for non-ascii characters
+                    # This font setting may differ from normal font, but this might be a false
+                    # positive, as there is a setting to use a different font for non-ascii
+                    # characters and this setting does not get updated when flipping the switch.
+                    # So it might be that the user checked the mark to use a different font,
+                    # selected one, but then decided to uncheck the "use different font" mark
+                    # again.
                     local nonAsciiFont=$(/usr/libexec/PlistBuddy -c "Print :New\ Bookmarks:${idx}:Non\ Ascii\ Font:" ~/Library/Preferences/com.googlecode.iterm2.plist)
                     if [[ "$term_font" -ne "$nonAsciiFont" ]]; then
                         term_font="$term_font (normal) / $nonAsciiFont (non-ascii)"
--- a/neofetch
+++ b/neofetch
@@ -1650,24 +1650,29 @@ get_term_font() {
             # thou, but that does not match to a guid in the plist.
             # So, be warned! Collisions may occur!
             # See: https://groups.google.com/forum/#!topic/iterm2-discuss/0tO3xZ4Zlwg
-            local currentProfileName=$(osascript -e 'tell application "iTerm2" to profile name of current session of current window')
+            local currentProfileName
+            currentProfileName=$(osascript -e 'tell application "iTerm2" to profile name of current session of current window')
 
             # Warning: Dynamic profiles are not taken into account here!
             # https://www.iterm2.com/documentation-dynamic-profiles.html
 
             # Count Guids in "New Bookmarks"; they should be unique
-            local profilesCount=$(/usr/libexec/PlistBuddy -c "Print :New\ Bookmarks:" ~/Library/Preferences/com.googlecode.iterm2.plist 2>/dev/null | grep "Guid" | wc -l)
-            for idx in $(seq 0 ${profilesCount}); do
-                local profileName=$(/usr/libexec/PlistBuddy -c "Print :New\ Bookmarks:${idx}:Name:" ~/Library/Preferences/com.googlecode.iterm2.plist 2>/dev/null)
+            local profilesCount
+            profilesCount=$(/usr/libexec/PlistBuddy -c "Print :New\ Bookmarks:" ~/Library/Preferences/com.googlecode.iterm2.plist 2>/dev/null | grep -c "Guid")
+            for idx in $(seq 0 "${profilesCount}"); do
+                local profileName
+                profileName=$(/usr/libexec/PlistBuddy -c "Print :New\ Bookmarks:${idx}:Name:" ~/Library/Preferences/com.googlecode.iterm2.plist 2>/dev/null)
                 if [[ "${profileName}" == "${currentProfileName}" ]]; then
                     # "Normal Font"
                     term_font=$(/usr/libexec/PlistBuddy -c "Print :New\ Bookmarks:${idx}:Normal\ Font:" ~/Library/Preferences/com.googlecode.iterm2.plist)
                     # Font for non-ascii characters
                     # Only check for a different non-ascii font, if the user checked
                     # the "use a different font for non-ascii text" switch.
-                    local useDifferentFont=$(/usr/libexec/PlistBuddy -c "Print :New\ Bookmarks:${idx}:Use\ Non-ASCII\ Font:" ~/Library/Preferences/com.googlecode.iterm2.plist)
+                    local useDifferentFont
+                    useDifferentFont=$(/usr/libexec/PlistBuddy -c "Print :New\ Bookmarks:${idx}:Use\ Non-ASCII\ Font:" ~/Library/Preferences/com.googlecode.iterm2.plist)
                     if [[ "$useDifferentFont" == "true" ]]; then
-                        local nonAsciiFont=$(/usr/libexec/PlistBuddy -c "Print :New\ Bookmarks:${idx}:Non\ Ascii\ Font:" ~/Library/Preferences/com.googlecode.iterm2.plist)
+                        local nonAsciiFont
+                        nonAsciiFont=$(/usr/libexec/PlistBuddy -c "Print :New\ Bookmarks:${idx}:Non\ Ascii\ Font:" ~/Library/Preferences/com.googlecode.iterm2.plist)
                         if [[ "$term_font" != "$nonAsciiFont" ]]; then
                             term_font="$term_font (normal) / $nonAsciiFont (non-ascii)"
                         fi
--- a/neofetch
+++ b/neofetch
@@ -1676,9 +1676,9 @@ get_term_font() {
                     if [[ "$use_different_font" == "true" ]]; then
                         local non_ascii_font
                         non_ascii_font="$(/usr/libexec/PlistBuddy -c "Print :New\ Bookmarks:${i}:Non\ Ascii\ Font:" ~/Library/Preferences/com.googlecode.iterm2.plist)"
+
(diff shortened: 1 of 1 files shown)


dritter:
With a different text and non-ascii font on a random profile (not default):
bildschirmfoto 2017-03-02 um 19 04 53

Same font for text and non-ascii on a random profile (not default):
bildschirmfoto 2017-03-02 um 19 05 30

jorgegonzalez:
This isn't working for me.

dritter:
@jorgegonzalez Do you have XCode installed? What does `which -a PlistBuddy` and `ls -l /usr/libexec/PlistBuddy` give you?
And on which OSX version are you?

jorgegonzalez:
Yes I do.

`PlistBuddy not found`

`-rwxr-xr-x  1 root  wheel  41104 Sep 13 20:57 /usr/libexec/PlistBuddy`

jorgegonzalez:
Sierra 10.12.3

dritter:
Could you post the output of 
`osascript -e 'tell application "iTerm2" to profile name of current session of current window'`
 and 
`/usr/libexec/PlistBuddy -c "Print :New\ Bookmarks:" ~/Library/Preferences/com.googlecode.iterm2.plist | grep "Guid" | wc -l`

jorgegonzalez:
Ah okay I got it working. It does not work when the profile name is an empty string (which I use to hide all text from the titlebar). And it works with more than one profile, so this fixes the issue that my PR had.

However I think you should at least consider using part of my solution in order to recreate the font names, simply because seeing `Source Code Pro` is much nicer than `SourceCodePro+Powerline+AwesomeRegular 13`, for example.

jorgegonzalez:
For example I just added these lines to your code and it improved the format of most of the fonts I tried it with:

```bash
term_font="${term_font//-/ }"
term_font="${term_font/Mono/ Mono}"
term_font="${term_font/medium/}"
term_font="${term_font/Medium/}"
term_font="${term_font/CodePro/ Code Pro}"
term_font="${term_font/ForPowerline/}"
```

jorgegonzalez:
No regex, just string replacements (if there is a better way to do this, do suggest it).

dritter:
Wow. I didn't know that it is possible to create a profile with an empty name..

Regarding your change requests (sorry):
  1. As said, I think recreating the fonts original name with string manipulation is quite error prone.
  2. A better approach to get the right name would be to ask "Font Book.app" for the correct name. E.g. it is possible to get all font names: `osascript -e 'tell application "Font Book" to name of font families'`. It might be possible to get the name for one font..
  3. I do not want to steal your credits ;)
  4. The current font names are sufficient for me.

But you could create a PR with your changes on top of this one. ;)

jorgegonzalez:
Yes there's a hacky way to remove all text in iTerm2's titlebar that involves changing your profile name to an empty string, which I think looks very nice (see my screenshot in #668).

As I said this PR fixes the issue with the profiles that mine didn't solve so I'm closing it. 

I will look into fixing the formatting via FontBook.app after this gets merged.

neofetch
```diff
@@ -1644,6 +1644,43 @@ get_term_font() {
...
         ;;
 
+        "iTerm2")
+            # Unfortunately the profile name is not unique, but it seems to be the only thing
+            # that identifies an active profile. There is the "id of current session of current window"
+            # thou, but that does not match to a guid in the plist.
+            # So, be warned! Collisions may occur!
+            # See: https://groups.google.com/forum/#!topic/iterm2-discuss/0tO3xZ4Zlwg
+            local currentProfileName
+            currentProfileName=$(osascript -e 'tell application "iTerm2" to profile name of current session of current window')
```
dylanaraps: Quote the command substitution.

`"$()"`

dylanaraps: All variables and functions in neofetch follow this naming scheme `example_var_name`. Change the variable name to reflect this. `current_profile_name`

neofetch
```diff
@@ -1644,6 +1644,43 @@ get_term_font() {
...
+
+            # Warning: Dynamic profiles are not taken into account here!
+            # https://www.iterm2.com/documentation-dynamic-profiles.html
+
+            # Count Guids in "New Bookmarks"; they should be unique
+            local profilesCount
+            profilesCount=$(/usr/libexec/PlistBuddy -c "Print :New\ Bookmarks:" ~/Library/Preferences/com.googlecode.iterm2.plist 2>/dev/null | grep -c "Guid")
```
dylanaraps: - `"$()"`
- `profiles_count`

neofetch
```diff
@@ -1644,6 +1644,43 @@ get_term_font() {
...
+            for idx in $(seq 0 "${profilesCount}"); do
```
dylanaraps: We aren't using `seq` anywhere else in the script. Instead you can use a C style for loop which removes the call to an external process.

```sh
for ((i=0; i<=profiles_count; i++)); do
```

neofetch
```diff
@@ -1644,6 +1644,43 @@ get_term_font() {
...
+                local profileName
+                profileName=$(/usr/libexec/PlistBuddy -c "Print :New\ Bookmarks:${idx}:Name:" ~/Library/Preferences/com.googlecode.iterm2.plist 2>/dev/null)
```
dylanaraps: - `"$()"`
- `profile_name`

neofetch
```diff
@@ -1644,6 +1644,43 @@ get_term_font() {
...
+                if [[ "${profileName}" == "${currentProfileName}" ]]; then
```
dylanaraps: You can remove the `{}` from these variables.

neofetch
```diff
@@ -1644,6 +1644,43 @@ get_term_font() {
...
+                    # "Normal Font"
+                    term_font=$(/usr/libexec/PlistBuddy -c "Print :New\ Bookmarks:${idx}:Normal\ Font:" ~/Library/Preferences/com.googlecode.iterm2.plist)
```
dylanaraps: `"$()"`

neofetch
```diff
@@ -1644,6 +1644,43 @@ get_term_font() {
...
+                    # Font for non-ascii characters
+                    # Only check for a different non-ascii font, if the user checked
+                    # the "use a different font for non-ascii text" switch.
+                    local useDifferentFont
+                    useDifferentFont=$(/usr/libexec/PlistBuddy -c "Print :New\ Bookmarks:${idx}:Use\ Non-ASCII\ Font:" ~/Library/Preferences/com.googlecode.iterm2.plist)
```
dylanaraps: - `"$()"`
- `use_different_font`

neofetch
```diff
@@ -1644,6 +1644,43 @@ get_term_font() {
...
+                    if [[ "$useDifferentFont" == "true" ]]; then
+                        local nonAsciiFont
+                        nonAsciiFont=$(/usr/libexec/PlistBuddy -c "Print :New\ Bookmarks:${idx}:Non\ Ascii\ Font:" ~/Library/Preferences/com.googlecode.iterm2.plist)
```
dylanaraps: - `"$()"`
- `non_ascii_font`

neofetch
```diff
@@ -1644,6 +1644,43 @@ get_term_font() {
...
+                        if [[ "$term_font" != "$nonAsciiFont" ]]; then
```
```sh
[[ "$term_font" != "$non_ascii_font" ]] && \
    term_font="$term_font (normal) / $non_ascii_font (non-ascii)"
```

dritter:
@dylanaraps Thanks for the hints. All done.

neofetch
```diff
@@ -1644,6 +1644,42 @@ get_term_font() {
...
+                    # "Normal Font"
+                    term_font="$(/usr/libexec/PlistBuddy -c "Print :New\ Bookmarks:${i}:Normal\ Font:" ~/Library/Preferences/com.googlecode.iterm2.plist)"
+                    # Font for non-ascii characters
+                    # Only check for a different non-ascii font, if the user checked
+                    # the "use a different font for non-ascii text" switch.
+                    local use_different_font
+                    use_different_font="$(/usr/libexec/PlistBuddy -c "Print :New\ Bookmarks:${i}:Use\ Non-ASCII\ Font:" ~/Library/Preferences/com.googlecode.iterm2.plist)"
+                    if [[ "$use_different_font" == "true" ]]; then
+                        local non_ascii_font
+                        non_ascii_font="$(/usr/libexec/PlistBuddy -c "Print :New\ Bookmarks:${i}:Non\ Ascii\ Font:" ~/Library/Preferences/com.googlecode.iterm2.plist)"
```
dylanaraps: Add a newline below this.

neofetch
```diff
@@ -1644,6 +1644,42 @@ get_term_font() {
...
+                local profile_name
+                profile_name="$(/usr/libexec/PlistBuddy -c "Print :New\ Bookmarks:${i}:Name:" ~/Library/Preferences/com.googlecode.iterm2.plist 2>/dev/null)"
+                if [[ "$profile_name" == "$current_profile_name" ]]; then
+                    # "Normal Font"
+                    term_font="$(/usr/libexec/PlistBuddy -c "Print :New\ Bookmarks:${i}:Normal\ Font:" ~/Library/Preferences/com.googlecode.iterm2.plist)"
+                    # Font for non-ascii characters
+                    # Only check for a different non-ascii font, if the user checked
+                    # the "use a different font for non-ascii text" switch.
+                    local use_different_font
+                    use_different_font="$(/usr/libexec/PlistBuddy -c "Print :New\ Bookmarks:${i}:Use\ Non-ASCII\ Font:" ~/Library/Preferences/com.googlecode.iterm2.plist)"
```
dylanaraps: Add a newline below this

neofetch
```diff
@@ -1644,6 +1644,42 @@ get_term_font() {
...
+
+            # Warning: Dynamic profiles are not taken into account here!
+            # https://www.iterm2.com/documentation-dynamic-profiles.html
+
+            # Count Guids in "New Bookmarks"; they should be unique
+            local profiles_count
+            profiles_count="$(/usr/libexec/PlistBuddy -c "Print :New\ Bookmarks:" ~/Library/Preferences/com.googlecode.iterm2.plist 2>/dev/null | grep -c "Guid")"
+            for ((i=0; i<=profiles_count; i++)); do
+                local profile_name
+                profile_name="$(/usr/libexec/PlistBuddy -c "Print :New\ Bookmarks:${i}:Name:" ~/Library/Preferences/com.googlecode.iterm2.plist 2>/dev/null)"
```
dylanaraps: Add a newline below this.

neofetch
```diff
@@ -1644,6 +1644,42 @@ get_term_font() {
...
+            # See: https://groups.google.com/forum/#!topic/iterm2-discuss/0tO3xZ4Zlwg
+            local current_profile_name
+            current_profile_name="$(osascript -e 'tell application "iTerm2" to profile name of current session of current window')"
+
+            # Warning: Dynamic profiles are not taken into account here!
+            # https://www.iterm2.com/documentation-dynamic-profiles.html
+
+            # Count Guids in "New Bookmarks"; they should be unique
+            local profiles_count
+            profiles_count="$(/usr/libexec/PlistBuddy -c "Print :New\ Bookmarks:" ~/Library/Preferences/com.googlecode.iterm2.plist 2>/dev/null | grep -c "Guid")"
```
dylanaraps: Add a newline below this.

dylanaraps:
I'll take a closer look later today but it looks good to me now. :+1:

dritter:
Great to hear!
Btw. added newlines.

neofetch
```diff
@@ -1644,6 +1644,46 @@ get_term_font() {
...
+                    # the "use a different font for non-ascii text" switch.
+                    local use_different_font
+                    use_different_font="$(/usr/libexec/PlistBuddy -c "Print :New\ Bookmarks:${i}:Use\ Non-ASCII\ Font:" ~/Library/Preferences/com.googlecode.iterm2.plist)"
+
+                    if [[ "$use_different_font" == "true" ]]; then
+                        local non_ascii_font
+                        non_ascii_font="$(/usr/libexec/PlistBuddy -c "Print :New\ Bookmarks:${i}:Non\ Ascii\ Font:" ~/Library/Preferences/com.googlecode.iterm2.plist)"
+                        [[ "$term_font" != "$non_ascii_font" ]] && \
+                            term_font="$term_font (normal) / $non_ascii_font (non-ascii)"
+
```
dylanaraps: My bad, Remove this newline and add it above `[[ "$term_font" != "$non_ascii_font" ]] && \`

dritter:
@dylanaraps Is there anything that needs to be done to get this PR merged?

dylanaraps:
Merged, thanks for working on this!

Issue #662, opened 2017-02-25
Screen resolution broken on macOS


jorgegonzalez:
## Description

Screen resolution is not displaying at all on macOS. 

The `system_profiler` method does work, although it seems to slow down the entire script. The refresh rate isn't displaying, however (and it is definitely turned on in my config file).

```
"Mac OS X")
    # if type -p screenresolution >/dev/null; then
    #     resolution="$(screenresolution get | awk '/Display/ {printf $6 "Hz, "}')"
    #     resolution="${resolution//x??@/ @ }"
    # 
    # else
        resolution="$(system_profiler SPDisplaysDataType | awk '/Resolution:/ {printf $2"x"$4" @ "$6"Hz, "}')"
    fi

    scale_factor="$(/usr/libexec/PlistBuddy -c "Print DisplayAnyUserSets:0:0:Resolution" /Library/Preferences/com.apple.windowserver.plist)"

    # If no refresh rate is empty.
    [[ "$resolution" == *"@ Hz"* ]] && \
        resolution="${resolution//@ Hz}"

    ((${scale_factor%.*} == 2)) && \
        resolution="${resolution// @/@2x @}"

    if [[ "$refresh_rate" == "off" ]]; then
        resolution="${resolution// @ [0-9][0-9]Hz}"
        resolution="${resolution// @ [0-9][0-9][0-9]Hz}"
    fi

    [[ "$resolution" == *"0Hz"* ]] && \
        resolution="${resolution// @ 0Hz}"
;;
```

dylanaraps:
We added `screenresolution` as a dependency on macOS because it's way faster than `system_profiler` and it shows refresh rates. (`system_profiler` doesn't iirc)

What's the output of this command?

```sh
screenresolution get
```

jorgegonzalez:
```
2017-02-25 17:31:46.140 screenresolution[5810:22510] starting screenresolution argv=screenresolution get
2017-02-25 17:31:46.154 screenresolution[5810:22510] Display 0: 1280x800x32@0
```

dylanaraps:
It looks to me like `screenresolution` is working, or am I wrong?

@iandrewt, can you take a look at this?

jorgegonzalez:
Yeah, I do have it installed, it's just not working with Neofetch for some reason.

dylanaraps:
Try this in the terminal:

```sh
screenresolution get | awk '/Display/ {printf $6 "Hz, "}'
```

[...]

PR #673, opened 2017-03-04
MacOS screenresolution: accepts output from stdout and stderr. Closes…

eliezio:
… #662

## Description

## Issues

#662


konimex:
Merged to master, thanks.

PR #676, opened 2017-03-21
WM: Add support for Amethyst on macOS

jorgegonzalez:
## Description

This PR adds detection for [Amethyst](https://github.com/ianyh/Amethyst).

## Issues 

Due to their incompatible behaviors, we will default to Amethyst if Spectacle is running, and then to Kwm.

screen shot 2017-03-21 at 2 02 28 pm

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -638,6 +638,7 @@ get_wm() {
             "Mac OS X")
                 wm="Quartz Compositor"
                 ps -e | grep -q '[S]pectacle' && wm='Spectacle'
+                ps -e | grep -q "[A]methyst" && wm="Amethyst"
                 ps -e | grep -q "[k]wm" && wm="Kwm"
             ;;
 
--- a/neofetch
+++ b/neofetch
@@ -636,10 +636,17 @@ get_wm() {
     else
         case "$os" in
             "Mac OS X")
-                wm="Quartz Compositor"
-                ps -e | grep -q '[S]pectacle' && wm='Spectacle'
-                ps -e | grep -q "[A]methyst" && wm="Amethyst"
-                ps -e | grep -q "[k]wm" && wm="Kwm"
+                ps_line="$(ps -e | grep -o '[S]pectacle\|[A]methyst\|[k]wm')"
+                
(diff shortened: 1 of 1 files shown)


dylanaraps:
Can we not turn this into one `ps` call?

```sh
ps -e | grep -o '[S]pectacle\|[A]methyst\|[k]wm'
```

jorgegonzalez:
The problem is that it is possible to have any combination of those three programs running at the same time. However if both Spectacle and Kwm are running, Spectacle doesn't really do anything; same with Spectacle and Amethyst, and if Amethyst and Kwm are both running, Kwm takes over.

So if grep finds a Spectacle process, `wm` is set to Spectacle. But if Amethyst is running _too_, then `wm` is set to Amethyst since it is behaving as the true window manager, not Spectacle. And again if Kwm is running, `wm` is set to Kwm since both Amethyst's and Spectacle's functionality are overtaken by Kwm. 

That is the logic that I was trying to convey by having a ps call on each line. Is there a better way to do this?

dylanaraps:
There might be a better way of doing this by using 1 ps call and then
either an if statement or case statement to filter the output.

I'll try and get this working in a couple of hours, I've got class now.

jorgegonzalez:
Just pushed a commit.

dylanaraps:
Do these work?

```sh
if [[ "$ps_line" =~ "kwm" ]]; then
    wm="Kwm"
elif [[ "$ps_line" =~ "Amethyst" ]]; then
    wm="Amethyst"
elif [[ "$ps_line" =~ "Spectacle" ]]; then
    wm="Spectacle"
else
    wm="Quartz Compositor"
fi

# OR (This would be prefered if it works)

case "$ps_line" in
    *"kwm"*) wm="Kwm" ;;
    *"Amethyst"*) wm="Amethyst" ;;
    *"Spectacle"*) wm="Spectacle" ;;
    *) wm="Quartz Compositor" ;;
esac
```

jorgegonzalez:
Yep the switch works.

dylanaraps:
Awesome, LGTM now :+1:

PR #677, opened 2017-03-22
OS: Add support for Nitrux

mstraube:
This PR adds support for Nitrux.  https://nxos.org/

The default konsole theme/settings
nitrux2

With linux colors
nitrux1

The logo they use (in startmenu for example)
start

## TODO
Maybe better colors?
I tried to find colors that work on the blue background and also with the standard colors... I ended up with `set_colors 4`


dylanaraps:
:+1:

Issue #679, opened 2017-03-26
Support locale information


hannesfrank:
The locale is an important system setting which often causes problem with applications. `neofetch` should be able to report it.

(On Ubuntu) one could use the `LANG` environment variable, or in general the output of the `locale` command:

```
$ locale
LANG=de_DE.UTF-8
LANGUAGE=
LC_CTYPE=de_DE.UTF-8
LC_NUMERIC=de_DE.UTF-8
LC_TIME=de_DE.UTF-8
LC_COLLATE="de_DE.UTF-8"
LC_MONETARY=de_DE.UTF-8
LC_MESSAGES="de_DE.UTF-8"
LC_PAPER=de_DE.UTF-8
LC_NAME=de_DE.UTF-8
LC_ADDRESS=de_DE.UTF-8
LC_TELEPHONE=de_DE.UTF-8
LC_MEASUREMENT=de_DE.UTF-8
LC_IDENTIFICATION=de_DE.UTF-8
LC_ALL=
```

The output could be then:

```
Language: de_DE.UTF-8
```

dylanaraps:
I've opened #682 which implements this.

PR #682, opened 2017-03-26
Info: Added function to display locale.

dylanaraps:
## Description

Adds a new, disabled by default info function to display the system locale. The output is just the value of `$LANG` and is set to `C` if empty.

Output: `Locale: en_US.UTF-8`

Closes: #679

### DIFF
--- a/config/config
+++ b/config/config
@@ -37,7 +37,7 @@ print_info() {
     # info "Public IP" public_ip
     # info "Users" users
     # info "Install Date" install_date
-    # info "Locale" locale
+    # info "Locale" locale # This only works on glibc systems.
 
     info line_break
     info cols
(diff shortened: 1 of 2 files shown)


konimex:
IIRC, this is only possible on systems with glibc, right?

dylanaraps:
Is `LANG` not always set? 

If not we can probably do:

```sh
locale="${LANG:-$(locale)}"
```

konimex:
IIRC, in musl systems, $LANG is not even set.

dylanaraps:
Yup, you're right. 

This is still fine to merge as it will be disabled by default. \*If\* it's enabled on a musl system the locale will just display `C` which I don't think is wrong in this case. (It's as right as we'll ever be)

dylanaraps:
@konimex, are you happy if I merge this?

konimex:
Well.. Why shouldn't I? Although maybe we should add some warning about this being glibc-exclusive.

dylanaraps:
Yeah, That's a good idea.

PR #683, opened 2017-03-28
Create Xmonad

Sdmeaney:
Functional programming love

## Description


dylanaraps:
I need some more info.

dylanaraps:
Xmonad is a Window Manager and we only include ascii art per distro/os so I won't merge this.

PR #692, opened 2017-04-05
OS: Added support for MinGW

konimex:
## Description

Self-explained.


Septenary:
Here's an example/proof of MinGW being successfully supported, if you're into having/keeping track those types of things.
http://prntscr.com/esojnb

PR #693, opened 2017-04-08
Add battery support for Bay Trail devices

mstraube:
This PR adds battery support for devices that use `axp288_fuel_gauge` for the info.
Some (not sure if all) Bay Trail Laptops/Tablets for example.

Note:
The Linux battery support for Bay Trail devices wasn't so good up to kernel 4.10. 
On my Laptop it started working with 4.11-rc1. I think..

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -1833,26 +1833,24 @@ get_disk() {
 get_battery() {
     case "$os" in
         "Linux")
-            # We use 'prin' here and exit the function early so that we can
-            # do multi battery support with a single battery per line.
-            for bat in "/sys/class/power_supply/BAT"*; do
+            # We use 'prin' here so that we can do multi battery support
+            # with a single battery per line.
+            for bat in "/sys/class/power_supply/"{BAT,axp288_fuel_gauge}*; do
                 capacity="$(< "${bat}/capacity")"
                 status="$(< "${bat}/status")"
 
-                # Fix for bash on Windows 10 which includes /proc files
-                # for battery usage despite there not being a battery
-                # installed.
-                [[ -z "$capacity" ]] && return
+                if [[ "$capacity" ]]; then
+                    battery="${capacity}% [${status}]"
 
-                battery="${capacity}% [${status}]"
+                    case "$battery_display" in
+                        "bar") battery="$(bar "$capacity" 100)" ;;
+                        "infobar") battery+=" $(bar "$capacity" 100)" ;;
+                        "barinfo") battery="$(bar "$capacity" 100)${info_color} ${battery}" ;;
+                    esac
 
-                case "$battery_display" in
-                    "bar") battery="$(bar "$capacity" 100)" ;;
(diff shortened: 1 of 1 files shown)


dylanaraps:
You should move this to outside of the for loop. This way you won't have to unset `$bat` and we won'y run into issues with the battery loop ending early.

mstraube:
Moving it below the foor loop it would never be reached because of the return in [this](https://github.com/dylanaraps/neofetch/blob/15451a5283f14238eec8548f36edc2dc870248ea/neofetch#L1845) line.

Changing the `return` to a `break` would require to write some code of the loop again.
Or am I missing something here?
```bash
        done  # end of for loop
            if [[ -z "$capacity" ]]; then
                capacity="$(< "/sys/class/power_supply/axp288_fuel_gauge/capacity")"
                status="$(< "/sys/class/power_supply/axp288_fuel_gauge/status")"
            fi
            [[ -z "$capacity" ]] && return

            battery="${capacity}% [${status}]"

            case "$battery_display" in
                "bar") battery="$(bar "$capacity" 100)" ;;
                "infobar") battery+=" $(bar "$capacity" 100)" ;;
                "barinfo") battery="$(bar "$capacity" 100)${info_color} ${battery}" ;;
            esac
        return
``` 
Moving it before the loop, does not work because then `$capacity` and `$status` are overwritten empty if `/sys/class/power_supply/BAT*` not exist.

I used `unset -v bat` to avoid printing `Battery*:` instead of `Battery:`
(`$bat` is `/sys/class/power_supply/BAT*` if there are no BAT* files.)

dylanaraps:
Does this work?:

```sh
for bat in "/sys/class/power_supply/"{BAT,axp288_fuel_gauge}*; do
```

mstraube:
No, `get_battery()` returns early at [L1845](https://github.com/dylanaraps/neofetch/blob/15451a5283f14238eec8548f36edc2dc870248ea/neofetch#L1845) because there are no `/sys/class/power_supply/BAT*` directories.

mstraube:
If I remove the early return line and add a check to only print if `$capacity` is not empty it works
```bash
[[ -n "$capacity" ]] && prin "${subtitle:+${subtitle}${bat: -1}}" "$battery"
```
but prints `Batterye:` instead of `Battery`, because of `${bat:-1}`

mstraube:
Pushed another version. I have no Windows installation to test, but as far as I can see it shouldn't break anything. Instead of returning early if `$capacity` is empty, it does just nothing. What do you think?

dylanaraps:
LGTM now. Will test on my laptop later today before merging. :+1:

dylanaraps:
LGTM, let me know if this is ok to merge. :+1:

mstraube:
Yes, can be merged. I squashed it into one commit.

dylanaraps:
Thanks :+1:

PR #703, opened 2017-04-25
Fix termite font matching

MatthewCox:
Each config parameter is allowed arbitrary whitespace before it.
It took me a long time to work out why it was always returning the /etc/xdg/termite/config value!


dylanaraps:
Thanks for the PR!

PR #705, opened 2017-04-26
Do not detect ibus* or indicator* as player

dawidd6:
## Description

This "absurdly long" command to get player is returning this on my ubuntu installation:
```sh
/usr/bin/ibus-daemon --daemonize
```

Then this after first addition (ibus) to awk
```sh
indicator-mpd
```

After this commit, player is being detected properly. In my case - spotify

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -1291,7 +1291,7 @@ get_memory() {
 
 get_song() {
     # This is absurdly long.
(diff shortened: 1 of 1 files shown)


dylanaraps:
Thanks! :+1:

PR #710, opened 2017-04-29
Term: Add font support for mate-terminal

mstraube:
I found a way to do font detection for MATE Terminal.

It requires to create a temporarily file. I decided to use the `/tmp` directory, but I'm not sure if it will be available (and writable)  on all systems?

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -1734,6 +1734,34 @@ get_term_font() {
             [[ "$profile_filename" ]] && term_font="$(awk -F '=|,' '/Font=/ {print $2 " " $3}' "$profile_filename")"
         ;;
 
+        "mate-terminal")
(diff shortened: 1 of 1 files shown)


dylanaraps:
This looks very complicated and we're using a lot of external processes. Is there not a simpler way of getting this info?

Hm

mstraube:
Yeah looks complicated, indeed. I managed to eliminate one call to awk and replace the other two with simpler greps. I'm afraid there is no way around using xprop and creating the temp file.

Let me know if it's ok, then I will squash the commits.

dylanaraps:
LGTM now :+1:

mstraube:
Ok, squashed it.

PR #711, opened 2017-05-01
Misc: Use fallback values for some functions

konimex:
## Description

Self-explanatory.

### Rationale

A lot of OS are *NIX, and as such, will use X (unless there are some arcane new method, aside from Wayland which only supports Linux and FreeBSD atm) for graphical environment.

It would be easier to maintain those functions as we don't have to add every OS to the cases.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -1173,7 +1173,16 @@ get_gpu() {
             esac
         ;;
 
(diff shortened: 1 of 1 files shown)


dylanaraps:
LGTM :+1:

konimex:
Since this is a small back-end change, I think a changelog wouldn't be needed.

PR #721, opened 2017-05-10
Disk: Use -k instead of -h / -g

konimex:
### Rationale

In some `df` systems, `-h` is unsupported. For AIX, we can use `-g` option. However, this caused problem in IRIX and HP-UX systems, should we merge them later.

The other alternative is detecting which `df` is being used, and act depends on which `df` (or the OS itself) is being used (keep the `-h` output or use the `-k` alternative and divide them manually). [The alternative is committed into my branch of `irix`](https://github.com/konimex/neofetch/commit/514b984b9126361103814af804b62115736c70a6), if either of these alternatives get accepted, I'll push it into the IRIX PR.

### Issues

- No more fancy outputs if >999G (1.1T becoming 1126G instead)
- More error prone in calculation if < 10G
- Less accurate

### TODO

- [ ] Testing
  - [x] Linux
  - [ ] *BSDs
  - [ ] Solaris

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -1789,9 +1789,9 @@ get_disk() {
     # Get "df" flags.
     case "$os" in
         "Haiku") err "Disk doesn't work on Haiku due to the non-standard 'df'"; return ;;
-        "Mac OS X") df_flags=(-P -h) ;;
-        "AIX") df_flags=(-P -g) ;;
-        *) df_flags=(-h) ;;
+        "Mac OS X") df_flags=(-P -k) ;;
+        "AIX") df_flags=(-P -k) ;;
+        *) df_flags=(-k) ;;
     esac
 
     # Create an array called 'disks' where each element is a separate line from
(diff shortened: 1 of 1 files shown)


neofetch
```diff
@@ -1810,7 +1810,7 @@ get_disk() {
         disk_info=($disk)
         disk_perc="${disk_info[4]/'%'}"
 
-        disk="${disk_info[2]/i} / ${disk_info[1]/i} (${disk_perc}%)"
+        disk="$((${disk_info[2]}/1024/1024))G / $((${disk_info[1]}/1024/1024))G (${disk_perc}%)"
```
dylanaraps: Inside `$(())` array variables can be shortened to `disk_info[2]`

dylanaraps:
If we divide manually then we're also assuming the disk size won't be over 1000GB or below 1GB. (We're hardcofing G)

konimex:
Yeah, this wouldn't be preferable to other systems that have `-h` in their options. I'm closing this one and going ahead with the other alternative, then.

PR #726, opened 2017-05-18
Add KSLinux Support

Takeya-Yuki:
## Description
Add KSLinux Support (issue #724 )
Add KSLinux Logo Preview (TEXT kurosawa)

Issue #724, opened 2017-05-17
Add Support for KSLinux


Takeya-Yuki:
## Description

I'd like to add support for KSLinux
kernel version: 4.4.2-kslinux

#### Neofetch version
neofetch 3.1.0-git

## Screenshot
![virtualbox_vm_17_05_2017_14_47_38](https://cloud.githubusercontent.com/assets/14171239/26142174/fb5eb826-3b11-11e7-8295-f957887bb211.png)

## Config file
default config
## Verbose log

Due to the init program is bash, so I can't get an error log
I have a /tool/cross/bin/env instead of  /usr/bin/env (but have /usr/bin folder)
It has a kpm package manager, use `kpm --get-selections` to get package list.

df -h has following output:
Filesys All     Used   %     Mounted
rootfs 1.8GB 1.6GB 87% /

It need 3.5GB Memory to startup the system due to / is rootfs

Takeya-Yuki:
I had added support for KSLinux on my personal master branch (Takeya-Yuki/neofetch).

dylanaraps:
Merged #727

PR #727, opened 2017-05-18
Fix KSLinux Support

Takeya-Yuki:
## Description
Fix Build for #726 and #724


Takeya-Yuki:
Add KSLinux Support
Add KSLinux Logo
Fix #726

neofetch
```diff
@@ -102,7 +102,8 @@ get_distro() {
 
             elif type -p tazpkg >/dev/null; then
                 distro="SliTaz $(< /etc/slitaz-release)"
-
+            elif type -p kpm > /dev/null; then
```
dylanaraps: Add a newline above this.

neofetch
```diff
@@ -434,6 +435,8 @@ get_packages() {
 
             type -p dpkg >/dev/null && \
                 packages="$((packages+=$(dpkg --get-selections | grep -cv deinstall$)))"
+            type -p kpm >/dev/null && \
```
dylanaraps: Add a newline above this.

neofetch
```diff
@@ -1633,8 +1636,13 @@ get_term() {
     # Check $PPID for terminal emulator.
     while [[ -z "$term" ]]; do
         parent="$(get_ppid "$parent")"
-        name="$(get_process_name "$parent")"
+        if [ "$parent" -lt 2 ]; then
```
dylanaraps: What does this do and why is it needed?

neofetch
```diff
@@ -3372,7 +3385,7 @@ get_distro_colors() {
             ascii_file="void"
         ;;
 
-        *"[Windows 10]"* | *"on Windows 10"* | "Windows 8"* | "Windows 10"* | "windows10" | "windows8" )
+        *"[Windows 10]"* | *"on Windows 10"* | "Windows 8"* | "Windows 10"*)
```
dylanaraps: Add these back.

neofetch
```diff
@@ -3067,6 +3075,11 @@ get_distro_colors() {
             ascii_file="korora"
         ;;
 
+        "KSLinux"*)
+            set_colors 4 7 1
+            ascii_file="ks"
```
dylanaraps: Name the file "kslinux"

ascii/distro/ks
```diff
@@ -0,0 +1,11 @@
+${c1} K   K U   U RRRR   ooo
```
dylanaraps: Name the file "kslinux"

neofetch
Takeya-Yuki: because shell is running as PID 1
(KSLinux has no init executable)

Takeya-Yuki:
I Had Fix All Bugs , and Modified as your notice.
KSLinux has a built-in initrd, so the PID of bash is 1
It'll get into a dead loop without PID=1 Check

dylanaraps:
Remove the parent check and instead make the loop look like this:

```sh
while [[ -z "$term" ]] || ((parent > 2)); do
```

Takeya-Yuki:
parent Check script I added has been removed

neofetch
```diff
@@ -1634,10 +1640,10 @@ get_term() {
     esac
 
     # Check $PPID for terminal emulator.
-    while [[ -z "$term" ]]; do
-        parent="$(get_ppid "$parent")"
-        name="$(get_process_name "$parent")"
+    while [[ -z "$term" ]] || ((parent > 1)); do
```
konimex: I don't think the whitespace here is necessary.

Takeya-Yuki: which whitespace?

Takeya-Yuki:
Remove the parent check and instead make the loop look like this:

while [[ -z "$term" ]] || ((parent > 2)); do
This should fix it.
@dylanaraps 
This will not work properly due to *sh will not set $term value ($term will always blank, and run into a dead loop)
((parent > 2)) test command will not run.

neofetch
konimex: [This one](https://github.com/Takeya-Yuki/neofetch/blob/4b0cf233a86e150766cf6f45d18045a3e71a29d7/neofetch#L1644).

dylanaraps:
Lets remove all of the `get_term` changes and we'll handle it in a separate PR.

Takeya-Yuki:
I Had Removed all the changes of get_term()

PR #729, opened 2017-05-24
Make: Set PREFIX inside neofetch on Make

### LINKED ISSUE #718: Target directories are assumed at runtime
## Description

`neofetch` checks for the existence of either `/etc/neofetch` or `/usr/local/etc/neofetch`, and for `/usr/share/neofetch`.  These directories might not exist.  This prevents neofetch from being installed to any prefix.

Likewise, the `Makefile` assumes that it can write to `$DESTDIR/etc`.  On Guix we install all files to a unique prefix, including `/etc/neofetch`.  This means that for our package we needed to patch the Makefile to replace `$(DESTDIR)/etc` with `$(PREFIX)/etc`, and in `neofetch` replace `/etc/neofetch` with `/the/actual/prefix/etc/neofetch`, and `/usr/share/neofetch` with `/the/actual/prefix/share/neofetch`.

It would be nice if these locations in `neofetch` were placeholders that could be substituted at configure time.  This could be achieved with a simple configure script and `m4` or `sed`.

#### Neofetch version

3.1.0

dylanaraps:
## Description

This PR removes the hardcoded config and ascii directories inside Neofetch and instead replaces them with a variable at the top of the script. (`$NEOFETCH_PREFIX`) (The variable defaults to `/usr/`)

The Makefile has also been updated to update this new variable inside Neofetch based on the PREFIX given to `make`.

Closes #718

## TODO

- [ ] Testing
- [ ] Simplify `sed` command if possible.

Can you guys take a look at this?

@konimex 
@rekado

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,11 +6,11 @@ all:
 install:
 	mkdir -p $(DESTDIR)$(PREFIX)/bin
 	mkdir -p $(DESTDIR)$(PREFIX)/share/man/man1
-	mkdir -p $(DESTDIR)$(PREFIX)/etc/neofetch
+	mkdir -p $(DESTDIR)/etc/neofetch
 	mkdir -p $(DESTDIR)$(PREFIX)/share/neofetch/ascii/distro
 	cp -p neofetch $(DESTDIR)$(PREFIX)/bin/neofetch
 	cp -p neofetch.1 $(DESTDIR)$(PREFIX)/share/man/man1/neofetch.1
-	cp -p config/config $(DESTDIR)$(PREFIX)/etc/neofetch/config
+	cp -p config/config $(DESTDIR)/etc/neofetch/config
 	cp -p ascii/distro/* $(DESTDIR)$(PREFIX)/share/neofetch/ascii/distro
 	# Set the PREFIX inside Neofetch.
 	sed -i'' -e "s|NEOFETCH_PREFIX=\"\/usr\"|NEOFETCH_PRE=\"$(PREFIX)\"|" $(DESTDIR)$(PREFIX)/bin/neofetch
@@ -19,4 +19,4 @@ uninstall:
 	rm -f $(DESTDIR)$(PREFIX)/bin/neofetch
 	rm -f $(DESTDIR)$(PREFIX)/share/man/man1/neofetch.1
 	rm -f -r $(DESTDIR)$(PREFIX)/share/neofetch
-	rm -f -r $(DESTDIR)$(PREFIX)/etc/neofetch
+	rm -f -r $(DESTDIR)/etc/neofetch
--- a/Makefile
+++ b/Makefile
@@ -6,15 +6,15 @@ all:
 install:
 	mkdir -p $(DESTDIR)$(PREFIX)/bin
 	mkdir -p $(DESTDIR)$(PREFIX)/share/man/man1
-	mkdir -p $(DESTDIR)/etc/neofetch
+	mkdir -p $(DESTDIR)$(PREFIX)/etc/neofetch
 	mkdir -p $(DESTDIR)$(PREFIX)/share/neofetch/ascii/distro
 	cp -p neofetch $(DESTDIR)$(PREFIX)/bin/neofetch
 	cp -p neofetch.1 $(DESTDIR)$(PREFIX)/share/man/man1/neofetch.1
-	cp -p config/config $(DESTDIR)/etc/neofetch/config
+	cp -p config/config $(DESTDIR)$(PREFIX)/etc/neofetch/config
 	cp -p ascii/distro/* $(DESTDIR)$(PREFIX)/share/neofetch/ascii/distro
 
 uninstall:
 	rm -f $(DESTDIR)$(PREFIX)/bin/neofetch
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -6,15 +6,17 @@ all:
 install:
 	mkdir -p $(DESTDIR)$(PREFIX)/bin
 	mkdir -p $(DESTDIR)$(PREFIX)/share/man/man1
-	mkdir -p $(DESTDIR)/etc/neofetch
+	mkdir -p $(DESTDIR)$(PREFIX)/etc/neofetch
```
konimex: If the default prefix is `/usr`, won't this put what supposed to be `/etc/neofetch` into `/usr/etc/neofetch` ?

dylanaraps: You're right, I totally misssed that.

How are we going to un-hard-code it then? Right now we assume that `/etc/` exists and that it's writable. I thought this was ok, but according to #718 it's an issue. 

Hm

konimex: Here's the weird thing, they used `$(PREFIX)` and not `$(DESTDIR)` to change their primary target.

`/the/actual/prefix/etc/neofetch` can be done with `make DESTDIR=/the/actual/prefix`. Most of the distros do this (Gentoo with `DESTDIR=/var/tmp/portage/neofetch-3.1.0/image` when they're testing their ebuild so they'll be put in `/var/tmp/portage/neofetch-3.1.0/image`) , so even if the PREFIX is `/usr/local`, if the DESTDIR is `/the/actual/prefix`, then the makefile will put it in `/the/actual/prefix/usr/local`.

dylanaraps: I see!

So I just have to revert the change to the `/etc` path?

konimex: Yes, but @rekado said explicitly that they need to change `$DESTDIR` to `$PREFIX`. That's what I find weird.

neofetch
```diff
@@ -3536,18 +3530,8 @@ get_full_path() {
...
-        default_config="/usr/local/etc/neofetch/config"
-
-    elif [[ -f "/data/data/com.termux/files/usr/etc/neofetch/config" ]]; then
-        default_config="/data/data/com.termux/files/usr/etc/neofetch/config"
-
-    elif [[ -f "/boot/home/config/non-packaged/etc/neofetch/config" ]]; then
-        default_config="/boot/home/config/non-packaged/etc/neofetch/config"
-
+    if [[ -f "${NEOFETCH_PREFIX}/neofetch/config" ]]; then
+        default_config="${NEOFETCH_PREFIX}/neofetch/config"
```
konimex: This will check the default config file in `/usr` or whatever `$NEOFETCH_PREFIX` should be instead of `/etc`.

dylanaraps:
I've updated the changes to also pass $DESTDIR to neofetch, this way we can do `$DESTDIR/etc/bla`.

konimex:
The changes look good, but I'm afraid we'll get complaints from package managers, so we'll need to check with them (rpm, deb, AUR, SlackBuild, etc.). I suggest we make a pre-release (or a release draft) with this branch, test this with the package manager, and see if `NEOFETCH_PREFIX` and `NEOFETCH_DESTDIR` have been set properly if the package has been installed.

konimex:
So I went ahead and create a [test tag](https://github.com/konimex/neofetch/releases/tag/3.1.0p) for use in my SlackBuild. This caused *more* problems since after installing package, `/usr/bin/neofetch` assumes `DESTDIR` is in `/tmp/SBo/package-neofetch`. When a package is being made, a temporary `DESTDIR` is made to accommodate packaging. This can cause problem when we're detecting `/etc`.

I'll test this more in the RPMs I maintain. But I'm afraid it will have the same problem.

dylanaraps:
What do you think is the best way of tackling this? We could look at how other packages do it too.

konimex:
To be honest, I don't really know.

Most OS and distros that we supported have complied with or have a compatibility towards the Unix filesystem/FHS. Even the package managers we worked with has no issues, and the issues mainly come from the niche ones with special requirements.

IMO, the best way of handling this is the current way we have.

-----

From the issue:
> On Guix we install all files to a unique prefix

I'm curious of Guix requiring prefixes as we haven't heard anything from Nix (from where Guix forked) and we have official packages there.

rekado:
Herdiansyah <notifications@github.com> writes:

> From the issue:
>> On Guix we install all files to a unique prefix
>
> I'm curious of Guix requiring prefixes as we haven't heard anything
> from Nix (from where Guix forked) and we have official packages there.

(Guix is not a fork of Nix.)

Nix patches the neofetch sources to override the embedded “/usr/share”.
We are doing the same in Guix.

--
Ricardo

GPG: BCA6 89B6 3655 3801 C3C6  2150 197A 5888 235F ACAC
https://elephly.net

rekado:
Traditionally, packages will have a configuration step during which embedded placeholders are replaced.  This works well for packages using the standard GNU build system, but also for packages using CMake or similar.

konimex:
Hmm.. It *may* be feasible, but keep in mind that we need to keep to use the GNU utilities at a minimum, considering that we support many OS that doesn't use GNU things by default. The only question remains is how.

PR #732, opened 2017-05-27
Model: Add support for Hackintosh

LER0ever:
## Description
This PR adds support for detecting if the macOS is running on a PC (Hackintosh)

Since FakeSMC.kext is a mandatory kernel extension for Hackintosh to boot, this PR checks if FakeSMC is loaded.

More about Hackintosh: [Wikipedia: OSx86](https://en.wikipedia.org/wiki/OSx86)

## Features
- Detect if the current macOS is a Hackintosh

## TODO
Maybe text rearrangement? "Model" line is a little bit too long (though still less than 80 chars)

## ScreenShot
image

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -243,7 +243,7 @@ get_model() {
         ;;
 
         "Mac OS X") 
-            if [[ $(kextstat | grep "FakeSMC") != "" ]]; then
+            if [[ "$(kextstat | grep "FakeSMC")" ]]; then
                 model="Hackintosh (SMBIOS: $(sysctl -n hw.model))"
             else
                 model="$(sysctl -n hw.model)"
--- a/neofetch
+++ b/neofetch
@@ -242,7 +242,13 @@ get_model() {
             fi
         ;;
 
-        "Mac OS X") model="$(sysctl -n hw.model)" ;;
+        "Mac OS X") 
+            if [[ $(kextstat | grep "FakeSMC") != "" ]]; then
+                model="Hackintosh (SMBIOS: $(sysctl -n hw.model))"
+            else
+                model="$(sysctl -n hw.model)"
+            fi
+        ;;
         "iPhone OS")
             case "$machine_arch" in
                 "iPad1,1") model="iPad" ;;


neofetch
```diff
@@ -242,7 +242,13 @@ get_model() {
             fi
         ;;
 
-        "Mac OS X") model="$(sysctl -n hw.model)" ;;
+        "Mac OS X") 
+            if [[ $(kextstat | grep "FakeSMC") != "" ]]; then
```
iandrewt: Change this line to `if [[ "$(kextstat | grep FakeSMC") ]]; then` and I'll merge this. We enclose all `$()` statements in quotes and the testing for `!= ""` isn't needed as the if statement will fail and go to else if the line is blank anyway.

iandrewt:
Also in regards to the line length, `MacBookProXX,Y` is the longest SMBIOS name, so if it's fitting in 80 chars then don't worry

neofetch
LER0ever: You mean `if [[ "$(kextstat | grep FakeSMC)" ]]; then`? Do I need to double quote "FakeSMC"?

iandrewt: Yep, that's the one. And yeah, double quotes (just checked, we use grep that way throughout neofetch so it's best to keep consistent)

iandrewt:
Thanks :+1:

dylanaraps:
https://github.com/dylanaraps/neofetch/pull/734

Please fix the failing build and open a new PR.

iandrewt:
The failing build was because of shellcheck, and the failing lines weren't related to the PR. I can't remember off the top of my head what the failing lines were but I checked it before I merged this.

LER0ever:
@dylanaraps Hi, sorry I did not notice the CI after pushing. I'll open a new PR after I can pass shellcheck.

The shellcheck output was:
```
In neofetch line 246:
            if [[ "$(kextstat | grep "FakeSMC")" ]]; then
                  ^-- SC2143: Use grep -q instead of comparing output with [ -n .. ].
```
But there is no `-n` in the code and I didn't quite get what it wanted me to do.
**Any advice on how I can eliminate this warning?**

konimex:
Is there any way to detect the presence of FakeSMC.kext directly (as in the FakeSMC.kext) instead of checking through kextstat? You can do
```sh
[[ -f "/path/to/FakeSMC.kext" ]]
```
and shellcheck will give you the clear.

LER0ever:
@konimex I'm afraid there isn't. When using Clover EFI for booting Hackintosh, the FakeSMC.kext is not in /System/Library/Extensions, but outside of the main boot disk, which is really hard to detect its presence.

iandrewt:
Yeah I don't think mounting the EFI partition just to check for Clover is a good idea (especially as we can't tell which disk it's on). We could check for the presence of the Clover preference pane (as a file), but this wouldn't cover all cases (different bootloaders (Ozmosis or Chameleon (pre-Yosemite))) or if they don't have the preference pane installed

konimex:
Since SC is complaining about `-n`. *Maybe* we can use `[[ -z "$(...)" ]]` instead.

LER0ever:
@iandrewt Maybe I should keep the `!=""`? The SC2143 went away after I put back the empty check.

PR #740, opened 2017-06-02
Terminal: Disable terminal detection if device is connected over SSH

### LINKED ISSUE #731: How to quit
I want to launch neofetch just at the beginning of a zsh session, jut like in the attached screenshot (the first screenshot shows screenfetch), but neofetch doesn't quit automatically so i have to press Ctrl+C… 
So if I log in on my device, it looks like that (second screenshot)…
would be great if there would be at least a launch option to automatically quit (or i didn't figure it out^^)…
screenshot_20170526_012254
screenshot_20170526_012516

konimex:
## Description

Self-explanatory. Closes #731.

### Features

Self-explanatory.

### TODO

Further testing. I've only done two tests in my own environment using a free shell service and a VM.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,7 @@
 
 - Added font support for mate-terminal. **[@mstraube](https://github.com/mstraube)**
 - [Termite] Fix font mismatch. **[@MatthewCox](https://github.com/MatthewCox)**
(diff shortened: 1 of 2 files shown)


dylanaraps:
LGTM :+1:

PR #742, opened 2017-06-03
Term: Break from loop if PPID can't be accessed/not found

konimex:
## Description

Self-explanatory.

Further testing recommended as I've only tested it in my Termux device.


chetgurevitch:
Working correctly on my device as well.

PR #754, opened 2017-06-20
Misc: Removed Makefile and added install script

konimex:
## Description

So, I've been thinking about #718 and figured out this will be my proposed (although it's less than ideal). A few alternatives will be `autoconf` or `m4` but it's overkill since we're just moving and editing small bits in shell scripts, and I figured that *maybe* a shell script would fill the hole better than Makefiles. 

### Issues

Testing will be needed to all packagers. Since most mainstream package managers worked with this script, now we need to test this with package managers like Nix and Guix.


dylanaraps:
This seems like a good alternative. :+1:

konimex:
Testing in Arch, rpm, gentoo and, SlackBuild succeed, will need testing from Termux, Homebrew, Debian, Nix, and Guix's camp, though.

Pinging @rekado, maybe this can fix the problems with prefixed systems?

dylanaraps:
The script can use standard `sh` if you convert `[[]]` to `[]` and change the shebang to `#!/usr/bin/env sh`. Here's my take on the install script:

```sh
#!/usr/bin/env sh
#
# Neofetch install script.

while [ "$1" ]; do
    case "$1" in
        "--destdir")  DESTDIR="$2" ;;
        "--bindir")   BINDIR="$2" ;;
        "--confdir")  CONFDIR="$2" ;;
        "--prefix")   PREFIX="$2" ;;
        "--asciidir") ASCIIDIR="$2" ;;
        "--mandir")   MANDIR="$2" ;;
    esac
    shift
done

if [ -z "$PREFIX" ]; then
    [ -z "$BINDIR" ] &&   BINDIR="/usr/bin"
    [ -z "$ASCIIDIR" ] && ASCIIDIR="/usr/share/neofetch/ascii/distro"
    [ -z "$MANDIR" ] &&   MANDIR="/usr/share/man/man1"
else
    [ -z "$BINDIR" ] &&   BINDIR="/bin"
    [ -z "$ASCIIDIR" ] && ASCIIDIR="/share/neofetch/ascii/distro"
    [ -z "$MANDIR" ] &&   MANDIR="/share/man/man1"
fi
[ -z "$CONFDIR" ] && CONFDIR="/etc/neofetch"

mkdir -p "${DESTDIR}${PREFIX}${BINDIR}"
mkdir -p "${DESTDIR}${PREFIX}${CONFDIR}"
mkdir -p "${DESTDIR}${PREFIX}${ASCIIDIR}"
mkdir -p "${DESTDIR}${PREFIX}${MANDIR}"

sed -i -e "s|CONFDIR|${PREFIX}${CONFDIR}|g"   neofetch
sed -i -e "s|ASCIIDIR|${PREFIX}${ASCIIDIR}|g" neofetch

cp -p neofetch       "${DESTDIR}${PREFIX}${BINDIR}"
cp -p config/config  "${DESTDIR}${PREFIX}${CONFDIR}"
cp -p ascii/distro/* "${DESTDIR}${PREFIX}${ASCIIDIR}"
cp -p neofetch.1     "${DESTDIR}${PREFIX}${MANDIR}"
```

konimex:
You may notice that I use `#!/bin/sh` instead of `#!/usr/bin/env sh`.

This is intentionally done because the installation file is meant to be portable, and every *nix (even the quasi-*nix ones like Haiku) system should have `#!/bin/sh` anyway.

konimex:
Alright, I have tested this with Nix (and in all seriousness, this package manager is very painful to work with) and confirmed that it actually works. Both config directory and ascii directory are set up correctly.

dawidd6:
I totally agree with you @konimex . I've thought that nix will be something fun, but then i've tried it and it' s maybe one of the worst managers imo. I'm glad that this issue is resolved, thanks for checking it

konimex:
After some more testing with Termux and Nix this PR is ready to merge.

dylanaraps:
Sounds good to me!

PR #758, opened 2017-06-24
Added measurement of the primary screen's diagonal in inches (Linux only)

khalilfazal:
## Description

Displays the measurement of the primary screen in inches.

## Issues
- Only works on Linux
- Only works with one screen

## TODO
- Implement on other OSes
- Test with more than one screens

### DIFF
--- a/config/config
+++ b/config/config
@@ -17,6 +17,7 @@ print_info() {
     info "Packages" packages
     info "Shell" shell
     info "Resolution" resolution
+    info "Screen Diagonal" diagonal
     info "DE" de
     info "WM" wm
     info "WM Theme" wm_theme
--- a/config/travis
+++ b/config/travis
@@ -14,6 +14,7 @@ print_info() {
     info "Packages" packages
     info "Shell" shell
     info "Resolution" resolution
+    info "Screen Diagonal" diagonal
     info "DE" de
     info "WM" wm
     info "WM Theme" wm_theme
--- a/neofetch
+++ b/neofetch
@@ -1511,7 +1511,7 @@ get_resolution() {
 
 get_diagonal() {
     case "$os" in
-        *)
+        Linux)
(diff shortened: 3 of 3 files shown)


dylanaraps:
This won't be merged into Neofetch.

FWIW you can improve the existing code like so:

```sh
# Using --nograb and --current make xrandr print REALLY fast.
# Replaced the grep and awk with a single awk command. The less pipes the better.
mm=($(xrandr --nograb --current | awk -F ')| x |mm' '/mm/ {print $3 " " $5}'))

# Using <<< you can remove the call to echo.
diagonal="$(bc <<< "scale=1; sqrt(${mm[0]}^2 + ${mm[1]}^2) / 25.4")\""
```

khalilfazal:
Would supporting other OSes increase the chances of this being merged?

dylanaraps:
No, I don't think this feature is a good fit for Neofetch. 

Also: `xrandr` seems to output the wrong `mm` size for one of my monitors resulting in an incorrect reading. (This isn't an issue with your PR)

liwakura:
Also another point:

Is the xrandr output even representative of the actual screen diameter? Im not really aware of a technical way for a computer to get its screen's physical size, so i would have assumed this value is user-supplied of some sort.

khalilfazal:
Check to see if `xdpyinfo | grep dimensions` gives a more accurate reading. For me `xdpyinfo` and `xrandr` give different values.

liwakura:
`xdpyinfo` is subject to DPI changes via `xrandr --dpi <n>` and therefore not reliable.
Just run `xrandr --dpi 20` and see it increase to like, 2 meters.

`xrandr` is not affected by that setting, but i dont know where it gets it measurements from.

konimex:
Since this is not going to be merged into upstream, I'm going to close this PR.

PR #763, opened 2017-06-30
config: Add .conf extensionn, remove shebang.

### LINKED ISSUE #762: Shebang format + config file convention for Fedora packaging
## Description

I'm packaging Neofetch for Fedora and the fedora-review package encountered a few things:
```
Checking: neofetch-3.2.0-1.fc27.noarch.rpm
          neofetch-3.2.0-1.fc27.src.rpm
neofetch.noarch: E: wrong-script-interpreter /usr/bin/neofetch /usr/bin/env bash
neofetch.noarch: W: non-conffile-in-etc /etc/neofetch/config
neofetch.noarch: E: wrong-script-interpreter /etc/neofetch/config /usr/bin/env bash
neofetch.noarch: E: non-executable-script /etc/neofetch/config 644 /usr/bin/env bash
2 packages and 0 specfiles checked; 3 errors, 1 warnings.
```

1) I changed the _/usr/bin/env bash_ in Neofetch to _/usr/bin/bash_ with a sed command in the [spec file](https://github.com/AquaL1te/neofetch/commit/d2937d93a94df82fdb5444fdffd587abbf14078a) of the RPM.
2) I also removed the _/usr/bin/env bash_ from the config file since it's [sourced](https://github.com/dylanaraps/neofetch/blob/master/neofetch#L3604) anyway.
3) Would it be possible to add the .conf extension to the config file? It won't break functionality, it's just a uniformity thing.

So in a nutshell; is the shebang in the config necessary? And could you please add the .conf extension to the config file? The most important bit is the question about the shebang in the config file because I don't want to break the functionality. The other errors generated by the fedora-review tool can either be ignored or fixed in the spec file.

dylanaraps:
## Description

Closes #762 

@konimex, you'll have to update #754 to add the `.conf` extension. 

Also I'm not sure about how we're going to handle existing people's configs, do we accept both formats?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ before_install:
     - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install screenresolution; fi
 
 script:
-    - time ./neofetch --ascii --config config/travis -v
+    - time ./neofetch --ascii --config config/travis.conf -v
     # See this wiki page for why we're disabling these errors.
     # https://github.com/dylanaraps/neofetch/wiki/Shellcheck-Exclusions
     - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then shellcheck neofetch -e SC1090,SC2009,SC2012,SC2016,SC2034,SC2128,SC2153,SC2154,SC2178,SC2010; fi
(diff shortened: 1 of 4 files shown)


konimex:
Since we're changing the config file's name, I won't merge #754 until this one gets merged, I'll merge this one to my PR after that.

About the existing config file, it's better to support both formats, IMO.

dylanaraps:
Did some testing and both the new and old config work. Merged.

PR #764, opened 2017-07-05
Rename LICENSE.md to LICENSE

rmccorm4:
Being named 'LICENSE' as opposed to 'LICENSE.md' adds the license to the top right of the repository page.

## Description


dylanaraps:
It *should* be fine as is according to GitHub's docs but you're right, the license doesn't appear in the top right of the repo info. 

> https://help.github.com/articles/licensing-a-repository/#determining-the-location-of-your-license

> Most people place their license text in a file named LICENSE.txt (or LICENSE.md) in the root of the repository; here's an example from Hubot.

If I go to the example repo from the docs it **does** work when named `LICENSE.md` so I'm a little confused. 

https://github.com/hubotio/hubot

dylanaraps:
Also, your fork of Neofetch doesn't seem to show the license info in the top right with the suffix removed. So I'm not sure if this PR actually fixes the issue or not.

https://github.com/rmccorm4/neofetch/tree/patch-1

dylanaraps:
I've figured out the issue, it's not the existence of the suffix. The issue is that the license detection fails to detect that we're using MIT since our license is *slightly* modified.

Our MIT license is missing this line:

```
The above copyright notice and this permission notice shall be included in all
copies or substantial portions of the Software.
```

dylanaraps:
I've added that offending line back to the license on one of my smaller repos and it fixes the issue. However, I'm not going to touch the license file on this repo, we removed that clause for a reason (Though it escapes me as to why).

Thanks for bringing this to my attention though, I'll fix it in my other repos as this was a side effect of copy-pasting the same license between my projects. =)

PR #765, opened 2017-07-11
Misc: Add back Makefile

konimex:
## Description

After many considerations, I decided to add back Makefile and remove the install scripts. Makefiles are more widely used, and I think this can solve the problem with our last Makefile.

So if someone wants to use `PREFIX`, they can set `make PREFIX=/usr/local SYSCONFDIR=/usr/local/etc install` (or in package managers, they can set flag `SYSCONFDIR=${PREFIX}/etc` or equivalent.

The `sed` command has been changed accordingly for non-GNU sed.

### Testing

Testing has been done in these package managers:

- pacman
- rpm
- Nix
- portage
- Termux

~~Note: Should this be merged, I'll squash it.~~


dylanaraps:
Fine by me. :+1:

Issue #766, opened 2017-07-12
Show GPU driver in use.


GrayJack:
## Description

Some info are kinda wrong. The WM Theme is showing "Kwin.aurorae", although I'm using a aurorae theme I'm using a specific one over a lot of the aurorae themes that exist, in this case is the theme Breezemite Dark. Another issue is the Theme, it's showing "Breeze" while it should show "Breath Dark", I know, Breath and Breath Dark are variants of Breeze, but yet, it would be cool if showed the right name.

- [x] KDE wrong Theme name
- [x] WM Theme wrong 

I also have a request: to show the GTK themes even on Qt DE, a lot friends ask me about the theme I use for the compatibility layer for GTK on KDE Plasma. Another fun think to show would be the used driver for the GPU.

- [x] Show GTK Themes info on KDE
- [ ] Show driver used for GPU

## Screenshot
![screenshot_20170712_042202](https://user-images.githubusercontent.com/17973728/28106668-1d6c0264-66bb-11e7-921b-c069114eed40.png)

mstraube:
Can you try the [kde branch](https://github.com/mstraube/neofetch/tree/kde) of my fork and report if it solves the theme issue?

GrayJack:
Tested the Theme fix, it worked!!

mstraube:
Thanks for testing GrayJack. 

Is the Breath Dark theme installed by default on Manjaro? I'm on Arch and installed plasma5-themes-breath from AUR, but there is no Breath Dark, just the normal one.

So I cannot set the window decoration to Breath Dark to look at the WM-Theme issue.
Can you post your `~/.config/kwinrc`?

GrayJack:
There only one Breath desktop theme, but seems that screenfetch get it from the colour scheme where there ir a Breath and a Breath Dark

mstraube:
Ah, ok. `WM Theme` is showing the selected window decoration, so if there is no Breath Dark decoration... 

Regarding the `Kwin.aurorae` issue: 
- What is your selected window decoration? 
- Please post your `~/.config/kwinrc`

mstraube:
I tested some other Aurorae themes and they were all detected as `Kwin.aurorae`.

Should be fixed by: df2c85b496c4089560570ba1c204288ef2343eb9

mstraube:
Tested some other Aurorae themes and they were all detected as `Kwin.aurorae`.
[...]

PR #767, opened 2017-07-12
Fix KDE themes

mstraube:
Fixes the issues with KDE themes in #766


konimex:
Merged to master, thanks.

PR #771, opened 2017-07-13
CPU: Fix inaccurate speed for processors below 1GHz

konimex:
## Description

Self-explanatory.

`speed_shorthand` will have to be disabled for CPU speed below 1 GHz. Unless we want to nest another `if` there.


dylanaraps:
LGTM :+1:

PR #774, opened 2017-07-14
WM: Fix/add Window Maker detection

mstraube:
## Description
Currently it looks like this:

wmaker

## Notes
Added `-len 100` because otherwise that call to `xprop` is very slow.
(the property _WINDOWMAKER_ICON_TILE is very long)

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -663,10 +663,13 @@ get_wm() {
 
     if [[ -n "$DISPLAY" && "$os" != "Mac OS X" ]]; then
         id="$(xprop -root -notype | awk '$1=="_NET_SUPPORTING_WM_CHECK:"{print $5}')"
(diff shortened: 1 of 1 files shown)


dylanaraps:
LGTM :+1:

Is it worth adding `-len 100` to any of the other `xprop` commands?

mstraube:
Hm, good question. Maybe for the one above, where we are just interested in following line?
```
_NET_SUPPORTING_WM_CHECK: window id # 0x2000006
```
But there is no noticeable speed increase on my system.

For other calls to xprop: 
As long as the needed info wouldn't be cut off, limiting the property length would be ok I think. But if it's worth the risk of breaking something in favor of not really noticeable speed improvement?

As mentioned for Window Maker, `_WINDOWMAKER_ICON_TILE` is **very** long..

dylanaraps:
Fair enough, better to leave the others as is. 👍

PR #778, opened 2017-07-16
OS: Add support for Parsix

mstraube:
http://www.parsix.org/wiki/WikiStart

parsix


konimex:
I don't think we should add support of a [shutting down](http://www.parsix.org/wiki/News#ParsixGNULinuxProjectTerminationAnnouncement-May192017) distro.

Although I'm not so sure, since we did add support for a dead OS (IRIX).

mstraube:
Oh, I didn't notice that it's shutting down.

dylanaraps:
If all it requires is ascii art and colors then it's fine by me, anything more than that for a inactive distro is too much.

konimex:
Alright. Merging this one.

PR #779, opened 2017-07-18
Misc: Disable 2>/dev/null if verbose mode is on

konimex:
## Description

Self-explanatory.

### Rationale

Before this commit, the verbose output will be just like this.

```
+ export LC_ALL=C
+ LC_ALL=C
+ [[ ascii != \o\f\f ]]
+ printf %b '\033[19A\033[9999999D'
+ old_functions
+ type printinfo
+ get_cache_dir
+ case "$os" in
+ cache_dir=/tmp
+ print_info
+ dynamic_prompt
+ case "$image_backend" in
+ printf '\n'
+ (( lines < info_height ))
+ [[ -n '' ]]
+ return
+ [[ ascii == *w3m* ]]
```

`print_info`, that's it. We have no idea what's happening *inside* `print_info`, so if errors are detected because of one malfunction inside `print_info`, we won't know what happened.

After this commit, the verbose output will be like this:

```
+ export LC_ALL=C
+ LC_ALL=C
+ [[ ascii != \o\f\f ]]
+ printf %b '\033[19A\033[9999999D'
+ old_functions
+ type printinfo
+ get_cache_dir
+ case "$os" in
+ cache_dir=/tmp
+ case "$verbose" in
+ print_info
+ info title
+ [[ -n '' ]]
+ unset -v prin
+ case "$verbose" in
+ get_title
+ user=koni
(...)
+ get_cpu
+ [[ Arch Linux  x86_64 == \N\e\t\B\S\D* ]]
+ case "$os" in
+ case "$machine_arch" in
++ awk -F ': | @' '/model name|Processor|^cpu model|chip type|^cpu type/ {printf $2; exit}' /proc/cpuinfo
+ cpu='Intel(R) Core(TM) i5-4460  CPU'
+ [[ Intel(R) Core(TM) i5-4460  CPU == *\p\r\o\c\e\s\s\o\r\ \r\e\v* ]]
+ speed_dir=/sys/devices/system/cpu/cpu0/cpufreq
+ temp_dir=/sys/class/hwmon/hwmon0/temp1_input
+ [[ -d /sys/devices/system/cpu/cpu0/cpufreq ]]
./neofetch: line 859: /sys/devices/system/cpu/cpu0/cpufreq/bios_limit: No such file or directory
+ speed=
./neofetch: line 860: /sys/devices/system/cpu/cpu0/cpufreq/bios_limit: No such file or directory
+ speed=
+ speed=3400000
+ speed=3400
(...)
+ (( ++info_height ))
+ prin=1
+ [[ -n 1 ]]
+ return
+ dynamic_prompt
+ case "$image_backend" in
+ printf '\n'
+ (( lines < info_height ))
+ [[ -n '' ]]
+ return
+ [[ ascii == *w3m* ]]
```

Yes, it will expose errors like `bad substitution` or `/sys/devices/system/cpu/cpu0/cpufreq/bios_limit: No such file or directory`, but hey, that's kinda the point of verbose mode, isn't it?

## TODO

N/A

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -2746,8 +2746,10 @@ info() {
     unset -v prin
 
     # Call the function.
-    "get_${2:-$1}" 2>/dev/null
-
+    case "$verbose" in
+        "on") "get_${2:-$1}" ;;
+        *) "get_${2:-$1}" ;;
+    esac
     # If the get_func function called 'prin' directly, stop here.
     [[ "$prin" ]] && return
 
@@ -4362,9 +4364,15 @@ get_args() {
 main() {
     cache_uname
     get_os
-    get_default_config 2>/dev/null
+    case "$verbose" in
+        "on") get_default_config ;;
+        *) get_default_config 2>/dev/null ;;
+    esac
     get_args "$@"
-    get_distro 2>/dev/null
+    case "$verbose" in
+        "on") get_distro ;;
+        *) get_distro 2>/dev/null ;;
+    esac
     get_bold
     get_distro_colors
 
@@ -4380,7 +4388,10 @@ main() {
     image_backend
     old_functions
     get_cache_dir
-    print_info 2>/dev/null
+    case "$verbose" in
+        "on") print_info ;;
+        *) print_info 2>/dev/null ;;
+    esac
     dynamic_prompt
 
     # w3m-img: Draw the image a second time to fix


neofetch
```diff
@@ -2746,8 +2746,10 @@ info() {
     unset -v prin
 
     # Call the function.
-    "get_${2:-$1}" 2>/dev/null
-
+    case "$verbose" in
+        "on") "get_${2:-$1}" ;;
+        *) "get_${2:-$1}" ;;
```
dylanaraps: Missing `2>/dev/null` here.

dylanaraps:
Could we not assign `2>/dev/null` to a value? (`cmd ${devnull}`). We can then set it if `$verbose` is off.

konimex:
Apparently when I set it as a variable, when I run a test command it was set as a... file? Whatever it is, it's not interpreted as the usual argument/options, interestingly this didn't happen with other variables.

```
$ devnull="2>/dev/null"
+ devnull='2>/dev/null'
$ df ${devnull}
+ df '2>/dev/null'
df: '2>/dev/null': No such file or directory
$ flag="-h -P"
+ flag='-h -P'
$ df ${flag}
+ df -h -P
Filesystem        Size  Used Avail Use% Mounted on
dev               3,9G     0  3,9G   0% /dev
run               3,9G  1,4M  3,9G   1% /run
(...)
```

Also, [I've created another solution which removes all `2>/dev/null` instance and put it into one single function.](https://github.com/konimex/neofetch/commit/eaa03bdeb0fad00db3de888f7c4abffd9b4ee211) However, I don't know where to put it, and I don't know where to put it in the main function.

dylanaraps:
I like the second solution better but I'm not sure about why those two functions should be moved elsewhere. I wonder if the `2>/dev/null` is failing because it's considered one argument as opposed to 3. Try setting an array: `devnull=("2" ">" "/dev/null")`

konimex:
>Try setting an array: `devnull=("2" ">" "/dev/null")`

```sh
$ devnull=("2" ">" "/dev/null")
+ devnull=("2" ">" "/dev/null")
$ df ${devnull[@]}
+ df 2 '>' /dev/null
df: '>': No such file or directory
Filesystem        1K-blocks      Used Available Use% Mounted on
/dev/mapper/koni 1165664380 632095220 474333728  58% /home
dev                 4045632         0   4045632   0% /dev
```

Nope. Doesn't work.

Anyway, I'm closing this one and opening PR for the other solution.

PR #788, opened 2017-07-26
Term: Fix mate-terminal font when maximized or fullscreen

mstraube:
## Description
In maximzed windows and fullscreen mode font detection for MATE Terminal did not work.
The grep commands failed because the config file contains extra lines if the window is maximized and/or fullscreen.

## Features
- Fixes font detection when maximized or fullscreen
- Faster (nearly 2 times)

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -1814,13 +1814,12 @@ get_term_font() {
                 mate-terminal --save-config="$mateterm_config"
 
                 role="$(xprop -id "${WINDOWID}" WM_WINDOW_ROLE)"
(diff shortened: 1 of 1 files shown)


dylanaraps:
Does the `awk` command work on non GNU `awk`?

mstraube:
Yes, I've just tested it with the GhostBSD live CD.
```
ghostbsd@ghostbsd:~ % awk --version
awk version 20121220 (FreeBSD)
```

dylanaraps:
Thanks!

PR #790, opened 2017-07-28
Prioritise maim(1) over scrot(1).

ybden:
If a user has maim(1) installed as well as scrot(1), they likely wish to use the former instead of the latter, as it is less commonly installed (and is even described as "supposed to be an improved scrot").


dylanaraps:
LGTM :+1:

PR #797, opened 2017-08-02
cleanup: Enforce a line length of '100'

dylanaraps:
## Description

I want to start cleaning up neofetch. This will include various PRs from me starting with this one. This PR limits the length of every line in the script to `100` characters max. This is done through breaking commands into multiple lines and using more variable where necessary.

I was originally going to limit lines to `80` characters but this seems impossible, hell limiting the lines to `100` proved difficult in places.

Later changes I'd like to make are as follows:

- Make all variables local to their functions.
    - The top of each function will include: `local var1 var2 var3`.
    - This makes it easier to see where variables were declared.
- Remove the dynamic variable use.
- Make each function return a value.
    - Instead of setting a variable we'll print the value.
- Document what each function takes as input and what it outputs.

This needs a thorough review as I could've very easily broken something in the process of doing this. Edit: I've already found some bugs in this PR.

## Changes

This includes some smaller changes.

- Add `/usr/libexec` to `$PATH`.

## TODO

- [x] Fix the remaining 6 or so lines that are still too long.
- [x] Add checking for line length to Travis.
    - Shellcheck might have an option for this.


dylanaraps:
I can't see anymore bugs in this PR so I'm going to merge it.

Issue #753, opened 2017-06-19
How to disable colors?


raxod502:
## Description

I would like to pipe the output of `neofetch` into another command, but cannot figure out how to disable the ASCII escape codes that litter the output. I did not see any options in `man neofetch` that appeared to be relevant.

#### Neofetch version

3.1.0

## Config file

None

ladislavskufca:
You mean "color bar"?

Open your configuration file in .config and use `#` to make comment and hide `info cols`. I think that this is the bar you want to hide.

dylanaraps:
No, he means the escape codes that format the text. 

Neofetch currently doesn't have an option to do this and I'm not sure if I'll add this feature since Neofetch wasn't intended to be piped into other commands. If you'd like to use some info from Neofetch I'd instead suggest looking into the source code to find the file or system command that displays the information you'd like to pipe.

raxod502:
The most frequent thing that I use Neofetch for is filling in the "System Information" section in an issue report. So I run `neofetch` and then copy/paste the output. But I would much prefer to just do `neofetch | pbcopy`!

liwakura:
Maybe this would suffice for your usecase?

`neofetch|sed 's/\x1B\[[0-9;]*m//g'`

Using this, we would not need to modify neofetch.

raxod502:
Well, that would technically work, but it seems a little bit of an excessively hacky workaround to me. I could certainly put it in an alias, though. Thanks!

raxod502:
Actually, I take it back. Before, it looked like this:

```
[?25l[?7l[0m[32m[1m                -/+:. 
               :++++. 
              /+++/. 
      .:-::- .+/:-``.::- 
   .:/++++++/::::/++++++/:` 
[0m[33m[1m .:///////////////////////:` 
 ////////////////////////` 
[0m[31m[1m-+++++++++++++++++++++++` 
/++++++++++++++++++++++/ 
[0m[31m[1m/sssssssssssssssssssssss. 
:ssssssssssssssssssssssss- 
[0m[35m[1m osssssssssssssssssssssssso/` 
 `syyyyyyyyyyyyyyyyyyyyyyyy+` 
[0m[34m[1m  `ossssssssssssssssssssss/ 
    :ooooooooooooooooooo+. 
     `:+oo+/:-..-:/+o+/- 
[16A[9999999D[32C[0m[1m[32m[1mraxod502[0m@[32m[1mRadon-MBP-New.local[0m 
[...]

PR #799, opened 2017-08-03
Misc: Reimplement stdout mode

konimex:
## Description

Seeing that there are users that uses Neofetch to pipe it for filling their "System Information" section in an issue report (#753), I think we can reimplement stdout mode by stripping all escape codes. So no actual code will be changed.

P.S.: I'm not sure why would we would have a trailing space in `printf` in `get_underline` and `prin` function, so for the time being I removed it.

## Issues

N/A


dylanaraps:
Nice!

LGTM :+1:

PR #802, opened 2017-08-11
Fix Terminal Font displaying incorrect information in urxvt, if one of the fonts has 'font' in its name

Vaernil:
## Description
Very specific case, but for example having Unifont (xft:Unifont:style=Medium:antialias=false") as your terminal font, makes Neofetch print it as "Terminal Font: style=Medium:antialias=false" instead of "Terminal Font: Unifont".
This is due to the fact, that it looks for "font:" string. Fixed it by making it look for "*font:" instead, but I'm not sure if that's the right approach.
Pictures say more than words :smiley: 
Before and after


dylanaraps:
LGTM 👍 

Thanks!

PR #810, opened 2017-08-17
Term: Fix mate-terminal font

mstraube:
It does not work with the linebreak and still fits in line length 100 without it.


dylanaraps:
LGTM :+1:

Thanks!

PR #811, opened 2017-08-20
Make Makefile more portable

AMDmi3:
## Description

- Make PREFIX, SYSCONFDIR overridable
- Make MANDIR customizable (for instance, FreeBSD uses $(PREFIX)/man instead of $(PREFIX)/share/man)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,6 @@
-PREFIX = /usr
-SYSCONFDIR = /etc
+PREFIX ?= /usr
+SYSCONFDIR ?= /etc
+MANDIR ?= $(PREFIX)/share/man
 
 all:
 	@echo Run \'make install\' to install Neofetch.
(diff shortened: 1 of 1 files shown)


konimex:
Isn't `?=` a GNU extension so it would make the Makefile *less* portable?

Also, IIRC, the prefix is already overrideable with the usual `make PREFIX=/usr/local install`.

AMDmi3:
> Isn't ?= a GNU extension so it would make the Makefile less portable?

Definitely not GNU extension. At least all BSD make variants I've seen support it just fine.

> Also, IIRC, the prefix is already overrideable with the usual make PREFIX=/usr/local install.

Yes, but the point is making it overridable through environment, which is the most common way of passing such stuff.

dylanaraps:
Thanks!

PR #815, opened 2017-08-25
fixed get_battery for device names starting with CMB

deadda7a:
## Description

On some devices I tested (Fujitsu Lifebook E754, ThinkPad T410s, ThinkPad T400), the battery gets assigned the device path /sys/class/power_supply/CMB1 when using Fedora 26 with Kernel 4.12.5-300.fc26.x86_64. I adapted the get_battery function for this case.

## Features
* Fix get_battery for Fedora 26


dylanaraps:
LGTM :+1:

Thanks!

PR #817, opened 2017-08-27
Fixed scrot_cmd not accepting options.

winneon:
## Description

Currently if you want to supply additional options to the `scrot_cmd` option in the config file or on the command line, neofetch does not use the command. For example, change `scrot_cmd` in the config or on the command line to `scrot -u`. This should make scrot only take a screenshot of the currently focused window. However, it does not do that. This is because neofetch is testing the whole command, including the option, against currently installed programs.

Now, with this patch, additional options now are accepted because neofetch is only testing the command call itself, not including the options.

### Current neofetch behavior

`test -p 'scrot -u'` -> `false`

### Fixed neofetch behavior

`test -p 'scrot'` -> `true`

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -2858,8 +2858,8 @@ scrot_program() {
     # falling back to OS specific screenshot tools.
     if [[ -n "$DISPLAY" ]]; then
         if [[ "$scrot_cmd" != "auto" ]] && type -p $(awk '{print $1;}' <<< $scrot_cmd) >/dev/null; then
-          scrot_program=$(awk '{print $1;}' <<< $scrot_cmd)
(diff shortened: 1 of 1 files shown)


dylanaraps:
Thanks for the PR and for reporting the issue, fixed in https://github.com/dylanaraps/neofetch/commit/bad43ac17ebe31c1e497fdcc8735db5c008b5bcb.

winneon:
Looks like you're much better at bash scripting than I am. Bash isn't my first language. Thank you for the fast and better fix!

dylanaraps:
No problem :)

PR #818, opened 2017-08-28
Fixed neofetch not finding urxvt fonts with the `.` declaration.

winneon:
## Description

Currently if you declare your urxvt fonts with `.` instead of `*` in your `.Xresources`, neofetch will not find your terminal font. This patch fixes that.

I'd imagine there's a much simpler way of doing this (without calling the same 3 lines twice with minor differences), but I'm not an expert at bash. Sorry about that.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -1976,6 +1976,12 @@ get_term_font() {
             term_font="${term_font/*"*font:"}"
             term_font="$(trim "$term_font")"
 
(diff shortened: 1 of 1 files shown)


dylanaraps:
Fixed in https://github.com/dylanaraps/neofetch/commit/076e4bd189caacfb335fd4d5c8d1f8b77e96fade.

Thanks!

PR #824, opened 2017-09-10
OS: add support for Proxmox VE

ThomasLamprecht:
## Description

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -84,6 +84,12 @@ get_distro() {
                     *) distro="Siduction ($(lsb_release -sic))"
                 esac
 
+            elif [[ -x "/usr/bin/pveversion" ]]; then
+                case "$distro_shorthand" in
+                    "on" | "tiny") distro="Proxmox VE" ;;
+                    *) distro="Proxmox VE $(pveversion | grep -oP 'pve-manager\/\K\d+\.\d+')"
+                esac
+
             elif type -p lsb_release >/dev/null; then
                 case "$distro_shorthand" in
                     "on")   lsb_flags="-sir" ;;
@@ -3532,6 +3538,11 @@ get_distro_colors() {
         "Porteus"*)
             set_colors 6 7
             ascii_file="porteus"
+        ;;
+
+        "Proxmox VE"*)
+            set_colors 7 202
+            ascii_file="proxmox_ve"
         ;;
 
         "Puppy"* | "Quirky Werewolf"* | "Precise Puppy"*)
--- a/ascii/distro/proxmox_ve
+++ b/ascii/distro/proxmox_ve
@@ -1,0 +1,20 @@
+${c1}         .://:`              `://:.
+       `hMMMMMMd/          /dMMMMMMh`
+        `sMMMMMMMd:      :mMMMMMMMs`
+${c2}`-/+oo+/:${c1}`.yMMMMMMMh-  -hMMMMMMMy.`${c2}:/+oo+/-`
+`:oooooooo/${c1}`-hMMMMMMMyyMMMMMMMh-`${c2}/oooooooo:`
+  `/oooooooo:${c1}`:mMMMMMMMMMMMMm:`${c2}:oooooooo/`
+    ./ooooooo+-${c1} +NMMMMMMMMN+ ${c2}-+ooooooo/.
+      .+ooooooo+-${c1}`oNMMMMNo`${c2}-+ooooooo+.
+        -+ooooooo/.${c1}`sMMs`${c2}./ooooooo+-
+          :oooooooo/${c1}`..`${c2}/oooooooo:
+          :oooooooo/`${c1}..${c2}`/oooooooo:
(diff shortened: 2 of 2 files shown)


dylanaraps:
Thanks for the PR!

- Lines must be kept below `100` chars long.
- Is the extra detection needed?
    - Does Neofetch not detect Proxmox correctly without the extra detection stuff?
- Remove the 1 space gap between the art and the variables.
- You have too many color variables, the art can be simplified.
    - Color formatting follows line breaks.

Simplified art:

```sh

${c1}         .://:`              `://:.
       `hMMMMMMd/          /dMMMMMMh`
        `sMMMMMMMd:      :mMMMMMMMs`
`-/+oo+/:`.yMMMMMMMh-  -hMMMMMMMy.`${c2}:/+oo+/-`
`:oooooooo/${c1}`-hMMMMMMMyyMMMMMMMh-`${c2}/oooooooo:`
  `/oooooooo:${c1}`:mMMMMMMMMMMMMm:`${c2}:oooooooo/`
    ./ooooooo+-${c1} +NMMMMMMMMN+ ${c2}-+ooooooo/.
      .+ooooooo+-${c1}`oNMMMMNo`${c2}-+ooooooo+.
        -+ooooooo/.${c1}`sMMs`${c2}./ooooooo+-
          :oooooooo/${c1}`..`${c2}/oooooooo:
          :oooooooo/`${c1}..${c2}`/oooooooo:
        -+ooooooo/.`${c1}sMMs${c2}`./ooooooo+-
      .+ooooooo+-`${c1}oNMMMMNo${c2}`-+ooooooo+.
    ./ooooooo+-${c1} +NMMMMMMMMN+ ${c2}-+ooooooo/.
  `/oooooooo:`${c1}:mMMMMMMMMMMMMm:${c2}`:oooooooo/`
`:oooooooo/`${c1}-hMMMMMMMyyMMMMMMMh-${c2}`/oooooooo:`
`-/+oo+/:`${c1}.yMMMMMMMh-  -hMMMMMMMy.${c2}`:/+oo+/-`
${c1}        `sMMMMMMMm:      :dMMMMMMMs`
       `hMMMMMMd/          /dMMMMMMh`
         `://:`              `://:`
```

GamerSource:
> Does Neofetch not detect Proxmox correctly without the extra detection stuff?

No, this is needed. Proxmox VE ships Debians original base-files which provides Debians upstream /etc/os_release.

> You have too many color variables, the art can be simplified.
>   Color formatting follows line breaks.

I saw that but kept it in sync with the screenfetch one, was less work.
But, yes yours is better, I'll adapt this and fix the line length.

Thanks for the feedback!

GamerSource:
Hi @dylanaraps , just FYI I force pushed changes to address you comments.
I.e. keep lines shorter and use your ASCII art adaption which switches only colors when really necessary.
There was a small visual bug in it (one color missed) - I'm noting this just so if one stumbles over this and wants to use it for whichever reason he is aware of this.
Nonetheless much thanks for your feedback, could you take another look to see if now everything is up to neofetchs standards, if you get some spare time?

dylanaraps:
I’ll take a look when I get home from uni. Thanks for making the changes :)

GamerSource:
Hi @dylanaraps  any chance you could look at the current version? WOuld be really great, thanks :)

neofetch
```diff
@@ -84,6 +84,12 @@ get_distro() {
                     *) distro="Siduction ($(lsb_release -sic))"
                 esac
 
+            elif [[ -x "/usr/bin/pveversion" ]]; then
```
dylanaraps: Change this to `elif type -p pveversion >/dev/null; then`. It's safer to not hardcode the path.

neofetch
```diff
@@ -84,6 +84,12 @@ get_distro() {
...
+                case "$distro_shorthand" in
+                    "on" | "tiny") distro="Proxmox VE" ;;
+                    *) distro="Proxmox VE $(pveversion | grep -oP 'pve-manager\/\K\d+\.\d+')"
```
dylanaraps: What's the output of `pveversion` on its own? Grep may not be needed if we can use bash builtins.

neofetch
wget: :+1:

neofetch
```diff
@@ -3534,6 +3540,11 @@ get_distro_colors() {
             ascii_file="porteus"
         ;;
 
+        "Proxmox VE"*)
+            set_colors 7 202
```
wget: Are we authorized to use 24 bits based colors here? The color chosen here is not really the same orange color as Proxmox.
https://en.wikipedia.org/wiki/ANSI_escape_code#24-bit
If not, maybe we could add a note to the README or the wiki in order to help newscomers?

wget:
I made a new pull request including the aforementioned requested changes: https://github.com/dylanaraps/neofetch/pull/1054

The null pull request has been rebased on the latest master.

cc @GamerSource @dylanaraps

Issue #821, opened 2017-09-05
Font looks odd on konsole


3k2:
## Description

is the font section suppose to look like that on KDE/plasma when using konsole?

## Screenshot

https://x0.at/TGp.png

Font: Update_info=RobotoMono Nerd Font,9,-1,5,50,0,0,0,0,0,Regular 

## Config file

https://ptpb.pw/K-Jl

## Verbose log

https://ghostbin.com/paste/4jxj8

1. Run `neofetch -vv 2> neofetchlog`
2. Upload the contents of `neofetchlog` to pastebin, gist or equivalent.

mstraube:
Not limitied to Konsole. Output in konsole and sakura on my System (using KDE/Plasma):
```
Font: Noto Sans,10,-1,5,50,0,0,0,0,0,Regular [KDE], Noto Sans Regular 10 [GTK2/3]
```
I have a one-line fix, but it does not print the font size of the KDE font.
I'm working on it..

mstraube:
Should be fixed with 8d8b2da8903a5c16172bcce0992ff070a15af16a

@3k2 
Can you try the kde branch of my fork?

@dylanaraps 
Maybe there is a simpler way  to get the info?

e.g.
`Noto Sans,10,-1,5,50,0,0,0,0,0,Regular` -> `Noto Sans Regular 10`

PR #828, opened 2017-09-17
Font: Fix KDE font, #821

mstraube:
Fix displayed font when using KDE. See #821

before
`Font: Noto Sans,10,-1,5,50,0,0,0,0,0,Regular [KDE], Noto Sans Regular 10 [GTK2/3]`

after
`Font: Noto Sans Regular 10 [KDE], Noto Sans Regular 10 [GTK2/3]`


dylanaraps:
LGTM :+1:

PR #833, opened 2017-09-22
CoreOS identification fix

chrisweeksnz:
## Description

CoreOS has a VERSION_ID identified in os-release that contains an epoch value since the coreos project started and hence will change every upgrade. This breaks the catchall os-release test.

Apologies for not catching this in the initial submission.

## Issues

Behaviour pre-patch:

```
$ distro="${NAME:-${DISTRIB_ID}} ${VERSION_ID:-${DISTRIB_RELEASE}}"; echo $distro
Container Linux by CoreOS 1409.1.0
```

Behaviour post-patch:

```
$ [[ $NAME == "Container Linux by CoreOS" ]] && distro="$ID"; echo $distro
coreos
```

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -134,7 +134,12 @@ get_distro() {
                     "tiny") distro="${NAME:-${DISTRIB_ID:-${TAILS_PRODUCT_NAME}}}" ;;
                     "off") distro="${PRETTY_NAME:-${DISTRIB_DESCRIPTION}} ${UBUNTU_CODENAME}" ;;
                 esac
-
+		
+                # Workaround for CoreOS - CoreOS has a VERSION_ID that is epoch since the project
+		# started and hence will change every upgrade. This breaks the catchall os-release
+		# test.	
+		[[ $NAME == "Container Linux by CoreOS" ]] && distro="$ID"
+		
                 # Workarounds for distros that go against the os-release standard.
                 [[ -z "${distro// }" ]] && distro="$(awk '/BLAG/ {print $1; exit}')" "${files[@]}"
                 [[ -z "${distro// }" ]] && distro="$(awk -F'=' '{print $2; exit}')"  "${files[@]}"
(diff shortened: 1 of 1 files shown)


konimex:
Sorry for replying this way too late, but I don't see how this breaks the catch-all release detection just because `VERSION_ID` contains an epoch value that will change every upgrade.

chrisweeksnz:
I think the existing check matches ```distro="${NAME:-${DISTRIB_ID}} ${VERSION_ID:-${DISTRIB_RELEASE}}"```from line 133 against the string ```Container Linux by CoreOS``` on line 3205, which results in ```Container Linux by CoreOS 1409.1.0``` not matching ```Container Linux by CoreOS``` because of the presence of the epoch timestamp, in turn causing the ascii art to not be displayed. 

Rather than adding the epoch timestamp to line 3205 (which would break every time CoreOS produced a new release), the proposed change matches on $NAME and uses $ID to be concise and have something that shouldn't change at each release.

Please let me know if that doesn't make sense and thanks for the reply!  :)

konimex:
I don't think it needs to be changed.

The case statement in line 3205 goes like this: `"Container Linux by CoreOS"*)`.

That means whenever `$distro` detects `Container Linux by CoreOS 1409.1.0`, when `get_ascii` does its job, it will check `Container Linux by CoreOS` *and* whatever comes after it (e.g. `Container Linux by CoreOS 1409.1.0` and `Container Linux by CoreOS 1408.1.0` will work just fine). It doesn't have to be exactly the same character.

This is what I got when I tested it with a temporary `/etc/os-release` for two different versions of CoreOS.

2017-10-13-203504_1920x1080_scrot

chrisweeksnz:
Cool - there must be something incorrect in my test instance - existing result was giving me this:

screen shot 2017-10-14 at 10 28 29 am

Given your test, please discard the PR.

PR #834, opened 2017-09-25
tiny fix for WM always being detected as Kwm on macOS

tokamach:
the grep command which included `[c]hunkwm` in it's search parameters was triggering the `*"kwm"*)` glob in the switch statement, since the **kwm** substring of [c]hun**kwm** fits the glob pattern, if that makes any sense....

This was the most obvious and simplest fix I could come up with

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -706,7 +706,7 @@ get_wm() {
     else
         case "$os" in
             "Mac OS X")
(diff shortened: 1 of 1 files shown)


dylanaraps:
LGTM :+1:

Thanks!

PR #840, opened 2017-10-13
Fix Install Date for macOS

iandrewt:
## Description

The Install Date function used to print a seemingly random date, I think it might have been the time last booted? Or at least it was when `/var/log/install.log` was last modified. I've fixed it so now it's based off of the creation date of that file. I'm not sure if other OSes that error on `ls --version` support the method I used (just adding `-u` to the ls command for macOS only), and its probably file system specific, so I made the change only work on macOS.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -2256,7 +2256,7 @@ get_install_date() {
 
         *)
             if [[ "$os" == "Mac OS X" ]]; then
-                install_date="$(ls -dlctuT "$install_file" | awk '{printf $9 " " $6 " " $7 " " $8}')"
+                install_date="$(ls -dlctUT "$install_file" | awk '{printf $9 " " $6 " " $7 " " $8}')"
             else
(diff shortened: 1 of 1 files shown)


konimex:
Since `-U` in `ls` is MacOS-specific, we can still use `ls --version` to detect MacOS output (use the errors with supported flags as the output)

For example in OpenBSD, `ls --version` will output

```
ls: unknown option -- -
usage: ls [-1AaCcdFfghikLlmnopqRrSsTtux] [file ...]
```

We can use `1AaCcdFfgh [...]` to detect that it's OpenBSD.

As for Mac, we can use the part where `U` exists.

iandrewt:
Alright, is it possible you can change it to that if we can't wait until mid next week? I should be studying for my HSC right now.

konimex:
I'll need the output of `ls --version` for Mac, though (including the error).

iandrewt:
```
ls: illegal option -- -
usage: ls [-ABCFGHLOPRSTUWabcdefghiklmnopqrstuwx1] [file ...]
```

PR #846, opened 2017-10-17
gpu_driver: Add macOS (Nvidia) support

iandrewt:
## Description

Adds `gpu_driver` support for macOS, really only affects Mac Pro and Hackintosh systems (although I think some older iMacs and MacBook Pros may have this work too). Only works for Nvidia cards with the Nvidia web drivers installed, otherwise it just falls back to "macOS Default Graphics Driver" (which is what the Nvidia web drivers refer to the included driver as, so I'm referring to the default driver for all GPUs as that)

### DIFF
--- a/config/config.conf
+++ b/config/config.conf
@@ -26,7 +26,7 @@ print_info() {
     info "GPU" gpu
     info "Memory" memory
 
-    # info "GPU Driver" gpu_driver  # Linux only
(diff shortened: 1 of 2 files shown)


iandrewt:
@dylanaraps one of the macOS builds for the PR is refusing to start, but the push test worked fine. Can I still merge this?

PR #871, opened 2017-11-22
disk: change Haiku's df detection

dawidd6:
'blocks' keyword is also present in busybox' df.
Busybox' df also has -P and -h flags, so it can be detected as normal df.
'Tracker' keyword seems to be exclusive to Haiku's df.

Tested on Alpine with busybox and on Haiku.


konimex:
Merged. Sorry I'm late. I was very busy these weeks.

dawidd6:
Nah, it's ok.

PR #872, opened 2017-11-23
fix in mac can not make_thumbnail

StarryTony:
# fix in mac high Sierra cannot replace '/' with '_', optionally can use ${image//\//_}

## Description

## Issues
in mac, iterm2 cannot process ${image//'/'//_}, therefore make_thumbnail() will fail and there will be no image displayed.
## TODO

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -2686,7 +2686,8 @@ get_image_size() {
 make_thumbnail() {
     # Name the thumbnail using variables so we can
     # use it later.
-    image_name="$crop_mode-$crop_offset-$width-$height-${image//'/'/_}"
+    # fix in mac cannot replace '/' with '_', optionally can use ${image//\//_}
+    image_name="$crop_mode-$crop_offset-$width-$height-${image##*/}"
 
     # Handle file extensions.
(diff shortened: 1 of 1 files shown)


neofetch
```diff
@@ -2686,7 +2686,8 @@ get_image_size() {
 make_thumbnail() {
     # Name the thumbnail using variables so we can
     # use it later.
-    image_name="$crop_mode-$crop_offset-$width-$height-${image//'/'/_}"
+    # fix in mac cannot replace '/' with '_', optionally can use ${image//\//_}
+    image_name="$crop_mode-$crop_offset-$width-$height-${image##*/}"
```
konimex: Because originally we used `${image//'/'/_}`, I think it's better to change it to `${image//\//_}` instead.

Aside from that, LGTM.

StarryTony: You are right. I have considered that. I am just trying to avoid generating very long filename (just in the case you wish to use images from any long filepath. And becasue there is no string checking, short filename could avoid potential issues from specific situations).

konimex:
Alright. Merging this one.

PR #891, opened 2018-01-05
OS: detect 32-bit IRIX

szfcbr:
## Description

neofetch currently only works on 64-bit installations of IRIX, change so that it matches on the uname output from 32-bit too. Works on my O2.

neofetch-o2

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -43,7 +43,7 @@ get_os() {
         "Haiku") os="Haiku" ;;
         "MINIX") os="MINIX" ;;
         "AIX") os="AIX" ;;
-        "IRIX64" | "IRIX") os="IRIX" ;;
+        "IRIX64") os="IRIX" ;;
         "FreeMiNT") os="FreeMiNT" ;;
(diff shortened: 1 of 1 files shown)


neofetch
```diff
@@ -43,7 +43,7 @@ get_os() {
         "Haiku") os="Haiku" ;;
         "MINIX") os="MINIX" ;;
         "AIX") os="AIX" ;;
-        "IRIX64") os="IRIX" ;;
+        "IRIX64" | "IRIX") os="IRIX" ;;
```
konimex: For simpler detection just change both `"IRIX64" | "IRIX")` to `"IRIX"*)`.

konimex:
Otherwise, LGTM

neofetch
szfcbr: Fair enough, unlikely to be some other kernel named irix-something

konimex:
LGTM. Thanks for the PR.

PR #892, opened 2018-01-05
Shells: add support for tcsh

szfcbr:
## Description

Specific support for detecting tcsh is needed as versions < 6.11.04 do not implement the --version flag. The only way to reliably get tcsh version regardless of release is to check the $tcsh variable

## Features

Also works when tcsh is emulating csh(getting the version of the original BSD csh is trickier though afaik)


konimex:
LGTM

PR #894, opened 2018-01-12
distro: fix OpenWrt/LEDE detection

dawidd6:
I was surprised that OpenWrt and LEDE are detected as generic Linux. Didn't noticed it or something. After this commit it's fine.


dylanaraps:
Thanks :+1:

PR #904, opened 2018-02-02
macOS fixes & improvements

### LINKED ISSUE #902: Several features broken due to 100-char line limit fixes
I've been trying to fix some issues with neofetch on macOS, and I've noticed most of them have to do with commits made ~6 months ago enforcing a 100-char line length.  For example, [the `awk` call checking for open music players](https://github.com/dylanaraps/neofetch/blame/2cecb142e1bc651427e0590672144b609714f903/neofetch#L1525) won't work for the first program on each line — `Google Play`, `xmms2d`, and `guayadeque`.  While I don't run any of those, I noticed it when I changed the regex for Spotify to `[sS]potify` which pushed it onto the next line.

Another place I've noticed this is the calls to `osascript` where the script to execute is split across multiple lines, reporting `syntax error: A unknown token can’t go after this property`.

I've been able to fix all of these by making them one-liners again (removing the backslash-newlines).  I'm not really a `bash` expert, so I'm not sure if there's another newline-escape method to use so I can continue my other macOS fixes.

### Code examples:

Detects Spotify on macOS (modified [line 1525](https://github.com/dylanaraps/neofetch/blame/2cecb142e1bc651427e0590672144b609714f903/neofetch#L1525)):
```bash
player="$(ps x | awk '!(/ awk|Helper|Cache|ibus|indicator/) && /mpd|mopidy|cmus|mocp|[sS]potify|\
Google Play|iTunes.app|rhythmbox|banshee|amarok|deadbeef|audacious|\
xmms2d|gnome-music|lollypop|clementine|pragha|exaile|juk|bluemindo|\
guayadeque|yarock|qmmp|quodlibet|deepin-music|tomahawk|pogo|elisa/ {printf $5 " " $6; exit}')"
```

Fails to detect Spotify on macOS:
```bash
player="$(ps x | awk '!(/ awk|Helper|Cache|ibus|indicator/) && /mpd|mopidy|cmus|mocp|\
[sS]potify|Google Play|iTunes.app|rhythmbox|banshee|amarok|deadbeef|audacious|\
xmms2d|gnome-music|lollypop|clementine|pragha|exaile|juk|bluemindo|\
guayadeque|yarock|qmmp|quodlibet|deepin-music|tomahawk|pogo|elisa/ {printf $5 " " $6; exit}')"
```

Detects Spotify on macOS:
```bash
player="$(ps x | awk '!(/ awk|Helper|Cache|ibus|indicator/) && /mpd|mopidy|cmus|mocp|[sS]potify|Google Play|iTunes.app|rhythmbox|banshee|amarok|deadbeef|audacious|xmms2d|gnome-music|lollypop|clementine|pragha|exaile|juk|bluemindo|guayadeque|yarock|qmmp|quodlibet|deepin-music|tomahawk|pogo|elisa/ {printf $5 " " $6; exit}')"
```

Fails to get artist & track from Spotify on macOS ([line 1580](https://github.com/dylanaraps/neofetch/blame/2cecb142e1bc651427e0590672144b609714f903/neofetch#L1580)):
```bash
song="$(osascript -e 'tell application "Spotify" to artist of current track as \
                      string & " - " & name of current track as string')"
```

Successfully gets artist & track from Spotify on macOS:
```bash
song="$(osascript -e 'tell application "Spotify" to artist of current track as string & " - " & name of current track as string')"
```

JadeMatrix:
## Description

Fixes #902 as well as makes a few fixes/tweaks for macOS; tested on 10.12.6

## Features

* Feature detection on macOS that relies on AppleScript snippets works again (#902)
* Music player detection on macOS is now more reliable at detecting iTunes vs. Spotify — no guarantees made if both are running (looks like it prefers iTunes on my machine)
* Now detects Light vs. Dark theme in addition to Blue vs. Graphite, outputting `WM Theme: Blue (Light)` (for example)
* neofetch's `uppercase()` function no longer prevents feature detection under `bash` < v4

## Issues

* Music player detection is still not perfect — ideally it would use fancier regex to ignore the iTunes Helper and Spotify daemons, but then the `grep` would match itself, so that would have to be handled as a failure to match any player.  This shouldn't be too much of a problem, because even if a user doesn't have Spotify installed or iTunes running, iTunes Helper will be running, so `grep` will match something.  The app-is-running checks in the AppleScript segments then keep neofetch from opening iTunes/Spotify simply to check if they're playing music.

## TODO

- [ ] Test if this works on non-macOS systems

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -3255,7 +3255,11 @@ strip_sequences() {
 }
 
 uppercase() {
-    ((bash_version >= 4)) && printf "%s" "${1^}"
+    if ((bash_version >= 4)); then
+        printf "%s" "${1^}"
+    else
+        echo $1
+    fi
 }
 
 # COLORS
--- a/neofetch
+++ b/neofetch
@@ -1522,7 +1522,7 @@ get_memory() {
 }
 
 get_song() {
-    player="$(ps -e | grep -m 1 -o -F \
+    player="$(ps -e | grep -m 1 -o \
                            -e "Google Play" \
                            -e "Spotify" \
                            -e "amarok" \
@@ -1537,7 +1537,7 @@ get_song() {
                            -e "exaile" \
                            -e "gnome-music" \
                            -e "guayadeque" \
-                           -e "iTunes.app" \
+                           -e "iTunes$" \
                            -e "juk" \
                            -e "lollypop" \
                            -e "mocp" \
@@ -1604,9 +1604,11 @@ get_song() {
 
                 "Mac OS X")
                     song="$(osascript <<END
-tell application "Spotify"
-    artist of current track as string & " - " & name of current track as string
-end tell
+if application "Spotify" is running then
+    tell application "Spotify"
+        artist of current track as string & " - " & name of current track as string
+    end tell
+end if
 END
 )"
                 ;;
@@ -1615,9 +1617,11 @@ END
 
         "itunes"*)
             song="$(osascript <<END
-tell application "iTunes"
-    artist of current track as string & " - " & name of current track as string
-end tell
+if application "iTunes" is running then
+    tell application "iTunes"
+        artist of current track as string & " - " & name of current track as string
+    end tell
+end if
 END
 )"
         ;;
--- a/neofetch
+++ b/neofetch
@@ -930,9 +930,7 @@ get_wm_theme() {
             wm_theme="$(PlistBuddy -c "Print AppleInterfaceStyle" "$global_preferences")"
             wm_theme_color="$(PlistBuddy -c "Print AppleAquaColorVariant" "$global_preferences")"
 
-            if [[ -z "$wm_theme" ]]; then
-                wm_theme="Light"
-            fi
+            [[ -z "$wm_theme" ]] && wm_theme="Light"
 
             if [[ -z "$wm_theme_color" ]] || ((wm_theme_color == 1)); then
                 wm_theme_color="Blue"
--- a/neofetch
+++ b/neofetch
@@ -1603,13 +1603,23 @@ get_song() {
                 "Linux") get_song_dbus "spotify" ;;
 
                 "Mac OS X")
-                    song="$(osascript -e 'tell application "Spotify" to artist of current track as string & " - " & name of current track as string')"
+                    song="$(osascript <<END
+tell application "Spotify"
+    artist of current track as string & " - " & name of current track as string
+end tell
+END
+)"
                 ;;
             esac
         ;;
 
         "itunes"*)
-            song="$(osascript -e 'tell application "iTunes" to artist of current track as string & " - " & name of current track as string')"
+            song="$(osascript <<END
+tell application "iTunes"
+    artist of current track as string & " - " & name of current track as string
+end tell
+END
+)"
         ;;
 
         "banshee"*)
@@ -1965,7 +1975,10 @@ get_term_font() {
         ;;
 
         "Apple_Terminal")
-            term_font="$(osascript -e 'tell application "Terminal" to font name of window frontmost')"
+            term_font="$(osascript <<END
+tell application "Terminal" to font name of window frontmost
+END
+)"
         ;;
 
         "iTerm2")
@@ -1976,7 +1989,10 @@ get_term_font() {
             # See: https://groups.google.com/forum/#!topic/iterm2-discuss/0tO3xZ4Zlwg
             local current_profile_name profiles_count profile_name diff_font none_ascii
 
-            current_profile_name="$(osascript -e 'tell application "iTerm2" to profile name of current session of current window')"
+            current_profile_name="$(osascript <<END
+tell application "iTerm2" to profile name of current session of current window
+END
+)"
 
             # Warning: Dynamic profiles are not taken into account here!
             # https://www.iterm2.com/documentation-dynamic-profiles.html
@@ -2620,7 +2636,10 @@ get_image_source() {
 get_wallpaper() {
     case "$os" in
         "Mac OS X")
-            image="$(osascript -e 'tell application "System Events" to picture of current desktop')"
+            image="$(osascript <<END
+tell application "System Events" to picture of current desktop
+END
+)"
         ;;
 
         "Windows")
--- a/neofetch
+++ b/neofetch
@@ -3265,7 +3265,7 @@ uppercase() {
     if ((bash_version >= 4)); then
         printf "%s" "${1^}"
     else
-        echo $1
+        echo "$1"
     fi
 }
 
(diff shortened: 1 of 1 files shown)


neofetch
```diff
@@ -926,14 +926,21 @@ get_wm_theme() {
...
         "Quartz Compositor")
-            wm_theme="$(PlistBuddy -c "Print AppleAquaColorVariant" \
-                        "${HOME}/Library/Preferences/.GlobalPreferences.plist")"
+            global_preferences="${HOME}/Library/Preferences/.GlobalPreferences.plist"
+            wm_theme="$(PlistBuddy -c "Print AppleInterfaceStyle" "$global_preferences")"
+            wm_theme_color="$(PlistBuddy -c "Print AppleAquaColorVariant" "$global_preferences")"
 
-            if [[ -z "$wm_theme" ]] || ((wm_theme == 1)); then
-                wm_theme="Blue"
+            if [[ -z "$wm_theme" ]]; then
```
```sh
[[ -z "$wm_theme" ]] && em_theme="Blue"
```

neofetch
```diff
@@ -1522,7 +1529,7 @@ get_memory() {
 }
 
 get_song() {
-    player="$(ps -e | grep -m 1 -o -F \
+    player="$(ps -e | grep -m 1 -o \
```
dylanaraps: I want to keep `-F` if possible. If we need regex we should benchmark the speed difference between the two.

neofetch
```diff
@@ -3232,7 +3262,11 @@ strip_sequences() {
 }
 
 uppercase() {
-    ((bash_version >= 4)) && printf "%s" "${1^}"
+    if ((bash_version >= 4)); then
+        printf "%s" "${1^}"
+    else
+        echo "$1"
```
```sh
printf "%s" "$1"
```

(For consistency)

neofetch
```diff
@@ -1603,13 +1610,27 @@ get_song() {
                 "Linux") get_song_dbus "spotify" ;;
 
                 "Mac OS X")
-                    song="$(osascript -e 'tell application "Spotify" to artist of current track as string & " - " & name of current track as string')"
+                    song="$(osascript <<END
```
dylanaraps: Can we indent these to match up with the lines above?

neofetch
JadeMatrix: Well, regex is only needed for the iTunes match, because it needs to see the `iTunes` binary running rather than `iTunesHelper`.  Without it, Spotify will never get detected.  Possibly a separate check for macOS?

neofetch
JadeMatrix: Not sure what you mean.  `AppleInterfaceStyle` is either `Dark` or missing/empty, and can be used with both Blue and Graphite.

neofetch
JadeMatrix: You mean like this?
```bash
        "spotify"*)
            case "$os" in
                "Linux") get_song_dbus "spotify" ;;

                "Mac OS X")
                    song="$(osascript <<END
                if application "Spotify" is running then
                    tell application "Spotify"
                        artist of current track as string & " - " & name of current track as string
                    end tell
                end if
END
)"
                ;;
            esac
        ;;
```

That's 99 chars wide, so still within the limit.

neofetch
JadeMatrix: Fixed

neofetch
dylanaraps: What I mean is there's no point in having `if; then fi` when there's only one statement inside the if. It's just a shorter way of writing the if.

neofetch
dylanaraps: Yeah

neofetch
JadeMatrix: Gotcha, I've changed it

neofetch
JadeMatrix: Fixed them in eb062af — looks like long AppleScript lines can be split with `\` in heredocs, so I was able to compress them a bit more.

dylanaraps:
This LGTM now. 👍 

Let me know if it's ready to be merged or not.

JadeMatrix:
Well, I agree about the `-F` flag to `grep`.  Actually, I'm not happy with the massive `grep` for player detection because of the issues I mentioned in the beginning, but I can't think of anything better and "it works on my machine."  If the performance is a big concern I can poke at it and see if I can come up with something more robust.

dylanaraps:
We could split it into two different `ps` commands. One for macOS and one for all other OS, the only downside is that we lose the simplicity of just one command.

We might be able to get around the ItunesHelper problem by matching `"Itunes "` (with a space). Does that work?

JadeMatrix:
Doesn't look like it, because my version is trying to match the executable name at the end of the `ps` line.  Here's all the iTunes-related processes running on my machine right now:
```
jadematrix-rmbp2-3:~ JadeMatrix$ ps -e | grep -F iTunes
  387 ??         0:07.08 /Applications/iTunes.app/Contents/MacOS/iTunesHelper.app/Contents/MacOS/iTunesHelper
 1252 ??         0:00.62 /Library/Frameworks/iTunesLibrary.framework/Versions/A/XPCServices/com.apple.iTunesLibraryService.xpc/Contents/MacOS/com.apple.iTunesLibraryService
21119 ??         0:12.68 /Applications/iTunes.app/Contents/MacOS/iTunes
80957 ??         0:12.25 /Library/Frameworks/iTunesLibrary.framework/Versions/A/XPCServices/com.apple.iTunesLibraryService.xpc/Contents/MacOS/com.apple.iTunesLibraryService
21368 ttys012    0:00.00 /usr/bin/grep --color=always -F iTunes
```
Last one being `grep` itself, of course.  Note the problem with `iTunesHelper` is that it's packaged as an application *inside* of the main iTunes application, so searching for `iTunes.app` matches both.  Since the helper's always running, it looks like iTunes is too.

Regarding the two separate `ps` commands, it might make more sense to run the existing one on all platforms, and do an additional check for `.app` versions on macOS.  On the other hand, I don't know if any other music players run daemons that should be ignored on macOS or other platforms; ideally this would handle those too.

dylanaraps:
Maybe we can literally match the new line char. Does “-F” support a way of doing this? 

I’ll be able to test when I’m back at home.

dylanaraps:
I did a google and it looks like “$’\n’” will work.

JadeMatrix:
[According to the die.net `man` page](https://linux.die.net/man/1/fgrep), `-F` means that the patterns passed to `grep` are treated as newline-separated lists of fixed patterns to match, i.e. `grep -F $'foo\nbar'` is the same as `grep -F -e "foo" -e "bar"`.  My `man` page doesn't say that, but I tested it and `$'iTunes\n'` ends up matching every running process because it contains an empty string (after the `\n`).

Looks like it might be simplest to leave it as-is for now, and figure something out if any performance/detection issues pop up.

dylanaraps:
Thanks :)

JadeMatrix:
Cool, I'll keep an eye on the issues to see if anything comes up 👍

PR #924, opened 2018-03-06
fix pacman name conflict bug

MitchWeaver:
Description: Add a os-release check to make sure the distro really is Arch Linux when `pacman` is in the $PATH

------

Reason: Before this same bug was solved by removing /usr/games from the $PATH. However this did not work for 2 reasons:

1. The pacman game could be somewhere else, (OpenBSD for example has /games, /usr/games, and /usr/local/games
2. It could be anything! Such as this color script in my ~/bin/ascii folder:

Image

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -484,16 +484,9 @@ get_uptime() {
 }
 
 get_packages() {
-    # Remove /usr/games from $PATH.
-    # This solves issues with neofetch opening the "pacman" game.
-    local PATH=":${PATH}:"
-    local PATH="${PATH/':/usr/games:'/:}"
-    local PATH="${PATH%:}"
-    local PATH="${PATH#:}"
-
     case "$os" in
         "Linux" | "BSD" | "iPhone OS" | "Solaris")
-            type -p pacman >/dev/null && \
+            type -p paclog-pkglist >/dev/null && \
                 packages="$(pacman -Qq --color never | wc -l)"
 
             type -p dpkg >/dev/null && \


konimex:
This patch won't work because there are several distros that use pacman and its /etc/os-release is not Arch (I can think about two out of my head, Manjaro and Artix Linux).

dylanaraps:
Can we somehow check that the pacman that is installed is the package
manager? Maybe a —version?

MitchWeaver:
Big mistake on my part, totally forgot about the derivatives.

-------

Now using `-Qq` to check for version of a package, using itself as we know it will be installed.

The output should just be "pacman", instead of pacman-$version because of the extra `q`.

dylanaraps:
Use [[ ]] and ==

Otherwise LGTM

MitchWeaver:
Ahh yes bashisms 😛

Fixed 👍

simoniz0r:
TBH, checking for pacman to be installed through pacman doesn't solve this problem.

The game `pacman` doesn't respond to any command line options and just launches regardless of what's passed to it.  When you run the check there, the `pacman` game will still launch, waiting for the user to close it until `neofetch` moves on.

You can't really rely on *all* programs responding to command line arguments; a lot of them don't... not really sure what a better solution would be here, though.

MitchWeaver:
Ah this is true...

This is a tough problem. An /etc/os-release for "Type=Arch", or something else greppable, but os-release just isn't reliable.

Perhaps some md5/hash trickery?

Perhaps this is better left as an `issue` than a PR and see if someone can engineer a solution.

MitchWeaver:
Sure, `paclog-pkglist`. Sounds the most esoteric. 👍

dylanaraps:
You’re missing a / in dev/null

konimex:
I'm going to merge this since this looks good enough.

Thanks for the PR!

dylanaraps:
I totally forgot about this. My bad

PR #930, opened 2018-03-20
Add ascii art support for openSUSE Tumbleweed

robertwolter:
openSUSE Tumbleweed uses a different logo than openSUSE Leap


konimex:
Merged, thanks for the PR.

PR #935, opened 2018-03-27
Fix kitty font_size regex

aidanharris:
## Description

This fixes a small bug in the regex used for retrieving the font size used in the kitty terminal.

The regex used is too greedy so matches the `map ctrl+shift+backspace restore_font_size` found in my config.

The proposed fix should match any preceding whitespace (before `font_size`) and require at least one whitespace character after `font_size` (so that we don't match too much).


dylanaraps:
LGTM :+1:

dylanaraps:
Can you post your kitty config?

dylanaraps:
Also, open a new issue. :)

PR #937, opened 2018-03-31
Install on OpenBSD

desnudopenguino:
## Description
The default file locations in the Makefile aren't proper for OpenBSD's file structure, and it adds unnecessary directories and "cruft" to the system, and makes the user add another directory to their PATH. This checks if the OS is OpenBSD, and if it is, puts the files in the proper location. It also (For OpenBSD) adds another dependency, GNU Make. I figured it would be easier to add that dependency than it would be to force it the other way and have all the GNU Make users load BSD Make (for the syntax to check the OS name).

## Features
Base install in OpenBSD.

## Issues
Requires GNU Make

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,20 +1,21 @@
-PREFIX = /usr
+PREFIX = /usr/local
+
 
 all:
 	@echo Run \'make install\' to install Neofetch
 
 install:
 	mkdir -p $(DESTDIR)$(PREFIX)/bin
-	mkdir -p $(DESTDIR)$(PREFIX)/share/man/man1
+	mkdir -p $(DESTDIR)$(PREFIX)/man/man1
 	mkdir -p $(DESTDIR)/etc/neofetch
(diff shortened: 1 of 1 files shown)


konimex:
I honestly don't know what is the proper structure of OpenBSD's hierarchy since I don't use it.

But looking from both `hier(8)` and [OpenBSD ports porting guide](https://www.openbsd.org/faq/ports/guide.html), I think neofetch already has a proper path.

(Assuming `PREFIX=/usr/local`)

ASCII files and default config files falls in `read-only architecture independent`, in `/usr/local/share`.

Manpages are installed in `$(MANDIR)`, so like Slackware, we can add `MANDIR=$(PREFIX)/man`. This allows them to be installed in `/usr/local/man`.

Also, before you add a PR, make sure to pull the latest master first.

I'm not going to close this one yet since I'm still not sure about the hierarchy.

desnudopenguino:
Ah, good point. you can kill this PR, as it can be achieved by using PREFIX=/usr/local.

PR #938, opened 2018-04-01
Use printf to trim instead of echo

JoshuaRLi:
## Description

This PR replaces the last `echo` in `neofetch` with an equivalent `printf`!

## Issues

@dylanaraps it looks like it's mostly equivalent, but there are a few extra spaces lying around (see screenshots below). I wasn't able to fix this issue at a cursory glance (neofetch is huge), do you have any insights?

Current `master`:

image

`better-trim`:

image

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -3263,7 +3263,7 @@ get_bold() {
 }
 
 trim() {
-    printf "%s" "${1//[[:space:]]/ }"
+    printf '%s\n' "$1" | tr -s "[:space:]"
 }
 
 trim_quotes() {
--- a/neofetch
+++ b/neofetch
@@ -3263,20 +3263,7 @@ get_bold() {
 }
 
 trim() {
-    # When a string is passed to "echo" all trailing and leading
-    # whitespace is removed and inside the string multiple spaces are
-    # condensed into single spaces.
-    #
-    # The "set -f/+f" is here so that "echo" doesn't cause any expansion
-    # of special characters.
-    #
-    # The whitespace trim doesn't work with multiline strings so we use
-    # "${1//[[:space:]]/ }" to remove newlines before we trim the whitespace.
-
-    set -f
-    # shellcheck disable=2086
-    builtin echo -E ${1//[[:space:]]/ }
-    set +f
+    printf "%s" "${1//[[:space:]]/ }"
 }
 
 trim_quotes() {


dylanaraps:
Yeah, no...

This doesn't actually work.

```sh
# Your printf:
black ~ > ./trim "          hello     world      "
          hello     world      black ~ > 

# Neofetch current:
black ~> ./trim2 "        hello     world     "
hello world
```

The neofetch output is the expected output.

JoshuaRLi:
Oops! Sorry about that, was making my own tiny fetch and needed a trim, and thought mine could just be dropped in.

How's this? Only quirk is if there are any leading spaces, there will remain one leading space in the output, but I'm getting the correct default `neofetch` output now.

```sh
trim() {
    printf '%s\n' "$1" | tr -s ' '
}
```

dylanaraps:
That adds the overhead of a pipe and a call to an external process. The difference is only 5ms~ but multiply that by every time we trim a string and you see where I’m going with this.

What’s the problem with echo? What was the motivation to replace it? I’m just curious.

JoshuaRLi:
I'm aware of the overhead, but 5 ms might be overestimating; I feel no difference on a low-end laptop. Depends on whether or not you think the tradeoff is worth it. I thought you didn't like `echo`? :stuck_out_tongue: 

https://github.com/dylanaraps/neofetch/search?q=echo&type=Commits
https://github.com/dylanaraps/neofetch/wiki/Neofetch-vs.-Screenfetch

dylanaraps:
Fixed in master. https://github.com/dylanaraps/neofetch/commit/a1044d728fd0260d66461a8cb357066a7b63b6e4

Here's a little test so you can see the overhead. Running trim on 100 strings in a loop. The echo/printf only version takes `5ms~` to finish 100 iterations of `"    hello    world   "` whereas your version with `| tr` takes a full `200ms~` to finish.

Run this with `./trim "    hello         world              "`.

```sh
#!/bin/bash
trim() {
    printf '%s\n' "$1" | tr -s "[:space:]"
}

trim2() {
    set -f
    set -- $*
    printf "%s\\n" "${*//[[:space:]]/ }"
    set +f
}

time for i in {1..100}; do
    trim "$1" >/dev/null
done

time for i in {1..100}; do
    trim2 "$1" >/dev/null
done
```

```sh
black ~ > ./trim 

# | tr
real	0m0.242s
user	0m0.085s
sys	0m0.232s

# just printf/echo
real	0m0.004s
user	0m0.004s
sys	0m0.000s
```

JoshuaRLi:
I stand corrected. Thanks @dylanaraps, and nice fix w/o `tr`. Wow.

PR #944, opened 2018-04-08
Font detection for suckless terminal (st)

aidanharris:
## Description

This adds font detection for the suckless terminal (st).

Command line arguments are checked first, if a font is not specified via command line arguments we run `strings` on the `st` binary and `grep` for something that looks like a font. In my testing this has worked well except in the following edge case:

**config.h**:
```h
static char font[] = "Liberation Mono";
```

The above is technically a valid font. As it doesn't contain anything that indicates it's a font declaration e.g "pixelsize=", "size=", etc it will not be matched.

Other than this one edge-case, in my testing the font has matched correctly 100% of the time. However, it's possible that some patches a user might have applied to `st` (or any modifications that a user has made) might cause `neofetch` to match the wrong thing. For that reason alone I understand if you want to close this pull-request.

For fonts specified as command-line arguments though e.g `st -f "Liberation Mono:pixelsize=12:antialias=true:autohint=true"` **should** be matched correctly.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -2178,6 +2178,32 @@ END
                          "${XDG_CONFIG_HOME}/sakura/sakura.conf")"
         ;;
 
+        "st")
+            if term_font="$(ps -o command= -p $parent | grep -E '\-f\s?.*?:?')"; then
+                term_font="${term_font/*-f/}"
+                term_font="${term_font/ -*/}"
+            else
+                binary=""
+
+                # On Linux we can get the exact path to the running binary through the procfs
+                # on other systems we just have to guess and assume `st` is invoked from somewhere
+                # in the users $PATH
+                [[ -L /proc/$parent/exe ]] && binary="/proc/$parent/exe"
+                [[ -z "$binary" ]] && binary="$(command -v st)"
+
+                # Grep the output of strings on the `st` binary for anything that looks vaguely
+                # like a font definition. NOTE: There is a slight limitation in this approach.
+                # Technically "Font Name" is a valid font. As it doesn't specify any font options
+                # though it is hard to match it correctly amongst the rest of the noise.
+                [[ -n "$binary" ]] && \
+                    term_font="$(strings "$binary" | \
+                    grep -E '(pixelsize=|size=|antialias=|autohint=)' | head -1)"
+            fi
+
+            term_font="${term_font/xft:}"
+            term_font="${term_font/:*}"
+        ;;
+
         "terminology")
             term_font="$(strings "${XDG_CONFIG_HOME}/terminology/config/standard/base.cfg" |\
                          awk '/^font\.name$/{print a}{a=$0}')"
--- a/neofetch
+++ b/neofetch
@@ -2179,6 +2179,8 @@ END
         ;;
 
         "st")
+            [[ -z "$parent" ]] && parent="$(get_ppid "$PPID")"
+
             term_font="$(ps -o command= -p "$parent" | grep -F -- "-f")"
             if [[ "$term_font" ]]; then
                 term_font="${term_font/*-f/}"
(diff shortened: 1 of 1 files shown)


dylanaraps:
Thanks for the PR.

I don't think I'll merge this though. It's got too many edge cases.

dylanaraps:
I spent some time thinking about this and if we can add some way to avoid printing anything when we hit an edge case I'll merge this.

neofetch
```diff
@@ -2178,6 +2178,32 @@ END
                          "${XDG_CONFIG_HOME}/sakura/sakura.conf")"
         ;;
 
+        "st")
+            if term_font="$(ps -o command= -p "$parent" | grep -E '\-f\s?.*?:?')"; then
+                term_font="${term_font/*-f/}"
+                term_font="${term_font/ -*/}"
+            else
+                binary=""
```
dylanaraps: This isn't needed.

neofetch
```diff
@@ -2178,6 +2178,32 @@ END
                          "${XDG_CONFIG_HOME}/sakura/sakura.conf")"
         ;;
 
+        "st")
+            if term_font="$(ps -o command= -p "$parent" | grep -E '\-f\s?.*?:?')"; then
```
```sh
term_font="$(ps -o command= -p "$PPID" | grep -F -- "-f")"

if [[ "$term_font" ]]; then
```

neofetch
```diff
@@ -2178,6 +2178,32 @@ END
...
+                term_font="${term_font/*-f/}"
+                term_font="${term_font/ -*/}"
+            else
+                binary=""
+
+                # On Linux we can get the exact path to the running binary through the procfs
+                # on other systems we just have to guess and assume `st` is invoked from somewhere
+                # in the users $PATH
+                [[ -L /proc/$parent/exe ]] && binary="/proc/$parent/exe"
+                [[ -z "$binary" ]] && binary="$(command -v st)"
```
dylanaraps: We know `$binary` will be unset/empty so we can remove the test.

Also, change `command -v` to `type -p` for consistency with the rest of the script.

neofetch
```diff
@@ -2178,6 +2178,32 @@ END
...
+
+                # Grep the output of strings on the `st` binary for anything that looks vaguely
+                # like a font definition. NOTE: There is a slight limitation in this approach.
+                # Technically "Font Name" is a valid font. As it doesn't specify any font options
+                # though it is hard to match it correctly amongst the rest of the noise.
+                [[ -n "$binary" ]] && \
+                    term_font="$(strings "$binary" | \
```
dylanaraps: This should be faster. It removes the pipe and `head` call and also does fixed string matching with `grep`.

```sh
term_font="$(strings "$binary" | \
             grep -F -m 1 -e "pixelsize=" -e "size=" -e "antialias=" -e "autohint=")"
```

neofetch
aidanharris: From what I can see this is still needed:

```bash
                term_font="${term_font/*-f/}"
                term_font="${term_font/ -*/}"
```

I've removed the `binary=""`

neofetch
aidanharris: `$PPID` from what I can see is the parent of `neofetch` (the shell: `zsh` in my case). `$parent` is the terminal.

neofetch
aidanharris: I've changed this as follows:

```bash
[[ -L /proc/$parent/exe ]] && binary="/proc/$parent/exe" || binary="$(type -p st)"
```

dylanaraps:
$parent can be changed to $PPID since $parent may not be set.

aidanharris:
@dylanaraps We can't use `$PPID` since this is the process of the shell (e.g `bash` or `zsh`) and not the terminal (`st`). I've added an explicit check for `$parent` though and set it if it's not set.

dylanaraps:
LGTM :+1:

PR #945, opened 2018-04-09
Misc: change all escape sequences from \\e to \e

### LINKED ISSUE #933: Unexpected backslash being inserted into title variable upon Neofetch launch
## Description
For some reason, a backslash is being inserted into the title variable, so the title shows up as `kamran\@Aurora` instead of `kamran@Aurora`. I know my username doesn't have a backslash in it, so that's not an issue. I have no idea why this is being caused. The issue happens in both Apple Terminal, and iTerm 2. My config file, and the verbose log, are attached below as a gist link.

#### Neofetch version
`3.3.1-git` (Homebrewed)

## Screenshot
screen shot 2018-03-24 at 4 28 16 pm

## Config file & verbose log
[Gist link](https://gist.github.com/KamranMackey/bd90826e1517feff2ecde700559d948c)

konimex:
## Description

Self-explanatory. Closes #933.

Interestingly enough, this only happens in UNIX-certified systems (macOS, AIX, Solaris), but not Linux (including Android/Termux) and BSDs.

## Issues

The most obvious one is this commit will trigger [SC1117](https://github.com/koalaman/shellcheck/wiki/SC1117). We'd have to exclude it, or change the double quotes to a single quote.

## TODO

Testing

- [x] Linux
- [ ] BSDs
- [x] MacOS (see issue)
- [ ] Windows
  - [ ] Cygwin/MSYS2
  - [ ] Windows subsystem for Linux
- [x] AIX
- [x] Sol aris
- [ ] IRIX

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ script:
     - time ./neofetch --ascii --config config/travis.conf -v
     # See this wiki page for why we're disabling these errors.
     # https://github.com/dylanaraps/neofetch/wiki/Shellcheck-Exclusions
-    - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then shellcheck neofetch -e SC1090,SC2009,SC2012,SC2016,SC2034,SC2128,SC2153,SC2154,SC2178,SC2010,SC1004; fi
+    - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then shellcheck neofetch -e SC1090,SC2009,SC2012,SC2016,SC2034,SC2128,SC2153,SC2154,SC2178,SC2010,SC1004,SC1117; fi
     # The if statement is here to invert the exit code from grep.
     # grep normally errors if no match is found but we want the opposite.
     # We invert it so grep fails if a match is found.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,2 +1,8 @@
 ## Contributors
 
+
+## Info
+
+**Title**
+
+- Fixed unexpected backslash being inserted on some systems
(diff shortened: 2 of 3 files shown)


iandrewt:
I was doing a bit more testing, oddly if I compile bash 4.2.10 (the version used in @konimex's linux screenshot) on my Mac, the issue still occurs (without this patch), but bash 4.4.19 from Homebrew works (and compiling 4.4.18 works too)

iandrewt:
Also, changing it to `\e` works better than just fixing the bash version, bash 4.4.19 will bold&colour the `@` without this patch but with it it's white and not-bold.

dylanaraps:
`\e` :+1:

konimex:
Hmm.. For some reason Shellcheck didn't detect SC1117. Maybe because this Shellcheck version was older when SC1117 was released.

iandrewt:
Why does shellcheck return errors for SC2206 and SC2207 when I run it on my Mac (or even an Arch Linux VM) but not in Travis? I get the SC1117 errors on the master branch too. Should I make a seperate issue for this?

konimex:
SC1117, SC2206, and SC2207 was released after Shellcheck 0.4.7, right now Travis has Shellcheck 0.4.6.

iandrewt:
Ah yeah, right now the [ShellCheck readme has instructions for manually upgrading the Travis version of ShellCheck if needed](https://github.com/koalaman/shellcheck#travis-ci)

dylanaraps:
Cool, we can simplify our Travis config and remove the install part for shell check since it’s installed by default.

PR #953, opened 2018-04-19
Add 'kitty' image backend

SolitudeSF:
## Description

Added [kitty](https://github.com/kovidgoyal/kitty) terminal image output.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -2629,7 +2629,7 @@ image_backend() {
         "off") image_backend="off" ;;
 
         "caca" | "catimg" | "jp2a" | "iterm2" | "termpix" |\
-        "tycat" | "w3m" | "sixel" | "pixterm")
+        "tycat" | "w3m" | "sixel" | "pixterm" | "kitty")
             get_image_source
 
             if [[ ! -f "$image" ]]; then
@@ -2654,7 +2654,7 @@ image_backend() {
 
         *)
             err "Image: Unknown image backend specified '$image_backend'."
-            err "Image: Valid backends are: 'ascii', 'caca', 'catimg', 'jp2a', 'iterm2',
+            err "Image: Valid backends are: 'ascii', 'caca', 'catimg', 'jp2a', 'iterm2', 'kitty',
                                             'off', 'sixel', 'pixterm', 'termpix', 'tycat', 'w3m')"
             err "Image: Falling back to ascii mode."
             get_ascii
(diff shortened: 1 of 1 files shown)


neofetch
```diff
@@ -2979,11 +2979,16 @@ make_thumbnail() {
     image_name="$crop_mode-$crop_offset-$width-$height-${image##*/}"
 
     # Handle file extensions.
-    case "${image##*.}" in
-        "eps"|"pdf"|"svg"|"gif"|"png")
-            image_name+=".png" ;;
-        *)  image_name+=".jpg" ;;
-    esac
+    if [ $image_backend == 'kitty' ]; then
```
konimex: Note that bashism is preferred so use `[[` instead of `[`.

Snuggle:
My gosh, I would love to see this merged. +1
Thanks for the PR, @SolitudeSF! :heart:

dylanaraps:
Merged manually with some changes. https://github.com/dylanaraps/neofetch/commit/2606e4b1d6c431efc2a9bea4f0bec2385ad805f2

Thanks!

SolitudeSF:
Kitty can display png images directly. If you supply a jpg or other format it uses imagemagick to convert it to png, so its additional startup overhead, is there no way to include special case for kitty?

dylanaraps:
Neofetch doesn’t convert the image, it just appends png to the filename

SolitudeSF:
in understand, but you call imagemagick anyway to generate thumbnail. or am i misunderstanding something?

dylanaraps:
Yes, `neofetch` generates a thumbnail. We do this to speed up the display of images. Neofetch mostly deals with wallpaper sized files so this is necessary to avoid 1s+ run times.

The lines of code you added around the `case` statement don't actually do anything and they don't skip the thumbnail generation.

SolitudeSF:
yes, they dont skip, they force thumbnail to `png` format when kitty is used to avoid additional convertation call from kitty's side.

dylanaraps:
This sounds like an issue with kitty more so than neofetch. Why should we add a workaround when it can be fixed directly?

SolitudeSF:
thats not an issue, thats how it works, it can display png data directly. i just optimize one `convert` call away.

SolitudeSF:
actually, after benchmarking both methods, my *workaround* consistently yields slower result, which i dont understand, but at least this issue is finally closed.

PR #955, opened 2018-04-20
Add 'setroot' wallpaper fetching

SolitudeSF:
## Description

Adds fetching wallpaper set with [`setroot`](https://github.com/ttzhou/setroot).
I added it with higher priority than `feh`, since it can be used only as wallpaper setter, while `feh` could still be used as image viewer. ~~even tho there is probably zero overlap between users.~~

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -2802,6 +2802,11 @@ END
                     if type -p feh >/dev/null && [[ -f "${HOME}/.fehbg" ]]; then
                         image="$(awk -F\' '/feh/ {printf $(NF-1)}' "${HOME}/.fehbg")"
 
+                    elif type -p setroot >/dev/null && \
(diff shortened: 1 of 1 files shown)


dylanaraps:
Put this after `feh`.

Also, lines need to be below 100 chars long:

```sh
elif type -p setroot >/dev/null && \
     [[ -f "${XDG_CONFIG_HOME}/setroot/.setroot-restore" ]]; then
    image="$(awk -F\' '/setroot/ {printf $(NF-1)}'
             "${XDG_CONFIG_HOME}/setroot/.setroot-restore")"
```

SolitudeSF:
done

dylanaraps:
Thanks :+1:

PR #963, opened 2018-04-27
de: fix Unity in 18.04

dawidd6:
XDG_CURRENT_DESKTOP in Ubuntu 18.04 with Unity is containing some extra info.

Before patch:
zrzut ekranu z 2018-04-27 11-56-56

After:
zrzut ekranu z 2018-04-27 11-57-42


dylanaraps:
LGTM :+1:

PR #965, opened 2018-05-03
Distribute neofetch as a single executable.

dylanaraps:
## Description

The goal of this PR is to make installing neofetch as easy as popping a file in your `$PATH`.

~~Note: This may or may not be merged. It's an experiment.~~

## TODO

- [x] Inline config file.
- [x] Inline ascii files.
- [x] Adjust Makefile.


dylanaraps:
Thoughts on this?

@konimex  @iandrewt

simoniz0r:
Just wanted to comment that I think this is a pretty cool idea and it works as expected on openSUSE Tumbleweed.
Perhaps it would be a better idea to put the ascii at the end of the script to make it easier to add new ones in the future?

dylanaraps:
> Just wanted to comment that I think this is a pretty cool idea and it works as expected on openSUSE Tumbleweed.

Thanks for testing. :+1: 

> Perhaps it would be a better idea to put the ascii at the end of the script to make it easier to add new ones in the future?

That's a good idea. Consider it done. :)

Morganamilo:
Why not just a Makefile that generates the single file from the other source files. Also if you're intending to distribute this as single file, maybe add the license as a comment in the code.

dylanaraps:
Do we still need a Makefile if neofetch is only an executable and a man page? The installation process is only a copy/paste now so it seems a little redundant.

dylanaraps:
I've changed my stance on this PR and it will be merged. If anyone has any opinions or thoughts about this, speak up. :)

dylanaraps:
This makes Neofetch much easier to install and distribute especially on operating systems and distributions that don't have Neofetch in their repositories. With this PR all you need to do to install Neofetch is to download the executable. It's as simple as that.

The only downside to this PR is that in-lining the configuration file and ASCII art doubled the size of the Neofetch executable. Contribution will be harder for newcomers as they won’t know how to navigate the file efficiently.

Once this is merged I'll be finally writing a `CONTRIBUTING.md` file outlining how to add a new distribution to Neofetch as well as adding new info functions, etc. Lowering the barrier of entry for those looking to contribute and hopefully alleviating the downsides of this change.

konimex:
I have no comment either way, but does this mean that the manpage is also no longer needed?

Morganamilo:
>The only downside to this PR is that in-lining the configuration file and ASCII art doubled the size of the Neofetch executable. Contribution will be harder for newcomers as they won’t know how to navigate the file efficiently.

This is why I would prefer separate files that are then merged together using a makefile. You code make the ascii files be actual bash that just saves the ascii to a variable, including the file would then give you access to the variable.

dylanaraps:
The man page is optional for installation. If a user just wants the executable they can do a `wget`/`curl` and have the latest Neofetch version "just work". Distributions packaging Neofetch can install the man page same as before.

iandrewt:
Sounds good to me. The space doubling doesn't really matter imo, in fact it saves a lot of space on 4K sector drives (the ASCII files are all less than 4K, for instance the Mac one is about 500B, but the drive can only store files in sizes multiples of 4K). Only immediate issue I can think of is breaking repo installs that are using `HEAD`

iandrewt:
Also, I'm assuming this would end up being neofetch 4.0? Given it won't exactly be compatible with previous versions

dylanaraps:
> This is why I would prefer separate files that are then merged together using
> a makefile. You code make the ascii files be actual bash that just saves the
> ascii to a variable, including the file would then give you access to the
> variable.

I don’t think using a `Makefile` to create a single executable is
possible. I can’t see how we would include the 150 separate ASCII files using
some sort of build script.

The `Makefile` *would* work if we checked for the current operating system and
only included the specific ASCII file the OS needs. However this would break
`--ascii_distro distro` and would require a “re-build” per OS.

The in-lined ASCII art in this pull request *are* stored as variables but
they’re stored as plain text same as separate files. The variables in the ASCII
are not interpreted by `bash`, instead we do a search and replace to add the
color escape sequences.

Do you have any ideas on how the `Makefile` would build Neofetch? This includes
in-lining the configuration file and in-lining all 150 ASCII art files. The
`Makefile` would need to also dynamically in-line the ASCII art as we add new
distributions to Neofetch at least once per week.


> Also, I'm assuming this would end up being neofetch 4.0? Given it won't
> exactly be compatible with previous versions

Yes.

dylanaraps:
This PR will be merged in 2 hours, I'm still going through the source looking for new bugs or errors made when merging the files. 

If anybody has any further thoughts about this, please don't hesitate to comment. :)

iandrewt:
I'll work on fixing the homebrew package for when this is merged so it's ready to go

iandrewt:
Will 4.0 be released as soon as this is merged or will the changed stay in master for a while?

dylanaraps:
4.0 will be released either tonight or tomorrow. It won't happen right after the merge.

fxcoudert:
Hi, Homebrew maintainer here. Please don't disable the ability to install with `make install PREFIX=/foo/bar`, which is precious.

The current pull request (https://github.com/Homebrew/homebrew-core/pull/27485) means installing script to `bin` and man page to the man directory. This is more fragile: if in the future you introduce a helper script, or documentation, an info page, or anything else, it will mean every distributor has to adapt their packaging. On the other hand, providing a Makefile means if you add or remove installable files, all you have to do is adjust your own Makefile, and installation still works without a hitch.

dylanaraps:
Done. 

https://github.com/dylanaraps/neofetch/blob/master/Makefile

fxcoudert:
Many thanks!

PR #967, opened 2018-05-03
term_font: add qterminal support

mstraube:
## Description
self-explanatory

## Issues
Works only after the user changed and saved some settings.


dylanaraps:
LGTM :+1:

PR #968, opened 2018-05-04
Combine

Morganamilo:
## Description

>Do you have any ideas on how the Makefile would build Neofetch? This includes
in-lining the configuration file and in-lining all 150 ASCII art files. The
Makefile would need to also dynamically in-line the ASCII art as we add new
distributions to Neofetch at least once per week.

Implements this idea, code is generated from the ASCII data and get_distro_colors.in as each image has its own colours.

## Issues

I whipped this up in 30 minutes as a proof of concept, it appears to work but I have not done to much testing. `make install` seems to fail but I think that's because of the current state of the one branch, not something I did.


dylanaraps:
Thanks for working on this.

The PR seems to be missing the `Makefile` and the referenced `get_distro_colors.in`.

Morganamilo:
My bad missed the add.

konimex:
So.. How would one run neofetch if the binary doesn't even exist?

Morganamilo:
`make neofetch`

dylanaraps:
This PR also seems to be missing the actual Neofetch source. :)

konimex:
And where's the Makefile? Also, ugh. `.in` files. I really don't like it in a program like Neofetch.

Morganamilo:
I'm a bit too used to `git commit -a` sorry.

Morganamilo:
If you don't like the .in and all that it's fine, I just have a dislike of too much code lumped in one file.

Just wanted to give some actual code, not just spout ideas, was quick to do, accept it or not I don't mind.

Either way thanks for creating this program in the first place :)

dylanaraps:
I'm still set on the other PR but it's good to know that the `Makefile` method is possible. I agree with the too much code in a single file but it can't really be helped with `bash`. Had Neofetch been written in Python, C or another language it'd be split into separate files from the start.

Bash just doesn't have first class support for this. I know there's `source` but it can't be used in this context. :woman_shrugging:

Morganamilo:
Alrighty glad you got to consider both sides.

PR #970, opened 2018-05-04
Add boot_method that can display BIOS or UEFI

mfs:
## Description

This PR adds a boot_method option that displays BIOS, UEFI or UNKNOWN

## TODO
I'd like to add support for more OSs though am not sure how to detect boot method on them. Modern Mac OS X is probably UEFI, not sure what Windows does these days. Can do some research.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -1203,9 +1203,6 @@ get_boot_method() {
         "Linux")
             [ -d /sys/firmware/efi ] && boot_method="UEFI" || boot_method="BIOS"
         ;;
-        *)
-            boot_method="UNKNOWN"
-        ;;
     esac
 }
 
--- a/neofetch
+++ b/neofetch
@@ -77,6 +77,7 @@ print_info() {
 
     # info "GPU Driver" gpu_driver  # Linux/macOS only
     # info "CPU Usage" cpu_usage
+    # info "Boot Method" boot_method
     # info "Disk" disk
     # info "Battery" battery
     # info "Font" font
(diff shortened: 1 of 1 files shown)


neofetch
```diff
@@ -1197,6 +1198,17 @@ get_kernel() {
     fi
 }
 
+get_boot_method() {
+    case "$os" in
+        "Linux")
+            [ -d /sys/firmware/efi ] && boot_method="UEFI" || boot_method="BIOS"
```
dylanaraps: Use `[[ ]]` and quote the path.

neofetch
```diff
@@ -1197,6 +1198,17 @@ get_kernel() {
...
+        ;;
+        *)
```
dylanaraps: Remove this whole block. Neofetch doesn't print "Unknown" for anything, instead it silently skips displaying the information. Using `-v` will show you what couldn't be detected instead.

iandrewt:
Modern macOS is always (U)EFI, since Mac OS X Tiger 10.4 on Intel. Not sure about PowerPC macs though

neofetch
mfs: Ah, nice. Will take care of these issues and then look at what is involved in supporting some of the other OS.

iandrewt:
I think the best way to check for EFI on macOS is to check for the existence of `/Applications/Utilities/Boot Camp Assistant.app`, it's been around since Mac OS X 10.5 Leopard and in beta form on Mac OS X 10.4 Tiger (for 10.4.6 and up). Intel (and EFI) support was added in 10.4.4.

I could be completely wrong about how this is worded though. Older (pre-Intel) Macs (so 10.4.3 and older) used different firmwares like Open Firmware and New World ROM, and I'm unsure how this slots in with the EFI vs BIOS comparison.

mfs:
Yeah, this is proving to be a bit awkward to work out on non Linux operating systems. The BSDs, well OpenBSD, do a check for the EFI frame buffer to check if they are in UEFI mode. I'll take another look at this on the weekend.

PR #978, opened 2018-05-09
song: fix audacious album

mstraube:
fallback to get_song_dbus() when song_shorthand is on
audtool won't work due to the `-` separator in it's output


dylanaraps:
LGTM :+1:

PR #985, opened 2018-05-12
song: update song_shorthand example

mstraube:
## Description

Add album to the song_shorthand example.


iandrewt:
Thanks!

LGTM 👍

PR #999, opened 2018-05-20
macOS: work get_term_size correctly

MakeNowJust:
## Description

After 80299bd064f9c2873305783cb538b1b14c6d0748, get_term_size on macOS is failed always, so image is not shown.

This PR corrects get_term_size to return term size if escape sequences work.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -3688,8 +3688,8 @@ get_term_size() {
 
     # If the sequences above don't work and the user is on a macOS system
     # or a system not running an X server, return early.
-    [[ -z "$DISPLAY" ]] && \
-        { term_width=0; return; }
+    [[ -z "$DISPLAY" || "$os" == "Mac OS X" ]] && \
+        { : "${term_width:=0}"; return; }
 
     # Get terminal width/height if \e[14t is unsupported.
     if [[ -z "$term_width" ]] || (( "$term_width" < 50 )); then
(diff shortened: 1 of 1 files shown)


dylanaraps:
What you wrote isn't valid bash. The change in master is working as intended. It tried universal techniques for window size then return on macOS before the X server stuff.

MakeNowJust:
Sorry I fix it soon.

However I think 80299bd064f9c2873305783cb538b1b14c6d0748 change is wrong totally: Why is it assign `0` to `term_width` before early return ([here](https://github.com/dylanaraps/neofetch/blob/80299bd064f9c2873305783cb538b1b14c6d0748/neofetch#L3692))?

dylanaraps:
I don't see the point in these changes at all.

MakeNowJust:
`term_width=0` in [line 3692](https://github.com/dylanaraps/neofetch/blob/master/neofetch#L3692) does not make sense to me. It discards the result of universal techniques for window size. And it is the only reason not to show image at all. It should be fixed.

You pushed fc229704c8ee03452f8f4e5cf30bccbe1e4bdd7d an hour ago, which removes macOS from early return condition. It can solve this issue, but we will have another issue. It lets xquartz start on macOS if failed to get window size for some reasons.

I will push very simplified version of this. A change I need is this in fact.

MakeNowJust:
Good. Thank you.

PR #1003, opened 2018-05-24
song: Fix matching

dylanaraps:
This fixes cases where `ItunesHelper` is running but `Itunes` is not causing an incorrect match. This needs testing as the output of `ps -e` may differ on different machines.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -2313,7 +2313,7 @@ get_memory() {
 }
 
 get_song() {
-    player="$(ps -e | grep -m 1 -o \
+    player="$(ps -e | grep -m 1 \
                            -e "Google Play" \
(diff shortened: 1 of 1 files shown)


JadeMatrix:
`iTunes Helper` is a daemon that is always running on macOS; removing the `$` from the regex means that neofetch will fail to detect the song being played by any player other than iTunes.  See be893d4

iandrewt:
That’s actually the commit that broke the detection. Adding the `-F` to the grep statement fixed it

JadeMatrix:
Hmm, well detection's broken on 10.12.6 now — neofetch won't detect Spotify now.

PR #1014, opened 2018-06-01
remove SC2010 ShellCheck exception

stevestock:
title says it all.  If you merge this, you may also want to [update the wiki](https://github.com/dylanaraps/neofetch/wiki/Shellcheck-Exclusions#sc2010)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ script:
     - time ./neofetch --travis -v
     # See this wiki page for why we're disabling these errors.
     # https://github.com/dylanaraps/neofetch/wiki/Shellcheck-Exclusions
-    - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then shellcheck neofetch -e SC1090,SC2009,SC2012,SC2016,SC2034,SC2128,SC2153,SC2154,SC2178,SC2010,SC1004,SC1091,SC1117; fi
+    - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then shellcheck neofetch -e SC1090,SC2009,SC2012,SC2016,SC2034,SC2128,SC2153,SC2154,SC2178,SC1004,SC1091,SC1117; fi
     # The if statement is here to invert the exit code from grep.
     # grep normally errors if no match is found but we want the opposite.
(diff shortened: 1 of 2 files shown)


dylanaraps:
There's more than one instance of SC2010. Also `find` should be able to count the files too.

stevestock:
It doesn't look like there are other instances of [SC2010](https://github.com/koalaman/shellcheck/wiki/SC2010).
In fact, [travis ran clean without the SC2010 exclusion](https://travis-ci.org/dylanaraps/neofetch/jobs/386844928#L446), by just removing the one instance mentioned in the [wiki](https://github.com/dylanaraps/neofetch/wiki/Shellcheck-Exclusions#sc2010)

And I did some searching and don't think that [find](https://linux.die.net/man/1/find) can count.  Please point me in the right direction if you think/know otherwise.

dylanaraps:
My bad.

We can probably do something like this which is even faster.

```sh
for _ in "${1:-$HOME}"/*; do ((packages+=1)); done
```

dylanaraps:
See #1015

stevestock:
instead of replacement, it looks like the code was removed entirely in ffcf82d10d4e6c610e8e67c8570e5650be987d65.  Number of packages has reduced significantly vs previous revisions of neofetch and screenfetch. It looks like many other changes were made to reduce lint errors, but the travis.yml file still has not been updated.  Let me know if you'd like any help

iandrewt:
ffcf82d was done because binaries that matched that criteria actually weren't from packages at all, they were just binaries placed there from other applications like VirtualBox.

dylanaraps:
Good catch on the travis file.

stevestock:
@dylanaraps want me to fix, or you already got it?

dylanaraps:
Done it, thanks for the offer though.

stevestock:
no problem.  Don't forget the [wiki](https://github.com/dylanaraps/neofetch/wiki/Shellcheck-Exclusions) too

dylanaraps:
That'll happen after a release is made :)

stevestock:
oh ok, got it

PR #1016, opened 2018-06-02
info: Remove get_install_date

dylanaraps:
- It parses `ls` output.
    - There's no alternative method of doing this (*but still*).
- It incorrectly assumes that  modification time is creation time.
    - Results can easily be inaccurate if the system or user modified one of the files.
- It requires over 150 lines of code to function.
    - This includes a separate function that does manual (but awesome) conversion of date formats.


arvchristos:
Hello @dylanaraps 
I have some things to add to this discussion that occured to me while developing a script to print elapsed time from installation.
*  >  It parses ls output. There's no alternative method of doing this (but still).

 There is an alternative to ls and that is stat. So instead of parsing ls output one could use the following:
 ```bash
 stat /lost+found | awk '/Birth:/ {print $2 $3 $4}'
 ``` 

*  >  It incorrectly assumes that modification time is creation time. 

This is an existing problem for certain distributions. There is a great number of distributions that provide certain logs for installation time (e.g. debian based have /var/log/installer and arch based have pacman.log )

* > It requires over 150 lines of code to function. 
This is true regarding the separate function and I think its unavoidable as for now.

I am more than willing to re-implement this one if needed, mainly because I think it is a useful and funny statistic to be offered.

PR #1018, opened 2018-06-03
Packages: Add support for Flatpak and Snap

konimex:
### Features

Self-explanatory.

### Issues

In case of Snap, it needs to be tested and I can't test it on any of my systems since Snap for some reason depends on systemd and none of my machine have systemd installed currently.

I'm not sure that the offset by 1 is needed for Snap since in case of Flatpak, even if they have offset 1, they're still accurate (bash didn't include the first offset line).

2018-06-03-205342_804x604_scrot

### TODO

Testing for Snap

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -1261,6 +1261,7 @@ get_packages() {
             has "sorcery"    && tot gaze installed
             has "alps"       && tot alps showinstalled
             has "butch"      && tot butch list
+            has "flatpak"    && tot flatpak list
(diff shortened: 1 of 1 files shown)


NatoBoram:
Let us know when it's published in the Snap Store :)

simoniz0r:
I'm confused why a bash script that needs nothing to run would need a universal package :thinking:

PR #1023, opened 2018-06-07
song: use get_song_dbus for pragha

mstraube:
## Description

Pragha supports mpris2.
Simplify song function by using `get_song_dbus`.


dylanaraps:
Thanks 👍

PR #1025, opened 2018-06-08
song: use get_song_dbus for amarok

mstraube:
## Description

Amarok supports mpris2 so we can use `get_song_dbus`.
Exaile 4.0.0 will also support mpris2, added a reminder for now.


dylanaraps:
:+1:

PR #1031, opened 2018-06-14
Use nvidia-smi for NVIDIA GPUs

StasE9o:
## Description

On systems with multiple NVIDIA GPUs neofetch returns inaccurate results (even number of GPUs). It is possible to list GPUs using nvidia-smi instead if it's installed.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -2021,6 +2021,16 @@ get_cpu_usage() {
     esac
 }
 
+prin_gpu() {
+    if [[ "$gpu_brand" == "off" ]]; then
+        gpu="${gpu/AMD }"
+        gpu="${gpu/NVIDIA }"
+        gpu="${gpu/Intel }"
+    fi
+
+    prin "${subtitle:+${subtitle}${gpu_name}}" "$gpu"
+}
+
 get_gpu() {
     case "$os" in
         "Linux")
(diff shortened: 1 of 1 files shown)


hykilpikonna:
On my machine with 3060 Ti LHR, using this would print `NVIDIA NVIDIA GeForce RTX 3060 Ti`

The existing method prints `NVIDIA GeForce RTX 3060 Ti Lite Hash Rate`, which has more information,

image

neofetch
```diff
@@ -2041,6 +2051,17 @@ get_gpu() {
                "${gpus[1]}" == *Intel* ]] && \
                unset -v "gpus[0]"
 
+            # If nvidia-smi is installed, use that for NVIDIA GPUs
+            nv_smi_cmd="$(command -v nvidia-smi)"
+            if [[ -x "$nv_smi_cmd" ]]; then
+                nv_gpu_cmd="$nv_smi_cmd --query-gpu=name --format=csv,noheader"
+                IFS=$'\n' read -d "" -ra nv_gpus <<< "$($nv_gpu_cmd)"
+                for nv_gpu in "${nv_gpus[@]}"; do
+                    gpu="NVIDIA $nv_gpu"
```
hykilpikonna, line 2060: The `nvidia-smi` command already has "NVIDIA" in it, so we don't need to add NVIDIA here

image

Issue #1045, opened 2018-06-25
Images not displaying


filipekiss:
## Description

I've updated my config to use an image instead of the ascii art but it's not working. If I don't pass any parameters, it will use the ascii art and if I pass, for example, `--imgcat --source /path/to/img` it breaks. iTerm is ok with the images because `imgcat` works as expected. I've also tried running `neofetch` under `bash` instead of `zsh` but the result is the same.

#### Neofetch version
Neofetch 5.0.0
OS: macOS High Sierra 10.13.5 17F77 
Host: MacBookPro14,1 
Shell: zsh 5.5.1 
Terminal: iTerm2 

## Screenshot

If i run `neofetch` with no arguments:

![image](https://user-images.githubusercontent.com/48519/41862152-7a866ad6-7879-11e8-8e63-b1c6cbe611ce.png)

If I pass `--imgcat --source $HOME/neofetch.png` it hangs for a while at this screen:

![image](https://user-images.githubusercontent.com/48519/41862510-7550ec20-787a-11e8-9c1e-0687393a252d.png)

Then this flashes:
![image](https://user-images.githubusercontent.com/48519/41862210-9d01f724-7879-11e8-91a0-44ba4e40fbb3.png)
and then the screen is cleared

## Config file

https://gist.github.com/filipekiss/39c459ba4d7b9703b3a11f306a1a8533#file-config-conf

## Verbose log

https://gist.github.com/filipekiss/39c459ba4d7b9703b3a11f306a1a8533#file-neofetch-log

onedr0p:
Same issue here, I am on 5.0.1. I installed with `brew install neofetch --HEAD`

MindTooth:
Well, I don't think there is a `--imgcat` option.

Tried with `neofetch --iterm2 $HOME/neofetch.png`?

sharkytowers:
I had the same issue on arch/arcolinux.  I rolled back to the version of neofetch that is on the aur (as opposed to the one in the repo it arcolinux wants to update) and it works.  Hope that helps someone.

filipekiss:
@MindTooth yeah, the result is the same :/

sharkytowers:
@filipekiss I think I have version 3.3.0-1 which I know is old, but it is working for me.  I did a makefile after cloning whatever was on the AUR

erikdubois:
I can confirm this is a problem and narrow it down for you.

[...]

PR #1048, opened 2018-06-29
Fix cannot correctly get window size

StarryTony:
Fix a bug caused by not correctly unset IFS to read array.

## Description

Image cannot correctly display due to get_window_size() cannot get window size. Tests passed on MacOS High Sierra.

## Issues
For the issue: [Images not displaying #1045](https://github.com/dylanaraps/neofetch/issues/1045 "Images not displaying")

## TODO

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -3557,6 +3557,7 @@ get_window_size() {
     # user input so we have to use read to store the out
     # -put as a variable.
     IFS=';t' read -d t -t 0.05 -sra term_size
+    unset IFS
 
     # Split the string into height/width.
     if [[ "$image_backend" == "tycat" ]]; then


konimex:
Does this also work in bash 3.2?

Since on #1053's log there's

`/usr/local/bin/neofetch: line 3559: read: 0.05: invalid timeout specification`.

StarryTony:
@konimex Yes it works. For older bash, it’s recommended to set  timeout as 1.

konimex:
I'm actually not sure whether to merge this commit or not. Since `IFS=";t` (...)` *should* only be applied to one particular command after the variable assignment (CMIIW).

So, without unsetting `IFS`, what about setting the timeout as 1? Does it work?

StarryTony:
@konimex No, it doesn’t work without unsettling IFS (at least for MAC, not sure for other Linux versions).

konimex:
OK then. Merging this PR. Thanks!

PR #1049, opened 2018-07-02
Display total resolution when using xrandr

cstok:
## Features
Displays total display resolution when using xrandr, similar to xwininfo, xdpyinfo. Individual displays are included in par

## Issues
* Runs xrandr command an extra time

## To Do
* Only display if there are multiple monitors (count number of connected monitors?)
* Toggle to allow hiding of individual monitor details

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -2526,21 +2526,26 @@ get_resolution() {
 
         *)
             if type -p xrandr >/dev/null; then
+                resolution="$(xrandr --nograb --current |\
+                                awk -F'current' -F',' 'NR==1 {gsub(" |current","");print $2}')"
                 case "$refresh_rate" in
                     "on")
-                        resolution="$(xrandr --nograb --current |\
+                        monitors="$(xrandr --nograb --current |\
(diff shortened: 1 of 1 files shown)


dylanaraps:
If there’s one monitor the output should be the full resolution. I don’t understand this change.

cstok:
Sorry, should have included screenshots. With multiple monitors, you don't see your total display resolution.

Current build, using xrandr:
image

Current build, using xwininfo and xdpyinfo:
image

Proposed change:
image

dylanaraps:
This is intended behaviour is it not? What’s the use case in seeing the total area?

wget:
I understand this change. This could come in handy to show to others we have actually 4 1080p screens giving the total resolution.
And since neofetch's main purpose is to showcase our current config in a neat and concise way, I understand this PR.
Maybe rebasing to make only one commit could be nice @cstok :)

dylanaraps:
I won’t be merging this as it requires too much code. I’ll work on implementing this with a single xrandr call.

PR #1054, opened 2018-07-14
Add Proxmox VE

wget:
## Description

I'm just following the discussion I continued on the previous old pull request: https://github.com/dylanaraps/neofetch/pull/824

Working like expected and with the changes requested in the previous pull request.
screenshot_20180714_153727

## Features
Add Proxmox VE (tested)

## Issues
N/A

## TODO
N/A

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -826,6 +826,14 @@ get_distro() {
                     "on" | "tiny") distro="Siduction" ;;
                     *) distro="Siduction ($(lsb_release -sic))"
                 esac
+            elif type -p pveversion >/dev/null; then
+                case "$distro_shorthand" in
+                    "on" | "tiny") distro="Proxmox VE" ;;
+                    *)
+                        distro="$(pveversion)"
+                        distro="${distro#pve-manager/}"
+                        distro="Proxmox VE ${distro%/*}"
+                esac
 
             elif type -p lsb_release >/dev/null; then
                 case "$distro_shorthand" in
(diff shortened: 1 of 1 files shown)


dylanaraps:
None of the default detection methods work?

wget:
@dylanaraps If I was using the default detection, the OS is detected as a simple Debian distribution.
What do you recommend here?

ThomasLamprecht:
Orginal author of referenced pull request and Proxmox VE dev here.

> None of the default detection methods work?

No, but this *is* the default method to detect a Proxmox VE version...

Effectively we're our own distro, we can be either installed over it's own ISO image but also, for sake of user-freedom, on-top of an existing Debian system.
But, installing on top of Debian is not just installing a small or big application, like (for example) virtualbox or Libreoffice, but it's transforming the original Debian installation, you get our Kernel, QEMU, LXC, ceph, depending on current issues with Debian upstream also tar, openvswitch, .. And more important, you're then fed from *our* repositories. And, to have the pveversion binary one needs to have the pve-manager package installed, and that one depends on pretty much all of our stack - which then pulls in aforementioned transformation to the Proxmox VE distribution.

And the check is like really really cheap, and it's a good one. Somebody would need to have made a manual binary in $PATH named "pveversion", which is a pretty edge case - Proxmox VE is around for well over 10 years, and we never got reported a clash which such a thing..

So, @wget thanks for picking this up, single request I'd have is to have a:
Co-developed-by: Thomas Lamprecht <t.lamprecht@proxmox.com>

trailer in the commit message, but if not the world will move on :)

And @dylanaraps it would be great to reconsider, as said we're here since a long time, clearly our own Distro and the check is IMO good and from my POV as a long-time Proxmox VE dev it's a least a good enough way, and as neither me nor @wget did come up with a "default detection method" (which clearly weren't enough, else I'd never put in that work) I think we can really say that it's required to do this cheap "non default" check, but at the same time has no implications for non-PVE distros (which is probably the most important thing here).

dylanaraps:
Understood :+1:

PR #1058, opened 2018-07-17
Several OpenBSD related fixes

julianaito:
Hi,
(edited to reflect further changes) 

What's new in that set of commits: 

- get_disk
    -  On OpenBSD you can't use `df` against a `/dev/...` unless being root or in the 'operator' group. A separate disks array creation using `awk` has been introduced only for this OS.
   
- get_wm
    - `ps -e` for non-EWMH WM won't work in OpenBSD. So i introduced a `ps_flags` array the same way we do for disks.  
   - added fvwm support

- get_model
    - Adds vmm(4) (OpenBSD's hypervisor) support

These changes impacts all OSes, so i've cross tested on Debian GNU/Linux, and it was OK. It should be the last PR for now, since i reviewed everything else and it works.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -1136,6 +1136,7 @@ get_model() {
 
     case "$model" in
         "Standard PC"*) model="KVM/QEMU (${model})" ;;
+        "OpenBSD"*)     model="vmm (${model})" ;;
     esac
 }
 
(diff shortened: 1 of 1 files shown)


konimex:
Alright.

The `get_model` and `get_wm` changes looks OK, but I have some concerns about `get_disk`.

Should OpenBSD get its own hard-coded uname check? AFAIK, using GNU utilities is possible in OpenBSD (through ports). And for some reason, GNU `df` supports `-P -h`.

Therefore, I propose to add a case for OpenBSD `df` in `df_version` (this *will* throw an error in OpenBSD `df`, we just need the unique output the same way we check AIX and IRIX.) instead of making a whole new uname check.

PR #1070, opened 2018-08-14
Modifiable info separator

xynxynxyn:
Add ability to modify the default ":" separator.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -3876,8 +3876,18 @@ prin() {
         length="${#length}"
     fi
 
+    # Make sure $sep is unset
+    unset -v sep
+
+    # Check if $seperator is set
+    if [[ $seperator ]]; then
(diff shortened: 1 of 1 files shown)


plgruener:
May I ask what is your usecase for changing the default ":" separator?
I cannot think of any other commonly used character for that except "=", which has no advantage over the colon in this context, so this PR (which is fine per se) just seems to add another (useless) --option to the already **74** other ones.

(PS: it's spelt sep**a**rator)

konimex:
Much better.

However since this is a new feature, I'll leave it to @dylanaraps to decide if the new feature is acceptable or not.

dylanaraps:
Add some documentation and we're good to go (usage function). I've got a use case for this.

dylanaraps:
Cheers :+1:

PR #1073, opened 2018-08-15
add color-reset to ascii logo drawing

### LINKED ISSUE #1065: --logo parameter changes primary terminal font
## Description
I just noticed that running Neofetch with `--logo` parameter changes the primary color of the terminal's font. By issuing `ls` , things turn back to normal for some reason.

#### Neofetch version
5.0.0 (arch)

## Config file
Default

## Verbose log
`neofetch --logo -vv 2`
https://gist.github.com/PedroBarcha/7b84a44ec8e895b3355edc6a0e5433bf

plgruener:
reset the color after drawing the ascii logo.

should fix #1065 and similar issues. This was probably not noticed before because ascii-drawing is the first called function and many other after it do reset the color.
Note: line_break() also does not reset the color, but I think that's fine, since it does not use color.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -3416,7 +3416,7 @@ get_ascii() {
     ascii_data="${ascii_data//\$\{c6\}/$c6}"
 
     ((text_padding=ascii_length+gap))
-    printf '%b\n' "$ascii_data"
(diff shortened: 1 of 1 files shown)


konimex:
LGTM, thanks!

PR #1084, opened 2018-09-02
Updated to fix new kitty bug prior to kitty fixes

HolyStephano:
Kitty recently updated and broke just using "kitty icat <image>" however, "kitty +kitten icat <image>" still works as intended.

This commit makes the neofetch script use +kitten.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -3578,7 +3578,7 @@ get_window_size() {
     fi
 
     [[ "$image_backend" == "kitty" ]] && \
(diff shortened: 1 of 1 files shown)


MindTooth:
Does this still work for older versions?

HolyStephano:
absolutely. "kitty icat" has always been shorthand for "kitty +kitten icat" if my memory serves.

MindTooth:
Good. 👍  Just wanted to be sure.

konimex:
Merged. Thanks for the PR!

PR #1091, opened 2018-09-13
Update config comment with kitty backend.

Snuggle:
## Description

The kitty image backend was merged in #953 and v4.0.0, but the comment hasn't been updated to reflect the new backend being a valid option. This is important so that people can stumble upon this while editing their config, and so that people know it's a valid option. :slightly_smiling_face:


konimex:
Merged, thanks.

PR #1094, opened 2018-09-16
OS: Add Namib GNU/Linux

arisinfenix:
## Description
Add support for Namib GNU/Linux.
https://www.namiblinux.org/

namib


dylanaraps:
Cheers

PR #1101, opened 2018-10-09
Add default config path for Kitty

dritter:
This adds `~/.config` to the locations to search for `/kitty/kitty.conf`. This is the default path as the [documentation](https://sw.kovidgoyal.net/kitty/conf.html) states.


dylanaraps:
That's what `XDG_CONFIG_HOME` is for.

PR #1106, opened 2018-10-23
Add support for HP-UX 11i

nfinit:
## Description

Adds initial support for HP-UX 11i (11.11+) on PA-RISC systems. Also should provide basic support on Itanium systems.

This is intended to be separate from the [existing HP-UX 11.31+ branch](https://github.com/dylanaraps/neofetch/pull/625), which mostly focuses on the later Itanium systems running the latest versions of HP-UX that have different tool sets for system information gathering. _(I'm still more than happy to move my work over to that branch, I just didn't want to contaminate it with anything potentially incompatible)_

#### Modified functions

* `get_os()`
* `get_distro():` Basic identification (needs further testing with other 11i versions to ensure version output is properly trimmed)
* `get_model():` Identifies and formats machine architecture, family and model numbers on PA-RISC systems.
* `get_kernel():` Function skipped as on AIX and IRIX
* `get_uptime():` Functions properly on 11i using similar methods to AIX and IRIX, may need to be tested on 11.23 and 11.31.
* `get_cpu():` Implemented for most PA-RISC systems except the C8000, *extremely* basic support for Itanium systems.
* `get_gpu():` Should work for all VISUALIZE graphics options and likely FireGL products as well.
* `get_memory():` Partially implemented, only shows total size and not amount used.

## Issues

* HP-UX 11i on PA-RISC systems does not seem to ship with any particularly useful utilities for obtaining detailed CPU information as a non-root user, thus the current `get_cpu()` implementation is not very dynamic, instead it uses information obtained by `get_model()` to determine the CPU configuration. As almost every PA-RISC system built by HP had a single fixed CPU configuration which is not upgradeable (to my knowledge,) this is still a safe strategy, though not an ideal one. The only exception to this rule is the HP C8000, which shipped with both PA-8800 and PA-8900 processors at multiple speed grades. `get_cpu()` should currently report the processor in one of these systems as `HP PA-8x00` with no speed. Perhaps a specific workaround can be implemented for these particular systems?

## TODO
* Proper support for Itanium systems in `get_model()` and `get_cpu()`
* Testing and implementation of some functions from the existing 11.31-focused branch (mainly `get_packages()` and `get_disk()` among others)
* Testing on other systems. This code has been tested and confirmed to work only on my own C3000, and might benefit from being tested with some older workstations, headless systems and systems running HP-UX 11.00, 10.20 and 11.23.


dylanaraps:
Thanks for this!

I'll add some comments later today. Looks good overall though from first glance :+1:

nfinit:
Somehow I forgot the entire lineup of RP series servers in `get_cpu()`. It shouldn't cause any real problems (the CPU will just be reported as a generic `PA-RISC` on those systems) but I'll have to throw those in as well. It looks like the one CPU to one model rule also holds true for those with the exception of some later PA-8800 and PA-8900 systems similar to the C8000.

neofetch
```diff
@@ -1127,6 +1133,24 @@ get_model() {
             model="$(/usr/bin/uname -M)"
         ;;
 
+        "HP-UX")
+            machine_type="$(model | awk -F'/' '{print $1}')"
```
dylanaraps: What's the output of `model`? We could read it into an array and save calling it (and awk) multiple times.

Also, change all occurrences of `[ ]` to `[[ ]]`.

neofetch
```diff
@@ -1127,6 +1133,24 @@ get_model() {
...
+            if [ "$machine_type" == "9000" ]; then
+                machine_family="$(model | awk -F'/' '{print $2}')"
+                machine_model="$(model | awk -F'/' '{print $3}')"
+                if [[ "$machine_model" =~ [A-Za-z] ]]; then
```
dylanaraps: Just `[a-z]` should be fine. `nocasematch` is set globally in neofetch.

neofetch
```diff
@@ -1902,6 +1927,114 @@ get_cpu() {
...
+                    823|833) cpu="HP PA-7300LC"; speed=160 ;;
+                    839|841|849|851|859|869) cpu="HP PA-7200"; speed=120 ;;
+                    898|899) cpu="HP PA-8200"; speed=200 ;;
+                esac
+            fi
+
+            # TODO: Detailed processor configuration for Itanium systems 
+
+            # Get CPU cores.
+            cores="$(ioscan -k | grep -c "processor")"
```
dylanaraps: Add a `-F` to the `grep` call.

neofetch
```diff
@@ -2176,6 +2311,12 @@ get_gpu() {
                    awk -F':' '/device beef/ {print $2}')"
         ;;
 
+        "HP-UX")
+           [[ ! -f /opt/graphics/common/bin/graphinfo ]] && return
+           gpu=$(/opt/graphics/common/bin/graphinfo | grep 'graphics product')
```
dylanaraps: Add an `-F` to the `grep` call.

neofetch
```diff
@@ -2358,7 +2507,7 @@ get_song() {
 
     printf -v players "|%s" "${players[@]}"
     player="$(ps aux | awk -v pattern="(${players:1})" \
-                '!/ awk / && !/iTunesHelper/ && match($0,pattern){print substr($0,RSTART,RLENGTH); exit}')"
```
dylanaraps: Looks like this branch needs a rebase with master.

neofetch
nfinit: `model` output on PA-RISC systems seems to most often be a three or two-part string delimiting the system's model using the old HP 9000 naming convention, on older systems the second and third component correspond to the model and submodel (e.g. `9000/715/100`) while on newer systems it seems to correspond more to family and model (`9000/785/C3000`) hence why I'm cutting up the string differently depending on whether characters are present in the last string or not.

It's definitely consistent enough that you could just parse it into an array, I'll take a look at it.

konimex: On IA64 systems, the usual format would be `ia64 hp server <actual model>`.

neofetch
```diff
@@ -1209,8 +1233,9 @@ get_uptime() {
                     seconds="${seconds/.*}"
                 ;;
 
-                "AIX" | "IRIX")
-                    t="$(LC_ALL=POSIX ps -o etime= -p 1)"
+                "AIX" | "IRIX" | "HP-UX")
+                    [[ "$os" != "HP-UX" ]] && t="$(LC_ALL=POSIX ps -o etime= -p 1)"
+                    [[ "$os" == "HP-UX" ]] && t="$(LC_ALL=POSIX UNIX95=1 ps -o etime= -p 1)"
```
konimex: It should be safe to use `UNIX95=1` on non-HP-UX (AIX/IRIX) systems since the variable will be just ignored.

neofetch
nfinit: I think I found some useful resources with `model` strings from a number of systems:
ftp://ftp.mrynet.com/operatingsystems/HP-MPE/docs.hp.com/en/5991-6552/ch02s03.html
ftp://ftp.mrynet.com/operatingsystems/HP-MPE/docs.hp.com/en/5991-1247B_ed2/ch02s04.html
There's even some Superdome output on here, which is great. There's also a couple four-part strings by the looks of it.

neofetch
nfinit: Condensed it, thanks!

neofetch
nfinit: I'm sorry about this, but I'm not terribly experienced with git (especially working on a larger project like this) and I can't for the life of me get this branch to rebase properly. I've tried a couple different methods so far (most recently [this](https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request)) and it still doesn't seem to apply the changes from these more recent commits, though it definitely seems to go through and says it's up to date with the remote master.

Am I just missing something simple? It's driving me crazy.

Issue #1102, opened 2018-10-20
Song info is broken for some players


arisinfenix:
## Description
Commit e6b8f4f broke some players that use a query command with a format string.

e.g.: `deadbeef --nowplaying-tf '%artist% \n %album% \n %title%'`
The newline in the format string is not interpreted/printed as was intended.

I have not tested all players that use a query command with format string,
but at least deadbeef, rhythmbox and qmmp are affected.

#### Neofetch version
git

## Screenshot
deadbeef
![deadbeef](https://user-images.githubusercontent.com/40315484/47254174-aed4dc00-d45e-11e8-9cb8-e2017875e1d2.png)

rhythmbox, qmmp
![rhythmbox](https://user-images.githubusercontent.com/40315484/47254179-c318d900-d45e-11e8-8f4f-73b7ebb47245.png)

arisinfenix:
Rhythmbox can use `get_song_dbus`, see #1110

A possible solution for the other affected players would be to add back separator to the query commands and then pipe the output to awk.
```bash
        "deadbeef"*)    "$(deadbeef --nowplaying-tf '%artist%-@-%album%-@-%title%' |
                           awk -F'-@-' '{print $1"\n"$2"\n"$3}')"
        ;;
```

And perhaps add a function.
```bash
    get_song_query() {
        song="$(eval "${1}" | awk -F'-@-' '{print $1"\n"$2"\n"$3')"
    }

    case "${player/*\/}" in
        ...
        "deadbeef"*)      get_song_query "deadbeef --nowplaying-tf '%artist%-@-%album%-@-%title%'" ;;
        "qmmp"*)          get_song_query "qmmp --nowplaying '%p-@-%a-@-%t'" ;;
        and so on
        ...
    ;;
```
What do you think @dylanaraps ?

dylanaraps:
No `eval`. 

I have a laptop again so I'll try and fix this tomorrow.

dylanaraps:
Does #1151 fix the issue?

arisinfenix:
:+1:

I have not tested other players yet.

arisinfenix:
mocp is not affected, works with master branch.

dylanaraps:
**Players**

- [x] amarok
- [x] audacious
- [] banshee
- [x] bluemindo
- [x] clementine
- [x] cmus
- [] deadbeef
- [x] deepin-music
- [x] dragon
- [x] elisa
- [] exaile
- [x] gnome-music
[...]

PR #1090, opened 2018-09-10
get_song fixes

xPMo:
## Description

When get_song is called, there are spaces added between songs twice in many cases, leading to output like:

```
2 Mello  -  Lena Raine  -  Mirror Temple (Mirror Magic Mix)
```
instead of 
```
2 Mello - Lena Raine - Mirror Temple (Mirror Magic Mix)
```
I doubt this is as designed, since not all commands have the extra spaces.  If this *is* by design, ignore f86c999.

The second issue is that `/xesem:album/` will match `"xesem:albumArtist"` depending on dbus print order.  The previous output should *actually* be:
```
2 Mello - Celeste B-Sides - Mirror Temple (Mirror Magic Mix)
```
Telling awk to match `/"xesem:album"/` (with surrounding quotes) will fix this in the default dbus function.  I am unsure if other player functions have a similar /album/ bug, skimming through shows banshee and quodlibet as possibilities.

## Issues

* Large space between song/album/artist

* Neofetch displays album artist instead of album

PR #1110, opened 2018-10-27
Song fixes

arisinfenix:
## Description

Fix some issues in get_song().

## Features
- Use `get_song_dbus` for rhythmbox. Partial fix for #1102 

- <del>Fix matching in `get_song_dbus`. With rhythmbox `AlbumArtist` was printed instead of `Album`. 
The issue was already discovered by @xPMo, see #1090</del>
- include PR #1090

- Fix detection of elisa player.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -2332,7 +2332,7 @@ get_song() {
         "deadbeef"
         "deepin-music"
         "dragon"
-        "elise"
+        "elisa"
         "exaile"
         "gnome-music"
         "gmusicbrowser"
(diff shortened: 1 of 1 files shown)


xPMo:
Sorry for no response on my PR, I hadn't been able to track down one of the players to test.

There is an issue with this.  Running `trim` on $song will [remove all spaces in `$song`](https://i.imgur.com/sW57fak.png).

What I did was just remove the extra spaces  in the awk command instead (`" \n "` → `"\n"`), since they're what cause the spacing problems.

xPMo:
Okay, the `trim` problem is from `e7a3f5c`.  I don't know the purpose of `trim`. By the name, I would think it would leading/trailing spaces, but it seems to remove all?

Some clarification from @arisinfenix would be helpful.

konimex:
I think it looks good and there are no problems.

@arisinfenix: Is this cleared for merge?

PR #1114, opened 2018-11-08
Add 'mpv' new value for 'player-name'

cniw:
## Description

Add 'mpv' new value for 'player-name'
Dependency: mpv-mpris (MPRIS plugin for mpv) https://github.com/hoyon/mpv-mpris

screenshot


Magicrafter13:
Was there something wrong with this implementation? I'd actually like to see some support for mpv personally...

PR #1118, opened 2018-11-20
OS: Add Bedrock Linux

paradigm:
I'd like to add Bedrock Linux support.  I believe there were discussions about doing this in the past, but at the time Bedrock was not in a suitable position to cleanly inter-operate with neofetch.  We are now beta testing Bedrock Linux 0.7 Poki which includes changes I explicitly made with the aim of getting it to play more nicely with neofetch.  I hope to have Poki out of beta and properly released in around a month, and it'd be great if I could show it off with neofetch then.

Looks like a lot of preliminary work was done in the intervening time period:

- https://gist.github.com/dylanaraps/ec1eb7e8dc13fef54740
- https://github.com/dylanaraps/neofetch/pull/229

which makes me hopeful this will go well.

Some background, in case readers are unfamiliar with Bedrock Linux:  A Bedrock system is composed of features from other distros.  For example, I'm typing this on a system where my init comes from Void Linux, my Xorg server from Debian Stretch, my window manager from Gentoo Linux, and my web browser comes form Arch.

A couple Bedrock specific concepts that I think neofetch should handle:

- Sometimes users want Bedrock to integrate different distros together, and sometimes they want the given piece of software to be restricted to its native stratum.  Bedrock offers a `strat -r` feature which can be used to disable cross-distro integration.  When `strat -r` is not used, neofetch should report the system as Bedrock and list all of the packages from all of the strata.  When `strat -r` is used, neofetch should report the system as the given stratum's distribution and only count its packages.  For example, while `neofetch` should report the system as Bedrock, `strat -r arch neofetch` should report the system as Arch Linux.  We can detect this `strat -r` feature is set by looking for the absence of `/bedrock/cross/` directories in the `$PATH`.
- A Bedrock system can have multiple instances of the same package manager.  For example, a Bedrock system could have components from both Debian and Ubuntu, and thus have two `dpkg`'s.  Package counts should reflect this.
	- For package counts that are generated by running a command, we can loop over all of the strata and run each package manager with `strat -r`.  For example, we can run `strat -r debian dpkg-query -f '.\n' -W` and `strat -r ubuntu dpkg-query -f '.\n' -W`.
	- For package counts that are generated by counting files, we can list all instances of files on all strata by prefixing `/bedrock/strata/*/`.  For example, we can count both `/bedrock/strata/gentoo/var/db/pkg/*/*/` and `/bedrock/strata/funtoo/var/db/pkg/*/*/`

I don't want Bedrock's unusual attributes to make a mess of the very clean neofetch code base.  If there is any disagreement with how I implemented these features I'm happy to change things accordingly.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -827,7 +827,12 @@ get_distro() {
 
     case "$os" in
         "Linux" | "BSD" | "MINIX")
-            if [[ -f "/etc/redstar-release" ]]; then
+            if [[ -f "/bedrock/etc/bedrock-release" ]] && [[ $PATH == */bedrock/cross/* ]]; then
+                case "$distro_shorthand" in
+                    "on" | "tiny") distro="Bedrock Linux" ;;
+                    *) distro="$(< /bedrock/etc/bedrock-release)"
+                esac
+            elif [[ -f "/etc/redstar-release" ]]; then
                 case "$distro_shorthand" in
                     "on" | "tiny") distro="Red Star OS" ;;
                     *) distro="Red Star OS $(awk -F'[^0-9*]' '$0=$2' /etc/redstar-release)"
@@ -1268,7 +1273,21 @@ get_packages() {
     has() { type -p "$1" >/dev/null && manager="$_"; }
     dir() { ((packages+=$#)); pac "$#"; }
     pac() { (($1 > 0)) && { managers+=("$1 (${manager})"); manager_string+="${manager}, "; }; }
-    tot() { IFS=$'\n' read -d "" -ra pkgs < <("$@");((packages+="${#pkgs[@]}"));pac "${#pkgs[@]}"; }
+    if [[ -f "/bedrock/etc/bedrock-release" ]] && [[ "$PATH" == */bedrock/cross/* ]]; then
+        tot() {
+            IFS=$'\n' read -d "" -ra pkgs < <(for s in $(brl list); do strat -r "${s}" "$@"; done)
+            ((packages+="${#pkgs[@]}"))
+            pac "${#pkgs[@]}"
+        }
+        prefix="/bedrock/strata/*"
+    else
+        tot() {
+            IFS=$'\n' read -d "" -ra pkgs < <("$@")
+            ((packages+="${#pkgs[@]}"))
+            pac "${#pkgs[@]}";
+        }
+        prefix=""
+    fi
 
     case "$os" in
         "Linux" | "BSD" | "iPhone OS" | "Solaris")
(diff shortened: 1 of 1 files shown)


paradigm:
Looks like the previous commit, 0168516a2c00b115b7b0ace7e36ac7f2da6ece7e, broke the test suite by having lines that are too long.  Could probably refactor it to have some new variable that contains the result of `hw_pagesize / 1024 / 1024`.

paradigm:
The code to calculate the width of the ASCII logo included backslashes used for escaping in its calculation, which made it overestimate the width of the logo I'm proposing here.  I've added a line to squash doubled up backslashes in the calculation which should resolve the concern for specifically the Bedrock logo.  It's probably possible to make a more generalized solution for escaped characters than the hard-coded escaped-backslash one I'm proposing here.

paradigm:
Other successfully merged OS additions had screenshots, figured I should add one in case it's a requirement:

https://i.imgur.com/EZaATD5.png

paradigm:
Apologies for the push spam.  Spent some time iterating on the ASCII logo with the Bedrock community.  I think we've settled on something, at least for the time being.

Other successfully merged OS additions had screenshots, and so I figured I should add one in case it's a requirement:

Neofetch Bedrock Linux

konimex:
I think this will need approval from @dylanaraps since it changed the way get_packages() works.

LGTM, though. And good luck for the Bedrock Linux project! I'll definitely try it one day.

dylanaraps:
I don’t like the changes to get_packages(). When I can get a Linux machine
I’ll come up with another solution.

dylanaraps:
I still don’t have access to a computer but I’ll try and get access to a VM on someone else’s so I can get this merged.

paradigm:
I certainly understand being busy.  Don't feel obligated to rush in a merge, especially at the expense of code quality/style preferences.  I'm happy to show off Bedrock with neofetch whenever the opportunity manifests.

dylanaraps:
It’s less that I’m busy and more that I don’t have access to anything other
than a phone nor do I have an address where I can purchase and ship a
charger for my laptop (so that I can use it again).

It will likely be several months before I’m able to work on anything again.
I’m currently living in a village and none of the buildings on this island
have a postal address (nor a real address either).

I can get access to an old Windows laptop (For an hour at a time) but I’ll
have to setup a VM or boot a Linux distro from a USB stick.

I don’t mind working to get this PR merged it’ll just take me some time to
setup a dev environment.

dylanaraps:
I posted a reply here a few days ago but I guess it didn't post. 

Lets merge this without the `get_packages()` changes and work on a proper solution to make Bedrock work correctly.

paradigm:
The way I'm viewing this:

- If a user finds neofetch does not support Bedrock at all, there's no surprise, as its a niche distro.  Either the user will [simply use neofetch without explicit bedrock support](https://reddit.com/r/unixporn/comments/acb2wu/sway_thought_a_flatter_color_scheme_would_look/), or go and [find this PR](https://reddit.com/r/unixporn/comments/aa1q3t/cinnamon_totally_windows_10/).
- If a user finds neofetch appears to support Bedrock, but the package counts are broken, it'll look like a bug.  In that case I there's a good chance of undue new bug reports, either to you or to me.

Given this, I'm inclined to wait until neofetch fully works under Bedrock before merging in partial, broken support.  However, I could be missing some impetus here, e.g. maybe the lack of Bedrock detection is blocking something else.  I don't feel very strongly about it either way, and if you're viewing it differently I'm content to go along with your preferences here.

dylanaraps:
I won't be making a new release until Bedrock is fully supported. It's just easier for me to split this up (if possible).

paradigm:
Sounds reasonable to me.  I'll remove the packages bit from the PR after work today.

paradigm:
#1154 looks good to me!  If you merge that, feel free to close this PR.

Also, happy to hear that Bedrock may be useful to you.  I can certainly see how it would be given your work on neofetch.  If you run into issues with it, don't hesitate to reach out.

dylanaraps:
Merged :)

PR #1131, opened 2018-12-06
fix GPU driver detection bug with bumblebee-managed dual video cards

xudong-yang:
## Description

gpu_driver is enabled in my console to reassure the video card I am currently using, as my laptop is equipped with dual graphic cards. 

However, the output seems a bit curious: 

bug_screenshot

The cause is explained in the picture below: 

cause

As I have used bumblebee to manage my graphic cards, normally the nvidia card is disabled. Simply add 2 to the NR where the nvidia card appears causes this incorrect result. 

## Features

Fix the bug described above. 

## Issues

~~I am sorry I did not come up with a simple solution to keep this line less than 100 characters. The line I altered has 117 characters now.~~

~~I do have some ideas to shorten it, like `tmp_str="first half"; str="$tmp_str && the rest"`, but I think that is even worse.~~

To pass the travis check, the field name match is set as containing "nel driv" instead of exactly matching "	Kernel driver in use" (as the line length should be smaller than 102 according to the .travis.yml). 

This could be potentially problematic, but I think the keywords I choose should avoid much. Please Let me know if you you have any thoughts about this. You are also welcome to edit it directly if you like. Thanks! 

## TODO

No further work is required so far.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -3304,7 +3304,7 @@ get_gpu_driver() {
     case "$os" in
         "Linux")
             gpu_driver="$(lspci -nnk | awk -F ': ' \
-                          '/Display|3D|VGA/{nr[NR+2]}; NR in nr {printf $2 ", "}')"
+                          '/Display|3D|VGA/{nr[NR+2]}; NR in nr && $1 == "	Kernel driver in use" {printf $2 ", "}')"
             gpu_driver="${gpu_driver%, }"
 
             if [[ "$gpu_driver" == *"nvidia"* ]]; then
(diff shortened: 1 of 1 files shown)


konimex:
Since I don't know the impact of this commit to Intel+AMD GPUs and I can't test it at the moment, I'm going to wait for Dylan for this commit.

xudong-yang:
Yea it's probably better to ensure it's okay for AMD cards. Although I think it will not cause problems on AMD machines, I don't have any devices to test it. I noticed neofetch has a discord channel. Is anybody there using AMD?

dylanaraps:
I'll ask around.

hykilpikonna:

This PR is [merged into hyfetch](https://github.com/hykilpikonna/hyfetch/commit/28506740) since this repo (dylanaraps/neofetch) seems no longer maintained.

PR #1135, opened 2018-12-08
Misc: Use comma syntax in awk to print spaces

arisinfenix:
## Description
Awk has built-in functionality to separate output fields. So why not use it? :wink:


konimex:
Tested and works with non-GNU awk. Merged.
Thanks!

PR #1142, opened 2018-12-23
Store thumbnails based on image hash to avoid name collisions

SolitudeSF:
I use album covers as image source and since they all named `cover` or `folder`, current naming scheme results in image never being changed.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -3696,7 +3696,8 @@ get_image_size() {
 make_thumbnail() {
     # Name the thumbnail using variables so we can
     # use it later.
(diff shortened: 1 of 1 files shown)


dylanaraps:
This probably won't work across every OS that neofetch supports. iirc on BSD it's just called `md5`. Not sure about other OS.

This should at least fix your issue: https://github.com/dylanaraps/neofetch/commit/04420479dc452b37957afdbd936ad70a25f33645

SolitudeSF:
yep, absolutely didnt think of that

Issue #1143, opened 2018-12-23
Add Star support


ozitraveller:
https://sourceforge.net/projects/linnix/

Star is a Devuan derivative, previously based on Debian

cat /etc/os-release
PRETTY_NAME="Star GNU/Linux 2.0.1 (kirk)"
NAME="Star GNU/Linux"
VERSION_ID="2.0.1"
VERSION="2.0.1 (kirk)"
ID=star
ID_LIKE=devuan
ANSI_COLOR="1;34"
HOME_URL="https://sourceforge.net/projects/linnix/"

arisinfenix:
It is detected as Devuan because `lsb_release` has higher priority than the `os-release` file.

![devuan-star](https://user-images.githubusercontent.com/40315484/50387751-50e59e80-0704-11e9-918b-2eb8dd3d651c.png)

konimex:
#1144 merged, closing this.

PR #1144, opened 2018-12-23
Distro: Add support for Star

arisinfenix:
## Description

devuan-star2

## Issues
Detected as Devuan if `lsb-release` is installed. See #1143


konimex:
About the issue, yeah nothing we can do except expecting Star to fill `lsb_release` properly.

Anyway, merged.

PR #1160, opened 2019-01-09
Add get_editor() for printing user's installed ${VISUAL:-$EDITOR}

xPMo:
Based off of current get_shell

## Features

- `--editor_path` a la `--shell_path`
- `--editor_version` a la `--shell_version` *(needs testing)*

## TODO

- [x] Update man page
- [ ] Test more editors
- - [x] busybox vi, ed, emacs, ex, dav, mcedit, micro, moe, nano, nvim, pico, red, vi (4.0), vim
- - [x] amp (no version)
- - [ ] (?) (please test your own editor if not listed above)

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -1490,33 +1490,39 @@ get_editor() {
     esac
 
     if [[ "$editor_version" == "on" ]]; then
-        # resolve symlink (for vi case)
-        case "$(realpath "$editor_full_path")" in
-            *"/amp")
-                # no flag provided
-                editor="$editor_full_path"
-            ;;
-            *"/pico")
-                editor="$("$editor_full_path" -version 2>&1)"
-            ;;
-            *"/micro")
-                editor="$editor$("$editor_full_path" -version 2>&1)"
-                editor="${editor/$'\n'*}"
-            ;;
-            *"/busybox")
-                editor="$editor$("$editor_full_path" --help 2>&1)"
-                editor="${editor/$'\n'*}"
-                editor="${editor/ \(*}"
-            ;;
-            *"/ex" | *"/vi")
-                editor="$(printf ':version\n:q\n' | "$editor_full_path" -s)"
-                editor="${editor/*$'\n'}"
-            ;;
-            *)
-                editor="$("$editor_full_path" --version 2>&1)"
-                editor="${editor/$'\n'*}"
-            ;;
-        esac
+        while
+            case "$editor_full_path" in
+                *"/amp")
+                    # no flag provided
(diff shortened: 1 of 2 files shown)


xPMo:
I have disabled shellcheck 2153 because both $EDITOR and $editor are used.

Please let me squash and force push all my commits before accepting

arisinfenix:
Shouldn't `editor_full_path` be set using `which` or `type -p`?
There is no path when editor is set with `EDITOR=nano` for example.

 Perhaps 
 ```bash
     editor="${VISUAL:-$EDITOR}"
     editor_full_path="$(which $editor)"
 ```
- [x] mcedit
`Editor: GNU Midnight Commander 4.8.20`

dylanaraps:
This works:

```sh
type -p "${VISUAL:-$EDITOR}"
```

Note: `type` is a bash builtin. `which` is a command.

arisinfenix:
> Note: type is a bash builtin. which is a command.

Ah, good to know, thanks. :)

dylanaraps:
Too many external commands.

Instead of `head -n1` you can do:

```sh
var="${var/$'\n'*}"
```

For `tail` you should be able to do the opposite:

```sh
var="${var/*$'\n'}"
```

Also I'm unsure of the need to get the full path.

dylanaraps:
Otherwise looks good so far. :+1:

xPMo:
Okay, so this is the `vi` problem:

- If `/usr/bin/vi -> /usr/bin/busybox`, the only available version is the version of BusyBox, which can be shown with `vi --help` (or `busybox --help`).

- If `/usr.bin/vi -> /usr/bin/ex`, then there is no command line flag to show version, instead we have to get the result of the vi/ex command `:version`, so piping it into `vi -` or `ex -` is really the only option.

- If `/usr/bin/vi -> /usr/bin/n?vim`, we can use the default case.

dylanaraps:
Can this not simply be: `${VISUAL:-${EDITOR:-vi}}`?

xPMo:
Hey, happy to see you again.

> Can this not simply be: `${VISUAL:-${EDITOR:-vi}}`?

Are you speaking in general `get_editor(){ editor="${VISUAL:-"${EDITOR:-vi}"}"; }` or something more specific in my PR? I agree in general,  but I this implements an  --editor_version as well, which is what the bulk of the function does.

On the other hand, if `--editor_version` is not something which we want to maintain, it should be no problem to clean it up to just support name and `--editor_path`. (The only special case for `_path` is when `VISUAL='editor --with --flags'`)

PR #1162, opened 2019-01-09
Where supported, keep on/tiny/off flags in the same order

xPMo:
Also adds 'tiny' to manpage where flags support it.


dylanaraps:
LGTM :+1:

Issue #1170, opened 2019-01-17
Option to show memory in GB


ShayBox:
An option for the memory module that will let you show in more formats than MiB, maybe allow bits, bytes, KB, MB, GB, TB, with/without decimal place, and maybe an option for how many decimal places to show.

So instead of `3433MiB / 23974MiB` It could show `3.4GiB / 24GiB`

TriMoon:
How about using this example to make it output like you want: :wink: 
```bash
#!/usr/bin/env bash
num=3000000
divider=1024
units="GiB"
printf -v out "%'.2f %s" $(($num / $divider)).$(($num % $divider)) $units
printf "out='%s'\n" $out
```
gives
```console
out='2,929.70 GiB'
```

ShayBox:
How do I use that with the neofetch memory module though :thinking:

dylanaraps:
We can also do this to "fake" it.

```sh
    # Use $mem_label if available.
    case "${mem_label:-$memory_format}" in
        "GiB")
            mem_used="${mem_used:0:-3}.${mem_used/${mem_used:0:-2}}"
            mem_total="${mem_total:0:-3}.${mem_total/${mem_total:0:-2}}"
            mem_label="GiB"
        ;;

        "KiB")
            ((mem_used=mem_used*1024))
            ((mem_total=mem_total*1024))
            mem_label="KiB"
        ;;
    esac
```

TriMoon:
> Only issue is it throws away the value after the decimal.

Why not use `printf` like i showed in https://github.com/dylanaraps/neofetch/issues/1170#issuecomment-455576889 ?
For example using a part of the code you posted:
```shell
    case "${mem_label:-$memory_format}" in
        "GiB")
            mem_label="GiB"
            printf -v mem_used "%'.*f" \
                        ${mem_precision:-2} \
                        $(($mem_used / 1024)).$(($mem_used % 1024))
            printf -v mem_total "%'.*f" \
                        ${mem_precision:-2} \
                        $(($mem_total / 1024)).$(($mem_total % 1024))
        ;;
    esac
```

konimex:
Neofetch almost never calls printf directly. It'll be taken care of in user's config (the `print_info()` part).

For instance, neofetch (from user config) will fetch the `memory` variable [here](https://github.com/dylanaraps/neofetch/commit/e3a6b9f40d479cea510e1a2c4b7d489b41b498c5#L2357)

TriMoon:
[...]

PR #1173, opened 2019-01-19
Memory [AIX]: Detect memory based on pages (like Solaris)

konimex:
## Description

Self-explanatory

## Rationale

With #1170 in mind, I decided to change AIX's `get_memory` method a little bit. This allows us to use `MiB` universally and not having one OS stuck with `MB` (although I think AIX interprets `MiB` and `MB` as the same).

## Issues

N/A, tested on AIX 7.1

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -2307,10 +2307,20 @@ get_memory() {
             esac
         ;;
 
-        "Solaris")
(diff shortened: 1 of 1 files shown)


dylanaraps:
LGTM :+1:

PR #1182, opened 2019-01-30
Fix big Puffy ascii logo

julianaito:
## Description

A very simple fix, big Puffy has a '\\' that isn't escaped. 

## Before

image

## After

image


dylanaraps:
Thanks :+1:

PR #1192, opened 2019-02-06
image: Add framebuffer support

Crestwave:
Tested and confirmed to work with the `caca`, `chafa`, `jp2a`,  `pixterm`, and `w3m` backends, although `w3m` is the only one that actually uses the framebuffer as far as I know. The rest failing was not specific to the framebuffer or tty.


konimex:
How does one test whether this works or not? Just using a framebuffer terminal (like `yaft`) enough?

Crestwave:
I've only tested it on the Linux console, but that should probably work, too.

PR #1203, opened 2019-02-25
adding ability to query remote hosts using SSH

mateusza:
## Description

Ability to query remote hosts using SSH without installing the tool remotely.

## Features

- Automatically deploys a copy of itself on a remote server, executes it and removes.
- Passes all the arguments.
- Can be chained multiple times to jump to another hosts. (i.e. to bypass firewalls, NATs; use remote keys etc.)

## Issues

No.

## TODO

Update manual.


mateusza:
Closing PR. Will open another one after code cleanup.

PR #1204, opened 2019-02-26
SSH mode

mateusza:
## Description

Query remote host using SSH by deploying a copy of a script into temporary location and executing it.

## Features

- Pass all arguments to target system.
- Multiple jumps are allowed.
- Deploy script code on-the-fly.

## Issues

None.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -4466,6 +4466,10 @@ OTHER:
     -v                          Display error messages.
     -vv                         Display a verbose log for error reporting.
 
+SSH:
+    --ssh HOST                  Run neofetch on remote host using SSH. All arguments except
+                                --ssh are passed to remote system. Can be chained to jump.
+
 DEVELOPER:
     --gen-man                   Generate a manpage for Neofetch in your PWD. (Requires GNU help2man)
 
@@ -4652,6 +4656,12 @@ get_args() {
                 help2man -n "A fast, highly customizable system info script" \
                           -N ./neofetch -o neofetch.1
                 exit 1
+            ;;
+
+            "--ssh")
+                ssh_probe="yes"
+                [[ "$ssh_host" == "" ]] && ssh_host="$2"
+                shift
             ;;
 
             "--json")
@@ -8823,6 +8833,54 @@ EOF
     fi
 }
 
+ssh_connect(){
+    ssh_command="ssh"
+    ssh_opts="-t"
+
+    # prepare argument list for passing to remote system
+    ssh_arg=no
+    newargs=""
+    while [[ "$1" ]]
+    do
+        if [[ "$ssh_arg" == "no" ]] && [[ "$1" == "--ssh" ]]
+        then
+            shift 2
+            ssh_arg=yes
+            continue
+        fi
+        newargs="$newargs $( printf "%q" "$1" )"
+        shift
+    done
+
+    # generate random filename and payload then pipe it to ssh
+    REMOTE_SCRIPT=/tmp/neofetch-$( uuidgen ).sh
+    BOUNDARY="_THE_END_OF_FILE_"
+    (
+        # deploy script
+        printf "%s\n" "base64 -d > $REMOTE_SCRIPT <<$BOUNDARY"
+        base64 < "${BASH_SOURCE[0]}"
+        printf "%s\n" "$BOUNDARY"
+
+        # launch it with arguments
+        printf "%s %s %s\n" "bash" "$REMOTE_SCRIPT" "$newargs"
+
+        # remove the file afterwards
+        printf "%s" "rm -f $REMOTE_SCRIPT"
+
+    ) | "$ssh_command" "$ssh_opts" "$ssh_host" "sh" 2> /dev/null
+
+    ssh_err="$?"
+    if [[ "$ssh_err" -ne "0" ]]
+    then
+        printf > /dev/stderr "%s\n" \
+            "SSH connection from $HOSTNAME to $ssh_host failed with exit code: $ssh_err" 
+        exit "$ssh_err"
+    fi
+
+    exit
+
+}
+
 main() {
     cache_uname
     get_os
@@ -8832,6 +8890,9 @@ main() {
 
     get_args "$@"
     get_simple "$@"
+
+    [[ "$ssh_probe" == "yes" ]] && ssh_connect "$@"
+
     [[ "$verbose" != "on" ]] && exec 2>/dev/null
     get_distro
     get_bold
--- a/neofetch
+++ b/neofetch
@@ -8852,12 +8852,23 @@ ssh_connect(){
         shift
     done
 
-    # generate random filename and payload then pipe it to ssh
-    REMOTE_SCRIPT=/tmp/neofetch-$( uuidgen ).sh
+    # generate random filename
+    if type -p uuidgen > /dev/null; then
+        randname=$( uuidgen )
+    elif type -p xxd > /dev/null; then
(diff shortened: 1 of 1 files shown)


dylanaraps:
I don't understand the purpose of this PR. Is it not easier to just `wget`/`curl` neofetch on the remote host?

mateusza:
> I don't understand the purpose of this PR. Is it not easier to just `wget`/`curl` neofetch on the remote host?

I don't think it's easier. I have plenty of servers running different OS, some of them are cloud VMs with relatively short life-cycle. Installing neofetch on every single one would be a lot of work, especially if you only need it for a single shot. Also this feature uses a copy of local version, so you don't need to care about updates on remote systems.

Everything works with a single command:

`neofetch --ssh remote.host.xyz`

You "easier method" would be:
```
ssh remote.host.xyz
wget https://somewhere.xyz/a/b/c/neofetch
bash neofetch
rm neofetch
```

Also, if you are using jump hosts to access hosts behind NAT, you can do it with a single command as well, which is very convenient:

`neofetch --ssh jump.host.xyz --ssh db.host.xyz`

dylanaraps:
I still see this as being out of scope for neofetch. To me this feature has its place in a separate script that you'd use. You could also just `curl bla | bash` to run neofetch on the machine without saving it to disk.

neofetch
```diff
@@ -8823,6 +8833,82 @@ EOF
     fi
 }
 
+ssh_connect(){
+    ssh_command="ssh"
+    ssh_opts="-t"
```
dylanaraps: This should be `ssh_opts=(-t)` as this opens up the possibility to make the flags configurable.

neofetch
```diff
@@ -8823,6 +8833,82 @@ EOF
...
+
+    # filter out first --ssh HOST argument before passing to remote host
+    ssh_arg=no
+    newargs=""
```
dylanaraps: This isn't needed.

neofetch
```diff
@@ -8823,6 +8833,82 @@ EOF
...
+    while [[ "$1" ]]
+    do
```
dylanaraps: `do` goes on the above line.

neofetch
```diff
@@ -8823,6 +8833,82 @@ EOF
...
+        if [[ "$ssh_arg" == "no" ]] && [[ "$1" == "--ssh" ]]
+        then
```
dylanaraps: `then` goes on the above line.

neofetch
```diff
@@ -8823,6 +8833,82 @@ EOF
...
+ssh_connect(){
+    ssh_command="ssh"
+    ssh_opts="-t"
+
+    # filter out first --ssh HOST argument before passing to remote host
+    ssh_arg=no
+    newargs=""
+    while [[ "$1" ]]
+    do
+        if [[ "$ssh_arg" == "no" ]] && [[ "$1" == "--ssh" ]]
```
```sh
if [[ "$ssh_arg" == "no" && "$1" == "--ssh" ]]; then
```

neofetch
```diff
@@ -8823,6 +8833,82 @@ EOF
...
+        then
+            shift 2
+            ssh_arg=yes
+            continue
+        fi
+        newargs="$newargs $( printf "%q" "$1" )"
```
dylanaraps: This avoids a sub-shell.

```sh
printf -v esc_string '%q' "$1"
newargs+=" $esc_string"
```

neofetch
```diff
@@ -8823,6 +8833,82 @@ EOF
...
+        shift
+    done
+
+    # generate random filename (with fallbacks)
```
dylanaraps: This can replace the entire list of tests.

```sh
IFS=$'\n' read -d "" -t 1 rand < /dev/urandom
rand=${rand//[^[:alnum:]]}
rand=n${rand:0:20}
```

buffet:
Wouldn't this be better suited in an external script, that would allow you to `remote-exec script host` with arbitrary scripts?

neofetch
```diff
@@ -8823,6 +8833,82 @@ EOF
...
+    ssh_err="$?"
+    if [[ "$ssh_err" -ne "0" ]]
+    then
+        printf > /dev/stderr "%s\n" \
+            "SSH connection from $HOSTNAME to $ssh_host failed with exit code: $ssh_err" 
+        exit "$ssh_err"
+    fi
+
+    exit
+
```
dylanaraps: remove this line

neofetch
```diff
@@ -8832,6 +8918,9 @@ main() {
 
     get_args "$@"
     get_simple "$@"
+
+    [[ "$ssh_probe" == "yes" ]] && ssh_connect "$@"
```
dylanaraps: remove the gap around this.

neofetch
```diff
@@ -8823,6 +8833,82 @@ EOF
...
+        printf > /dev/stderr "%s\n" \
+            "SSH connection from $HOSTNAME to $ssh_host failed with exit code: $ssh_err" 
+        exit "$ssh_err"
+    fi
+
+    exit
+
+}
+
+
```
dylanaraps: remove this line

neofetch
dylanaraps: All variables and functions should use `_` to separate words.

neofetch
```diff
@@ -8823,6 +8833,82 @@ EOF
     fi
 }
 
+ssh_connect(){
+    ssh_command="ssh"
+    ssh_opts="-t"
+
+    # filter out first --ssh HOST argument before passing to remote host
+    ssh_arg=no
```
dylanaraps: Quote this (and others).

neofetch
```diff
@@ -8823,6 +8833,82 @@ EOF
...
+        randname=$( openssl rand -hex 20 )
+    else
+        randname="$RANDOM-$RANDOM-$RANDOM-$RANDOM-$RANDOM-$RANDOM"
+    fi
+    REMOTE_SCRIPT="/tmp/neofetch-$randname.sh"
+
+    # pipe commands to ssh
+    (
+        # workaround for old systems not providing base64 command directly
+        base64enc(){
```
dylanaraps: Why does the script need to be run through `base64` at all?

mateusza: To prevent variable substitution, but this can actually by achieved with noninterpolating heredoc.

neofetch
mateusza: You mean zeroing the variable before using it?

neofetch
mateusza: Fixed using $RANDOM and printf "%x"

neofetch
dylanaraps: I mean don't declare the variable at all before using it since its empty.

mateusza: Actually you never know if it's empty. It should be, unless somebody used the same name in other place of script that's executed before or for whatever reason it was exported from parent process.

Declaring them as local is the solution.

mateusza:
> Wouldn't this be better suited in an external script, that would allow you to `remote-exec script host` with arbitrary scripts?

Of course there are such tools already, including powerful orchestration engines like ansible, chef, puppet etc. But again: all of them need to be installed first and require more than one simple command to achieve such functionality.

buffet:
> > Wouldn't this be better suited in an external script, that would allow you to `remote-exec script host` with arbitrary scripts?
> 
> Of course there are such tools already, including powerful orchestration engines like ansible, chef, puppet etc. But again: all of them need to be installed first and require more than one simple command to achieve such functionality.

I mean the exact functionality you added here. This doesn't seem to need any installation.

dylanaraps:
Closing this, see my comments above.

PR #1220, opened 2019-03-14
Strip color codes for ascii length calc

fydo:
Ascii images with lots of color codes would result in too large of a gap between the image and information. This uses a function to strip out color codes during the routine that determines ascii_len (and ultimately, text_padding) for ascii images.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -3450,6 +3450,25 @@ image_backend() {
     [[ "$image_backend" != "off" ]] && printf '\e[%sA\e[9999999D' "${lines:-0}"
 }
 
+strip_escape_codes() {
+    local input="${1//\"/\\\"}" output="" i char within_code=0
+    for ((i=0; i < ${#input}; ++i)); do
+        char="${input:i:1}"
+        if (( within_code == 1 )); then
+            case "${char}" in
+                [a-zA-Z]) within_code=0 ;;
+            esac
+            continue
+        fi
+        if [[ "${char}" == $'\e' ]]; then
+            within_code=1
+            continue
+        fi
+        output+="${char}"
+    done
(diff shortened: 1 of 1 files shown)


dylanaraps:
Can I see what kind of ascii you're trying to use?

fydo:
Certainly! I've uploaded an example here: 

https://fydo.net/misc/fydo.asc

dylanaraps:
Neofetch doesn't support ascii files with embedded colors. Instead the colors are defined like this:

```sh
${c6}  .--.---.
 ( \\'--'/ )
( '..-...' )
 '.'-._.'.'
${c1}  <'-.,.->
     `\(  _
      <\\ / \
     __ \\|_/
    /  \\ \\
     '--'.\\ >
           \\
```

fydo:
I did see that way of defining the colors in the documentation (and being able to re-assign the color numbers at runtime is great) but it does lack the ability to provide background colors or do other fancy stuff like bold or italic escape codes. 

Please let me know if you need anything else from me for this pull request. Thank you!

konimex:
>but it does lack the ability to provide background colors or do other fancy stuff like bold or italic escape codes.

For background color, I think you can modify `color()` (or add a new function like `bcolor()` which deals exclusively with background colors, whichever you prefer) in your PR, adding your own case like `printf '\e[48;5;%bm' "$1"` and then set it in `set_colors()` like `b1="$(bcolor "$1")"` or something like that.

fydo:
I see this pull request was closed. Was this issue ever resolved? I see v6.1.0 is still exhibiting the issue where ascii color codes are causing the right info box to shift to the right (in some cases even off of the console window) due to the incorrect length calculation. Thanks in advance.

PR #1226, opened 2019-03-31
Handle GTK2_RC_FILES with multiple values

edward-p:
## Description

https://github.com/dylanaraps/neofetch/pull/185 just use `[ -f "${GTK2_RC_FILES:-$HOME/.gtkrc-2.0}" ]` as the condition which will fail when `${GTK2_RC_FILES}` have multiple values.

For me, it is `/etc/gtk-2.0/gtkrc:/home/edward/.gtkrc-2.0:/home/edward/.config/gtkrc-2.0`, which causes wrong GTK2 theme detected.

## Features
image

## Issues
Fixed GTK2 theme ditection when ${GTK2_RC_FILES} contains multiple values.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -2683,7 +2683,7 @@ get_style() {
 
         # Check for general GTK2 Theme.
         if [[ -z "$gtk2_theme" ]]; then
-            if [[ -n "${GTK2_RC_FILES}" ]]; then
+            if [[ -n "$GTK2_RC_FILES" ]]; then
                 IFS=: read -ra rc_files <<< "$GTK2_RC_FILES"
                 gtk2_theme="$(grep "^[^#]*${name}" "${rc_files[@]}")"
             elif [[ -f "${HOME}/.gtkrc-2.0"  ]]; then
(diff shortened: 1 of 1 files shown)


dylanaraps:
Would something like this work instead?:

```sh
IFS=: read -ra rc_files <<< "$GTK2_RC_FILES"
gtk2_theme="$(grep -m 1 "^[^#]*${name}" "${rc_files[@]}")"
```

Reduces the `grep` usage to a single call (`-m 1`) may need to be removed.

neofetch
```diff
@@ -2683,8 +2683,11 @@ get_style() {
 
         # Check for general GTK2 Theme.
         if [[ -z "$gtk2_theme" ]]; then
-            if [[ -f "${GTK2_RC_FILES:-${HOME}/.gtkrc-2.0}" ]]; then
-                gtk2_theme="$(grep "^[^#]*${name}" "${GTK2_RC_FILES:-${HOME}/.gtkrc-2.0}")"
+            if [[ -n "${GTK2_RC_FILES}" ]]; then
```
dylanaraps: `"$GTK2_RC_FILES"`

dylanaraps:
LGTM :+1:

PR #1239, opened 2019-04-25
support uptime from Android 9+

lightful:
## Description
Since Android 9 the uptime is returned as empty string.

This is due to new permissions restrictions (see https://www.cl.cam.ac.uk/~rja14/Papers/interrupts_pets16-1.pdf)

The proposed change fails back in these cases to the uptime command and has been tested with `/system/bin/uptime` (which supports the -s option in Android 9), the uptime busybox applet (Termux) as well as the uptime command from procps-ng.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -1211,8 +1211,14 @@ get_uptime() {
             # Get uptime in seconds.
             case "$os" in
                 "Linux" | "Windows" | "MINIX")
-                    seconds="$(< /proc/uptime)"
-                    seconds="${seconds/.*}"
+                    if [ -f /proc/uptime ]; then
+                        seconds="$(< /proc/uptime)"
+                        seconds="${seconds/.*}"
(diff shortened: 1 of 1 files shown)


dylanaraps:
Does Android have `uptime -p`?

konimex:
[Toybox](https://landley.net/toybox/help.html#uptime) has `-p` since 0.7.6, but since we're talking about Android here, some versions can stuck in Toybox 0.7.5 or earlier, which has no `-p`.

dylanaraps:
Change it to this:

```sh
if [[ -r /proc/uptime ]]; then
    seconds="$(< /proc/uptime)"
    seconds="${seconds/.*}"
else
    boot="$(date -d"$(uptime -s)" +%s)"
    now="$(date +%s)"
    seconds="$((now - boot))"
fi
```

:+1:

dylanaraps:
LGTM :+1:

PR #1251, opened 2019-05-08
 update Red Hat with new 2019 logo

infinitewarp:
## Description

Red Hat recently (May 2019) updated its official logo and brand standards, and the new logo is now a slightly different shaped fedora _without_ the shadowman figure beneath it. 

You can see the new logo here: https://www.redhat.com/en/about/brand/standards

This PR just swaps the old logo with a representation of the new logo. I used an online image-to-ASCII generator with Red Hat's new logo and then cleaned up some bits by hand.

## Features

Screen Shot 2019-05-08 at 12 47 49 PM

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -7898,24 +7898,26 @@ EOF
         ;;
 
         "Redhat"* | "Red Hat"* | "rhel"*)
-            set_colors 1 7 3
+            set_colors 1
             read -rd '' ascii_data <<'EOF'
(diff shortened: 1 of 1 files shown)


dylanaraps:
We should keep the old logo. Add it before the Red Hat section as `redhat_old` and I'll merge. :+1:

dylanaraps:
Thanks! :+1:

Will add the old logo back in another commit.

PR #1256, opened 2019-05-19
shoppe support

knuxify:
## Description

Add support for the [shoppe](https://github.com/knuxify/shoppe) package manager.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -1337,6 +1337,7 @@ get_packages() {
             has "sorcery"    && tot gaze installed
             has "alps"       && tot alps showinstalled
             has "butch"      && tot butch list
+			has "shoppe"     && tot shoppe query list +rq
 
             # Counting files/dirs.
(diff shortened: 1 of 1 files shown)


neofetch
```diff
@@ -1337,6 +1337,7 @@ get_packages() {
             has "sorcery"    && tot gaze installed
             has "alps"       && tot alps showinstalled
             has "butch"      && tot butch list
+			has "shoppe"     && tot shoppe query list +rq
```
dylanaraps: Fix the indentation. :>

dylanaraps:
This seems to be a wrapper around other package manager's right? What would be the point of adding support for it?

knuxify:
Actually, it's a package manager by itself, the wrapper is just a little addon (and will probably be separated from the main project at some point). The query command only shows packages installed with shoppe itself.

PR #1257, opened 2019-05-19
Added ApertureScience Logo

LuigiTheHunter:
## Description

## Features
- Added ApertureScience Logo

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -4850,6 +4850,8 @@ ${c1}              .,-:;//;:=,
                =++%%%%+/:-.
 
 EOF
+        ;;
+
         "AIX"*)
             set_colors 2 7
             read -rd '' ascii_data <<'EOF'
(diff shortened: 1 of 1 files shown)


LuigiTheHunter:
Hopefully you got the new commit aswell, because I forgot the ;; :)

konimex:
I see no reason to merge this unless "ApertureScience" is an actual Linux distro.

LuigiTheHunter:
> I see no reason to merge this unless "ApertureScience" is an actual Linux distro.

Hmh sad. It looks really good though. But Windows isn't a Linux distro aswell.

konimex:
I should've clarified: A distro or an actual OS. Like Solaris, AIX, etc.

konimex:
>I understand your decision just one question though, what is the downside of adding it?

It would set a dangerous precedent. If we merge this, in the future people will ask to add a Black Mesa logo or a sport team (e.g. Barcelona or Liverpool) logo because they saw this PR being merged, so we have to be fair, and if we have all non-OS related logo included, the filesize will be too big. Either we include them all, or we don't.

Since Dylan seems to endorse my decision here, I'm closing this PR.

LuigiTheHunter:
Thanks for the explanation I fully understand :)

PR #1193, opened 2019-02-07
Robuster kitty font parsing

SibrenVasse:
I've got a shortcut defined in kitty to change the font size. The relevant part of 'kitty --debug-config':
```
Changed shortcuts:
         shift+control+minus KeyAction(func='decrease_font_size', args=())
```
My neofetch output:
```
Terminal Font: Hack ', args=())
```
This commit changes the parsing to an awk pattern that only matches the start of lines in the debug output.
New output:
```
Terminal Font: Hack 8.9 
```

PR #1262, opened 2019-06-03
Fix kitty font parsing when font name has whitespaces

lebensterben:
## Description

As mentioned in #1193, the current implementation doesn't work for me. In my settings, the font is Iosevka Term SS08. And if I won't get the full font name.


dylanaraps:
The line needs to be below 100 columns before I can merge. Otherwise LGTM :+1:,

PR #1263, opened 2019-06-03
Works for multiple GPUs now

lebensterben:
Work for multiple GPUs now
   
1. Replace `\\"` with `\"`, otherwise awk warns `\"' is not a
   known regexp operator
2. Use the slot number of `lspci -mm` return value to
   determine whether two video cards are in different PCI slots,
   and therefore are two different GPUS.


dylanaraps:
LGTM :+1:

Thanks!

PR #1265, opened 2019-06-07
improve checking for Guix

Chaformbintrano:
## Description
neofetch thought it was on GuixSD when it was on a different distro with Guix as an additional manager


konimex:
I'd move the whole guix block after [this](https://github.com/Chaformbintrano/neofetch/blob/8b327e2550903737ee865d493a756ef97392d0c5/neofetch#L915) if block instead of adding another condition.

konimex:
Replaced by #1271.

PR #1269, opened 2019-06-19
Set nullglob for package counting

14mRh4X0r:
## Description

This prevents package managers showing up as having installed 1 package,
while it's actually 0.


dylanaraps:
Thanks :+1:

Issue #1267, opened 2019-06-13
GUIXSD is reported as the distribution if GUIX is installed.


dylanaraps:
The `guix` package manager can be installed on any distribution. Installing it outside of GUIXSD causes `neofetch` to incorrectly detect the distribution as GUIXSD.

dylanaraps:
Problem code:

```sh
            elif type -p guix >/dev/null; then
                case "$distro_shorthand" in
                    "on" | "tiny") distro="GuixSD" ;;
                    *) distro="GuixSD $(guix system -V | awk 'NR==1{printf $5}')"
                esac
```

konimex:
#1265 solves this, but I suggested to move the whole if block below the `/etc/os-release` instead of an additional `/etc/os-release` check so I don't merge it yet.

PR #1271, opened 2019-06-20
Moved guix check below /etc/os-release check

Renzix:
## Description

Only fill in the fields below if relevant. Basically [this PR](https://github.com/dylanaraps/neofetch/pull/1265) which fixes this [issue](https://github.com/dylanaraps/neofetch/issues/1267) done right. It could still have issues thinking a distro is guixsd if the guix package manager is installed and it is still a unknown distro


dylanaraps:
LGTM :+1:

Issue #1241, opened 2019-04-29
[Feature Request] Formatting


twodotcs:
## Description
Formatting with identical spacing between the title and the info, like `lscpu`.
Example from `lscpu`:
```
Architecture:        x86_64
CPU op-mode(s):      32-bit, 64-bit
Byte Order:          Little Endian
Address sizes:       43 bits physical, 48 bits virtual
CPU(s):              12
```
I achieved this in Python with this over complicated function:
```
  neofetch = "neofetch grabbed by subprocess or the such would go here"
  neolist1 = neofetch.split('\n')
  neolist1 = ([i.split(':', 1)[0] for i in neolist1])
  width = 20
  iterator = 2
  neolist2 = neofetch.split('\n')[1:-2]
  neolist2 = ([i.split(':', 1)[1] for i in neolist2])
  print(neolist1[0]  + '\n' +  neolist1[1]) ##first 2 lines
  while(iterator<len(neolist2)-2):
        line = neolist1[iterator] + ':'.ljust(width - len(neolist1[iterator])) + neolist2[iterator]
        print(line)
        iterator+=1
```

This is the only workaround I have currently (probably due to my awful coding skills), it would be ideal to have this built into neofetch.

konimex:
The `info` system we currently have cannot accommodate, I think. But you can set it manually.

e.g.

```sh
print_info() {
    info "Shell     " shell
    info "Resolution" resolution
    info "DE        " de
    info "WM        " wm
    info "WM Theme  " wm_theme
}
```

and it'll show up properly.

dylanaraps:
I may be able to add alignment to the current info system. Will look into it.

dylanaraps:
Duplicate of: https://github.com/dylanaraps/neofetch/issues/895

Issue #895, opened 2018-01-13
Is there padding/align option?


lapsio:
## Description

I'm not sure if there already is such option - but is there a way to align results? I tried to just insert spaces in categories names but they seem to be trimmed as only one space after name remains displayed. It'd make much more sense to align results for barinfo display mode as currently it looks really messy and kind of defeats purpose of displaying bars if they don't look clean at all...

Snuggle:
This kind of already exists, it's a hacky fix though. I had to search through Neofetch's source code. I used it to align my rainbow, since my terminal supports 16-million truecolor and I like showing it off.

Basically you just need to add `printf '\e[%bC%b' "$text_padding" "${zws}${cols}"` for aligning things.

![image](https://user-images.githubusercontent.com/26250962/45259232-9a151b00-b3bf-11e8-90ee-0a4695185802.png)
![image](https://user-images.githubusercontent.com/26250962/45259254-0f80eb80-b3c0-11e8-894f-de1cd19900d3.png)


^ Here is my config for the above. 
(Requires kitty terminal for native image support - https://github.com/kovidgoyal/kitty)
https://pastebin.com/C4LbSySJ

joelostblom:
Another workaround is to pipe the output through `column`:

```bash
neofetch --stdout | column -t -s ':'
```

gives

```
OS                  Ubuntu 18.04.2 LTS x86_64                    
Kernel              4.15.0-45-generic                               
Uptime              2d 4h 48m                                              
Packages            1823                                                   
Shell               zsh 5.4.2   
...
```

jackcogdill:
Another workaround is using `--separator '\t'`

svenknoke:
if you want color output with the piping through `column` trick you can use:
`neofetch --off --color_blocks off | column -t -s ':'`

dylanaraps:
Closing as stale.

hikkidev:
A cringe-way workaround
```bash
neofetch --ascii_distro arch_small --color_blocks off | column -ts ":" -o "" && neofetch --off --disable $(cat .config/neofetch/config.conf | grep -oP "(?<=info )\w+|(?<=.\" )\w+" | sed -e "/^cols$/d" | tr "\n" " ") | tail -n +2
```

wave864:
[...]

PR #1281, opened 2019-07-12
Alignment

zjp:
## Description

Adds the option to align the output to the space after the end of the separator string ":", "->", etc by padding all arguments to printf by the same amount regardless of length 

## Features
--align_output 'left', 'right', 'off'
left: Don't right-justify labels like "OS", but align what gets listed after
right: Align labels to make their contents aligned
off: Default behavior of neofetch

## Issues

1. Addresses #1241 and #895 
2. I'm not sure why I'm listed on those first 3 commits -- maybe I went too far back in rebasing? I definitely made some kind of error here that obfuscates the git log without meaning to, so I can save my changes, nuke the branch and try again if need be. 
3. I'm not sure what trim and strip_sequences do -- does using raw arguments to prin() pose some kind of security risk? 

## TODO
[ ] Move repetitive calculations that only need to be done once out of prin()
[ ] Figure out dynamic alignment -- right now the offset has to be set in prin() by hand according to the longest label

Basically solve for columns 4 and 6 of this table:
```
       |         |            | Subtract to|             | Subtract to |
Distro | Padding | Len(TF)    | Align "TF" |  Len(res)   |  Align "res"|
-------|----------------------|------------|-------------|-------------|
macOS  | 33      |    13      |      5     |     10      |    8        |
Gentoo | 38      |    13      |     10     |     10      |   13        |
Arch   | 41      |    13      |     13     |     10      |   16        |
```
Where the values in columns 4 and 6 correspond to the offset that will align the entire output according to the longest label, if that label is present. So if I have "Terminal Font" enabled for macOS I need to correct the offset by subtracting 5, but if I have that commented out then the next longest label is "Resolution" and to align that subtract 8. 

I did notice that subtracting column 4 from column 2 always returns 28, and 6 from 2 always returns 25, though, no matter the distro.
Screenshot 2019-07-12 at 16 46 11
Screenshot 2019-07-12 at 16 46 28
Screenshot 2019-07-12 at 16 46 49


dylanaraps:
You need to rebase this on master first. A lot of unrelated commits in this PR.

Issue #1270, opened 2019-06-20
Add SailfishOS


Michal-Szczepaniak:
You have its (outdated) logo but you don't support it huh?

                case "$distro_shorthand" in
                    "on")   distro="${NAME:-${DISTRIB_ID}} ${VERSION_ID:-${DISTRIB_RELEASE}}" ;;
                    "tiny") distro="${NAME:-${DISTRIB_ID:-${TAILS_PRODUCT_NAME}}}" ;;
                    "off")  distro="${PRETTY_NAME:-${DISTRIB_DESCRIPTION}} ${UBUNTU_CODENAME}" ;;
                esac

this (as shorthand is off by default) will put PRETTY_NAME into distro but in sfos PRETTY_NAME is PRETTY long and ID would be better.

    NAME="Sailfish OS"
    ID=sailfishos
    VERSION="3.0.3.9 (Hossa)"
    VERSION_ID=3.0.3.9
    PRETTY_NAME="Sailfish OS 3.0.3.9 (Hossa)"
    SAILFISH_BUILD=9
    SAILFISH_FLAVOUR=release
    HOME_URL="https://sailfishos.org/"

this is how os-release looks like on SailfishOS. Using ID or NAME would do. You can put VERSION as another field or something. Oh also don't forget to detect SailfishOS BEFORE detecting android. Otherwise you will detect android instead (/system partition exists)

konimex:
`PRETTY_NAME` has been standard across distros I believe. Arch used `Arch Linux`, Void used `Void Linux` (IIRC) and so on, and we can't change just because Sailfish is too long. I think Fedora and/or Ubuntu have an even longer one than Sailfish.

Michal-Szczepaniak:
I'm just throwing my proposal. You can do it however you like

konimex:
Right. I'm closing this as a non-issue then.

Michal-Szczepaniak:
So no adding another os even though you have logo for it?

konimex:
Honestly, I don't know what is the issue. You don't elaborate anything apart of "Sailfish's `PRETTY_NAME` it's too long, shorten it".

1. Neofetch does properly detect Sailfish OS as they have `/etc/os-release`.
2. We have the logo.
3. You don't give any logs so we can't look into the issue.
4. We have it listed [here](https://github.com/dylanaraps/neofetch/wiki/Operating-System-Support). Just Ctrl+F Sailfish.

[...]

PR #1286, opened 2019-07-18
Linux: Fix ASCII for Sailfish OS

Michal-Szczepaniak:
Fixed wrong offset and missing colour for SailfishOS which also closes and finishes #1270


dylanaraps:
LGTM :+1:

PR #1291, opened 2019-07-26
support other versions of ksh

MitchWeaver:
`ksh` and `mksh` are not the only versions of ksh. 

there are also `pdksh`, `oksh`, `loksh`, `ksh{95,98,?}`, etc... While their bin *may* be named `ksh`, its not guaranteed.

This glob allows for detection of any `*ksh` shell, whatever its bin name is.

before: http://ix.io/1Px8.png
after:   http://ix.io/1Px9.png


dylanaraps:
LGTM :+1:

Issue #1295, opened 2019-07-31
Images too small with Chafa backend


hpjansson:
## Description

Images sometimes do not use all the provided space when using the Chafa backend. This happens because Chafa by default preserves the aspect of the input image unless the --stretch option is passed.

I will make a pull request for this in a minute.

Issue #1292, opened 2019-07-27
Images displayed using pixterm image backend are stretched horizontally


enveeed:
## Description

`neofetch` displays images in `--pixterm` mode stretched horizontally.

This issue is probably within the way `neofetch` calls pixterm, cause `pixterm` apparently does not automatically keep aspect ratio unless specified with `-s fit` or the correct `-tc` / `-tr` numbers, so either that mode is not used or the calculation of those numbers are wrong.

#### Neofetch version
`6.0.0`

## Screenshot

![](https://s.put.re/Lje2kLTT.png)

(the [image used](https://s.put.re/JFPaC6Tm.jpg), for reference)

## Config file
(default, nothing changed at all)

## Verbose log

https://s.put.re/stQiZ5Ch.neofetchlog

hpjansson:
Does output look ok with the chafa backend? It's character-based also.

enveeed:
![](https://s.put.re/9CX8QHgG.png)

Also not right, way too small this time and square for some reason.
I thought for a second that it may be a problem in `gnome-terminal` but I tried it out in tilix and it has the exact same problem. :/

enveeed:
I just tried it out with terminus, and for some reason it *works fine* there (both chafa and pixterm).
All the 3 terminals I tested use the same font in the same size (Source Code Pro Semibold), so this can't be the problem.

I'm not sure if this is neofetch related after all, considering that it works in terminus, but not in gnome-terminal or tilix. Do you have any other idea on how I could test this?

hpjansson:
gnome-terminal and tilix both use the VTE terminal widget, so it's likely related to that, or how neofetch deals with it. Do chafa and pixterm display the image fine when not running under neofetch? In that case you're probably right in suspecting an issue with how neofetch fetches/calculates the dimensions.

dylanaraps:
Try the latest `git` master.

```sh
git clone https://github.com/dylanaraps/neofetch --depth 1
cd neofetch
./neofetch --flags go here
```

enveeed:
@hpjansson That could be it. Both chafa and pixterm display images correctly in all 3 terminals when used directly.

@dylanaraps using the latest version from git fixed it for pixterm, but the image is still too small and in the wrong aspect ratio for chafa
![](https://s.put.re/NvkhFjX4.png)

dylanaraps:
[...]

PR #1296, opened 2019-07-31
display_image: Allow Chafa backend to stretch image to desired size

hpjansson:
Fixes issue #1295.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -3864,7 +3864,7 @@ display_image() {
         ;;
 
         "chafa")
-            chafa --size="$((width / font_width))x$((height / font_height))" "$image"
+            chafa --stretch --size="$((width / font_width))x$((height / font_height))" "$image"
         ;;
 
         "jp2a")


dylanaraps:
I'd much rather aspect ratio be respected. Can I see some screenshots as a comparison?

hpjansson:
When you're scaling the image to multiples of, say, 8 pixels per character cell, the aspect will often be off by a fraction, since the scaling factor that yields a multiple of 8 for the width will not necessarily yield an exact multiple of 8 for the height. The gaps are likely down to differences in rounding between Neofetch and Chafa.

The quick fix is adding --stretch (and maybe --zoom), which makes Chafa behave exactly like Pixterm. I.e. it will produce output at the exact size specified, respecting Neofetch's aspect calculation.

What prompted this is the following comment, which has pictures:

https://github.com/dylanaraps/neofetch/issues/1292#issuecomment-516933096

I think what's happening there is that the image gets aspect corrected in make_thumbnail() and then Chafa is trying to preserve the aspect on top of that by adding borders.

dylanaraps:
:+1:

PR #1299, opened 2019-08-05
packages: Check brew on Linux

dawidd6:
## Description

Homebrew officially supports Linux since 2.0.0 release.

I used `brew --cellar` instead of a fixed directory, because one can install Homebrew on Linux in whatever directory one desires, though `/home/linuxbrew/.linuxbrew` is the blessed one.

## Features

Show brew packages count on Linux.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -1347,6 +1347,7 @@ get_packages() {
             # shellcheck disable=SC2086
             {
             shopt -s nullglob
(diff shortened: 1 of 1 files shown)


dylanaraps:
CI needs fixing, :+1: to the feature though.

dylanaraps:
:+1:

PR #1300, opened 2019-08-06
WM: Add support for Rectangle on macOS

rxhanson:
## Description

Add support for [Rectangle](https://github.com/rxhanson/Rectangle) on macOS.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -1635,7 +1635,7 @@ get_wm() {
                     "[S]pectacle"
                     "[A]methyst"
                     "[k]wm"
-                    "[c]chun[k]wm"
+                    "[c]hun[k]wm"
                     "[y]abai"
                     "[R]ectangle"
                 )
--- a/neofetch
+++ b/neofetch
@@ -1631,7 +1631,18 @@ get_wm() {
     else
         case "$os" in
             "Mac OS X")
-                ps_line="$(ps -e | grep -o '[S]pectacle\|[A]methyst\|[k]wm\|[c]hun[k]wm\|[y]abai\|[R]ectangle')"
+                wm_names=(
+                    "[S]pectacle"
+                    "[A]methyst"
+                    "[k]wm"
+                    "[c]chun[k]wm"
+                    "[y]abai"
+                    "[R]ectangle"
+                )
+
+                regex_part=$( IFS='|'; echo "${wm_names[*]}" )
+
+                ps_line=$(ps -e | grep -Eio "$regex_part")
 
                 case "$ps_line" in
                     *"chunkwm"*)   wm="chunkwm" ;;
--- a/neofetch
+++ b/neofetch
@@ -1631,7 +1631,7 @@ get_wm() {
     else
         case "$os" in
             "Mac OS X")
-                ps_line="$(ps -e | grep -o '[S]pectacle\|[A]methyst\|[k]wm\|[c]hun[k]wm\|[y]abai')"
+                ps_line="$(ps -e | grep -o '[S]pectacle\|[A]methyst\|[k]wm\|[c]hun[k]wm\|[y]abai\|[R]ectangle')"
 
                 case "$ps_line" in
                     *"chunkwm"*)   wm="chunkwm" ;;
@@ -1639,6 +1639,7 @@ get_wm() {
                     *"yabai"*)     wm="yabai" ;;
                     *"Amethyst"*)  wm="Amethyst" ;;
                     *"Spectacle"*) wm="Spectacle" ;;
+                    *"Rectangle"*) wm="Rectangle" ;;
                     *)             wm="Quartz Compositor" ;;
                 esac
             ;;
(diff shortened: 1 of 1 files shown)


nibblonian:
from the travis build config:

```sh
    # Check for lines longer than 100 chars.
    - if grep '.\{102\}' neofetch; then (exit 1); else (exit 0); fi
```

the new code pushes the line beyond the max length

rxhanson:
@nibblonian Thanks for taking a look. I'll have to rethink this and see if I can get this below the max line length.

JadeMatrix:
I didn't realize neofetch actually checked window managers on macOS; if you're going to be refactoring this section you may consider adding Moom, too.  (Not a request, just a suggestion, as this PR is about Rectangle)

neofetch
```diff
@@ -1631,14 +1631,15 @@ get_wm() {
     else
         case "$os" in
             "Mac OS X")
-                ps_line="$(ps -e | grep -o '[S]pectacle\|[A]methyst\|[k]wm\|[c]hun[k]wm\|[y]abai')"
+                ps_line="$(ps -e | grep -o '[S]pectacle\|[A]methyst\|[k]wm\|[c]hun[k]wm\|[y]abai\|[R]ectangle')"
```
nibblonian:
```suggestion
                wm_names=(
                    "[S]pectacle"
                    "[A]methyst"
                    "[k]wm"
                    "[c]chun[k]wm"
                    "[y]abai"
                    "[R]ectangle"
                )

                regex_part=$( IFS='|'; echo "${wm_names[*]}" )

                ps_line=$(ps -e | grep -Eio "$regex_part")
```

That _might_ work better and make it easier to add further wms.

nibblonian: I'm not sure what the purpose of the `[x]` pattern was since it's only a single letter in each case, but I left that alone in case there is some regex magic I am unaware of

JadeMatrix: I was wondering about that too, but I'm pretty sure it's so that `ps-e | grep …` doesn't match itself:
Screen Shot 2019-08-06 at 11 17 36 PM

nibblonian: Oh! That rings a bell!  Nice catch. I also inadvertently made it case-insensitive with my unnecessary `-i` flag.

rxhanson:
Thanks for the help on this! I think it should be good to go.

neofetch
```diff
@@ -1631,7 +1631,18 @@ get_wm() {
     else
         case "$os" in
             "Mac OS X")
-                ps_line="$(ps -e | grep -o '[S]pectacle\|[A]methyst\|[k]wm\|[c]hun[k]wm\|[y]abai\|[R]ectangle')"
+                wm_names=(
+                    "[S]pectacle"
+                    "[A]methyst"
+                    "[k]wm"
+                    "[c]chun[k]wm"
```
JadeMatrix: Looks like `[c]c` is a typo here

rxhanson: Good catch, thanks!

rxhanson:
@JadeMatrix, I think this one's all set now. Thanks!

dylanaraps:
I'm not happy with how the lines were split. The string can just be broken with `\` no?

nibblonian:
It sure could. IMO the way it is split out is cleaner and more maintainable, but I guess that decision is mostly yours.  I tend to think shell scripts with randomly broken lines `\` are pretty gross, but again it's just a personal preference.

dylanaraps:
This method includes an additional subshell. Were they identical in their overhead I'd agree with you.

rxhanson:
I'm having a little trouble figuring out the best way to do it when splitting the strings with `\`. If each string component on a newline is indented to stay consistent in the file, there's whitespace between each component when it's evaluated:
```                
wm_names="[S]pectacle\
        \|[A]methyst\
        \|[k]wm\
        \|[c]hun[k]wm\
        \|[y]abai\
        \|[R]ectangle"
```
results in

`[S]pectacle \|[A]methyst \|[k]wm \|[c]hun[k]wm \|[y]abai \|[R]ectangle`

Is there a recommended way to remove the whitespace in between each of these without using an additional subshell?

One option is 
```
    wm_names="[S]pectacle"
    wm_names+="\|[A]methyst"
    wm_names+="\|[k]wm"
    wm_names+="\|[c]hun[k]wm"
    wm_names+="\|[y]abai"
    wm_names+="\|[R]ectangle"
```

It's not very elegant, but it works.

dylanaraps:
This is one way:

```sh
ps -e | grep -o "\
[S]pectacle\|\
[A]methyst\|\
[k]wm\|\
[c]hun[k]wm\|\
[y]abai\|\
[R]ectangle\
"
```

dylanaraps:
This is IMO the best way:

```sh
ps -e | grep -o \
    -e [S]pectacle \
    -e [A]methyst \
    -e [k]wm \
    -e [c]hun[k]wm \
    -e [y]abai \
    -e [R]ectangle
```

rxhanson:
Thanks! Let me know if there's anything else that's not quite right (indentation, etc).

dylanaraps:
LGTM :+1:

PR #1305, opened 2019-08-27
Rename GuixSD to Guix System

fjallarefur:
## Description

GuixSD is known as Guix System these days¹, so I updated the script accordingly.

1. https://guix.gnu.org/download/


dylanaraps:
LGTM :+1:

PR #1310, opened 2019-09-02
add MX definition

dolphinoracle:
## Description

add MX definition, which due to changes in the way debian apt system is linked into os-release, means we want to avoid overwriting our information into that file.  also, lsb_release in debian no longer sources /etc/lsb-release.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -949,6 +949,15 @@ get_distro() {
                     [[ -f "/etc/pacbsd-release" ]] && distro="PacBSD"
                 fi
             fi
+            
+            #mx-linux
+             if [[ -f "/etc/mx-version" ]]; then
+			 case "$distro_shorthand" in
+				"on" | "tiny") distro="MX" ;;
+				*) distro="MX"
+			 esac
+			 fi
+			 #mx-linux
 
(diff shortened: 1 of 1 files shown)


dylanaraps:
I'm a little confused by this change. From what I gather in the past you used to ship your own `os-release` file but you now use Debian's?

dolphinoracle:
we've never shipped our own os-release file.  What we have done in the past
is have a custom /etc/lsb-release file, but debian's version of lsb_release
no longer sources that file.

The long and short of it is that under buster, the current neofetch
identifies MX19 as Debian 10, with exactly the same file sets that MX18
had.

dylanaraps:
You either need to ship your own `/etc/os-release` file or you need to modify your `lsb-release` package. I'm against adding more and more special cases for downstream distributions which don't make these changes. 

If the system's identification files identify it as the parent distribution then neofetch is working correctly here. You *are* technically running Debian.

Adding a `/etc/os-release` file should be stupid simple. I do it in my own Linux distribution and ship it with the base files.

dylanaraps:
Closing this due to inactivity.

PR #1319, opened 2019-09-21
Better handling of wmic output on Windows

VonJagdstrumel:
## Description

As `wmic` is a Windows command, its output contains `\r` (quite erratically). It also appends space-padding at each line. This PR should fix some faulty parsing.

Raw `wmic` looks like:

```
$ printf '%q\n' "$(wmic os get Caption)"
$'Caption                             \r\r\nMicrosoft Windows 10 Professionnel  \r\r\n\r\r'
```

With according trimming, it looks like:

```
$ printf '%q\n' "$(wmic_query os get Caption)"
Caption\ Microsoft\ Windows\ 10\ Professionnel
```

`get_de()` suffered from that and guessed `"Aero"` instead of `"Modern UI/Metro"` on Windows 10.

## Issues

Nothing from what I witnessed. Might be problematic if `wmic` outputs more than one line (header omitted).

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -4103,7 +4103,8 @@ strip_sequences() {
 }
 
 wmic_query() {
-    local wmic="$(wmic "$@")"
+    local wmic
+    wmic="$(wmic "$@")"
     wmic=${wmic//$'\r'}
     trim "$wmic"
 }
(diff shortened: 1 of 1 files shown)


dylanaraps:
I don't think this is needed a we trim white-space regardless. Thanks for the PR :+1:

PR #1321, opened 2019-09-22
Added font detection for ConEmu

VonJagdstrumel:
## Description

This PR enhances `get_term_font()` by checking if `$term` matches with `conemu-*` and sets `$term_font` accordingly. Settings scanning is done according to [ConEmu's settings search sequence](https://conemu.github.io/en/ConEmuXml.html#search-sequence)

* `$APPDATA/ConEmu.xml` file :

* ConEmu's `-LoadCfgFile` argument :

* `HKEY_CURRENT_USER\Software\ConEmu\.Vanilla\FontName` value :

## Issues

* This enhancement relies on `get_term()` which can't detect ConEmu if its [Cygwin/MSYS connector](https://conemu.github.io/en/CygwinMsysConnector.html) is not used.
* Very basic XML parsing is done on `ConEmu.xml` settings file. Font detection might silently fail if the `<value/>` node is `\n`-splitted. However, the default file structure is ok.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -3208,7 +3208,7 @@ END
             local ce_arg_idx
             local ce_conf
 
-            # Could have used `eval set -- "$ConEmuArgs"` instead for arg parsing but eval is evil...
+            # Could have used `eval set -- "$ConEmuArgs"` instead for arg parsing
             readarray -t ce_arg_list < <(xargs -n1 printf "%s\n" <<< "${ConEmuArgs-}")
 
             for ce_arg_idx in "${!ce_arg_list[@]}"; do
@@ -3237,13 +3237,14 @@ END
                 [[ ! -f "$ce_conf" ]] && continue
 
                 # Very basic XML parsing
-                term_font="$(awk '/name="FontName"/ && match($0, /data="([^"]*)"/) {print substr($0, RSTART+6, RLENGTH-7)}' "$ce_conf")"
+                term_font="$(awk '/name="FontName"/ && match($0, /data="([^"]*)"/) \
+                    {print substr($0, RSTART+6, RLENGTH-7)}' "$ce_conf")"
                 break
             done
 
-            # Null-terminated contents in /proc/registry files triggers a Bash warning. Use read instead
-            [[ -z "$term_font" ]] && \
-                read -r term_font < /proc/registry/HKEY_CURRENT_USER/Software/ConEmu/.Vanilla/FontName
+            # Null-terminated contents in /proc/registry files triggers a Bash warning
+            [[ -z "$term_font" ]] && read -r term_font < \
+                /proc/registry/HKEY_CURRENT_USER/Software/ConEmu/.Vanilla/FontName
         ;;
     esac
 }
--- a/neofetch
+++ b/neofetch
@@ -3203,7 +3203,7 @@ END
             [[ -z "$term_font" ]] && term_font="Monospace 12"
         ;;
 
-        "conemu-"*)
+        conemu-*)
             local ce_arg_list
             local ce_arg_idx
             local ce_conf
@@ -3213,37 +3213,28 @@ END
 
             for ce_arg_idx in "${!ce_arg_list[@]}"; do
                 # Search for "-LoadCfgFile" arg
-                [[ "${ce_arg_list[$ce_arg_idx]}" != -LoadCfgFile ]] && continue
-
-                # Conf path is the next arg
-                ((++ce_arg_idx))
-                ce_conf="${ce_arg_list[$ce_arg_idx]}"
-                break
+                [[ "${ce_arg_list[$ce_arg_idx]}" == -LoadCfgFile ]] && {
+                    # Conf path is the next arg
+                    ce_conf=${ce_arg_list[++ce_arg_idx]}
+                    break
+                }
             done
 
             # https://conemu.github.io/en/ConEmuXml.html#search-sequence
-            local ce_seq=(
-                "$ce_conf"
-                "$ConEmuDir\ConEmu.xml"
-                "$ConEmuDir\.ConEmu.xml"
-                "$ConEmuBaseDir\ConEmu.xml"
-                "$ConEmuBaseDir\.ConEmu.xml"
-                "$APPDATA\ConEmu.xml"
-                "$APPDATA\.ConEmu.xml"
-            )
-
-            for ce_conf in "${ce_seq[@]}"; do
+            for ce_conf in "$ce_conf" "${ConEmuDir-}\ConEmu.xml" "${ConEmuDir-}\.ConEmu.xml" \
+                           "${ConEmuBaseDir-}\ConEmu.xml" "${ConEmuBaseDir-}\.ConEmu.xml" \
+                           "$APPDATA\ConEmu.xml" "$APPDATA\.ConEmu.xml"; do
                 # Search for first conf file available
-                [[ ! -f "$ce_conf" ]] && continue
-
-                # Very basic XML parsing
-                term_font="$(awk '/name="FontName"/ && match($0, /data="([^"]*)"/) \
-                    {print substr($0, RSTART+6, RLENGTH-7)}' "$ce_conf")"
-                break
+                [[ -f "$ce_conf" ]] && {
+                    # Very basic XML parsing
+                    term_font="$(awk '/name="FontName"/ && match($0, /data="([^"]*)"/) {
+                        print substr($0, RSTART+6, RLENGTH-7)}' "$ce_conf")"
+                    break
+                }
             done
 
             # Null-terminated contents in /proc/registry files triggers a Bash warning
-            [[ -z "$term_font" ]] && read -r term_font < \
+            [[ "$term_font" ]] || read -r term_font < \
                 /proc/registry/HKEY_CURRENT_USER/Software/ConEmu/.Vanilla/FontName
         ;;
     esac
--- a/neofetch
+++ b/neofetch
@@ -3202,6 +3202,49 @@ END
             # Default fallback font hardcoded in terminal-preferences.c
             [[ -z "$term_font" ]] && term_font="Monospace 12"
         ;;
+
+        "conemu-"*)
+            local ce_arg_list
+            local ce_arg_idx
+            local ce_conf
+
+            # Could have used `eval set -- "$ConEmuArgs"` instead for arg parsing but eval is evil...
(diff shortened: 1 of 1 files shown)


neofetch
```diff
@@ -3202,6 +3202,49 @@ END
...
             [[ -z "$term_font" ]] && term_font="Monospace 12"
         ;;
+
+        "conemu-"*)
+            local ce_arg_list
+            local ce_arg_idx
+            local ce_conf
+
+            # Could have used `eval set -- "$ConEmuArgs"` instead for arg parsing but eval is evil...
+            readarray -t ce_arg_list < <(xargs -n1 printf "%s\n" <<< "${ConEmuArgs-}")
```
dylanaraps: There's is probably an even better way of doing this. What does the value of "$ConEmuArgs-" look like?

neofetch
```diff
@@ -3202,6 +3202,49 @@ END
...
+                "$ConEmuDir\.ConEmu.xml"
+                "$ConEmuBaseDir\ConEmu.xml"
+                "$ConEmuBaseDir\.ConEmu.xml"
+                "$APPDATA\ConEmu.xml"
+                "$APPDATA\.ConEmu.xml"
+            )
+
+            for ce_conf in "${ce_seq[@]}"; do
+                # Search for first conf file available
+                [[ ! -f "$ce_conf" ]] && continue
```
```sh
[[ -f "$ce_conf" ]] || continue
```

dylanaraps: OR even better:

```sh
[[ -f "$ce_conf" ]] && {
    term_font="$()"
    break
}
```

neofetch
```diff
@@ -3202,6 +3202,49 @@ END
...
+
+                # Very basic XML parsing
+                term_font="$(awk '/name="FontName"/ && match($0, /data="([^"]*)"/) {print substr($0, RSTART+6, RLENGTH-7)}' "$ce_conf")"
+                break
+            done
+
+            # Null-terminated contents in /proc/registry files triggers a Bash warning. Use read instead
+            [[ -z "$term_font" ]] && \
```
```sh
[[ "$term_font" ]] || \
```

(You may be able to squeeze this onto a single line with the above change)

neofetch
```diff
@@ -3202,6 +3202,49 @@ END
...
+            # Could have used `eval set -- "$ConEmuArgs"` instead for arg parsing but eval is evil...
+            readarray -t ce_arg_list < <(xargs -n1 printf "%s\n" <<< "${ConEmuArgs-}")
+
+            for ce_arg_idx in "${!ce_arg_list[@]}"; do
+                # Search for "-LoadCfgFile" arg
+                [[ "${ce_arg_list[$ce_arg_idx]}" != -LoadCfgFile ]] && continue
+
+                # Conf path is the next arg
+                ((++ce_arg_idx))
+                ce_conf="${ce_arg_list[$ce_arg_idx]}"
```
```sh
ce_conf=${ce_arg_list[++ce_arg_idx]}
```

Can then remove the above line too.

neofetch
```diff
@@ -3202,6 +3202,49 @@ END
...
+                break
+            done
+
+            # https://conemu.github.io/en/ConEmuXml.html#search-sequence
+            local ce_seq=(
```
dylanaraps: This could just be:

```sh
for ce_conf in str str str
               str str str; do

done
```

neofetch
VonJagdstrumel: When ConEmu is launched without argument, this var is empty so I tried this command for example :

```
"%ProgramFiles%\ConEmu\ConEmu64.exe" ^
    -NoUpdate ^
    -LoadCfgFile %APPDATA%\Conemu.xml ^
    -Single ^
    -Title "This is great!" ^
    -Dir "%ProgramFiles%" ^
    -run "{Bash::CygWin Bash}" ^
    -new_console
```

And it looks like :

```
$ printf '%s\n' "$ConEmuArgs"
-NoUpdate     -LoadCfgFile C:\Users\Steve\AppData\Roaming\Conemu.xml     -Single     -Title "This is great!"     -Dir "C:\Program Files"
```

Note that any argument after `-run` is ignored. According to [ConEmu's doc](https://conemu.github.io/en/ConEmuArgs.html) :

> This must be the last used switch (excepting -new_console and -cur_console).

neofetch
VonJagdstrumel: How could I have forgotten this as I use it like everyday at work... 🤦‍♂

neofetch
VonJagdstrumel: I just found there is a line length check in `.travis.yml`. I'll apply this change but the `read -r term_font` is still too long. I'll try a different wrapping.

neofetch
VonJagdstrumel: Uh, I didn't know you could write this. I found `${ce_arg_list[$((++ce_arg_idx))]}` too cumbersome. Nice to know, thanks !

neofetch
VonJagdstrumel: Like this ?

```bash
for ce_conf in "$ce_conf" "$ConEmuDir\ConEmu.xml" "$ConEmuDir\.ConEmu.xml"
               "$ConEmuBaseDir\ConEmu.xml" "$ConEmuBaseDir\.ConEmu.xml"
               "$APPDATA\ConEmu.xml" "$APPDATA\.ConEmu.xml"; do

```

neofetch
dylanaraps: Yeah, all arithmetic works inside `${arr[]}`!

Even stuff like this:

```sh
var=${arr[i > j ? i : j]}
```

neofetch
dylanaraps: Yeah, saves creating a local variable and accompanying array.

neofetch
dylanaraps: You can do this:

```sh
read -ra ce_arg_list <<< "$ConEmuArgs"

for ce_arg_idx in "${!ce_arg_list[@]}"; do
    # Search for "-LoadCfgFile" arg.
    [[ "${ce_arg_list[$ce_arg_idx]}" != -LoadCfgFile ]] && continue

    # Conf path is the next arg.
    ce_conf=${ce_arg_list[++ce_arg_idx]}
    break
done
```

dylanaraps: I think this is even simpler though:

```sh
[[ "$ConEmuArgs" =~ -LoadCfgFile\ +([^ ]+) ]] &&
    ce_conf=${BASH_REMATCH[1]}
```

You can see the REGEX in action here: https://regex101.com/r/IoZVQw/1

This avoids yet another array/local variable as well as the loop entirely!

dylanaraps: You could even drop the `ce_conf` variable and use `${BASH_REMATCH[1]}` directly.

neofetch
```diff
@@ -3202,6 +3202,49 @@ END
...
+                "$APPDATA\ConEmu.xml"
+                "$APPDATA\.ConEmu.xml"
+            )
+
+            for ce_conf in "${ce_seq[@]}"; do
+                # Search for first conf file available
+                [[ ! -f "$ce_conf" ]] && continue
+
+                # Very basic XML parsing
+                term_font="$(awk '/name="FontName"/ && match($0, /data="([^"]*)"/) {print substr($0, RSTART+6, RLENGTH-7)}' "$ce_conf")"
```
dylanaraps: There's probably a simpler way of parsing this too. Got an example config I can look at?

neofetch
```diff
@@ -3202,6 +3202,49 @@ END
             # Default fallback font hardcoded in terminal-preferences.c
             [[ -z "$term_font" ]] && term_font="Monospace 12"
         ;;
+
+        "conemu-"*)
```
dylanaraps: The quotes can be omitted here.

neofetch
```diff
@@ -3202,6 +3202,49 @@ END
...
+                [[ ! -f "$ce_conf" ]] && continue
+
+                # Very basic XML parsing
+                term_font="$(awk '/name="FontName"/ && match($0, /data="([^"]*)"/) {print substr($0, RSTART+6, RLENGTH-7)}' "$ce_conf")"
+                break
+            done
+
+            # Null-terminated contents in /proc/registry files triggers a Bash warning. Use read instead
+            [[ -z "$term_font" ]] && \
+                read -r term_font < /proc/registry/HKEY_CURRENT_USER/Software/ConEmu/.Vanilla/FontName
```
dylanaraps: Another option is: 

```
term_font=${term_font:-$(< /proc/registry/HKEY_CURRENT_USER/Software/Conemu/.Vanilla/FontName)}
```

The above test can be omitted entirely if this works too.

neofetch
dylanaraps: Oops, I put this in the wrong comment thread didn't I?

neofetch
VonJagdstrumel: [Yup](https://github.com/Maximus5/ConEmu/blob/master/Release/ConEmu/ConEmu_Sample.xml).

neofetch
VonJagdstrumel: I already tried `$(< file)` but it gives me this warning with `/proc/registry` files as their content is null-terminated :

```
C:\cygwin64\bin\bash: warning: command substitution: ignored null byte in input
```

```
$ hexdump -C < /proc/registry/HKEY_CURRENT_USER/Software/Conemu/.Vanilla/FontName
00000000  43 6f 75 72 69 65 72 00                           |Courier.|
```

As far as I know, either we keep the test + `read` or sanitize with `awk`, `sed`, `tr` or the like.

neofetch
VonJagdstrumel: `-LoadCfgFile "C:\Users\Steve\AppData\Roaming\Conemu with spaces.xml"` won't be parsed correctly this way.

dylanaraps:
```sh
[[ "$ConEmuArgs" =~ -LoadCfgFile\ +([^ ]+) ]] &&
    ce_conf=${BASH_REMATCH[1]//[\'\"]}
```

The regex still matches when it is quoted so we can just strip the quotes if they exist.

neofetch
dylanaraps: This is one way of doing it in `bash` (same limitations as your method).

```sh
while IFS="= />\'\"" read -ra line; do
    [[ ${line[3]} == FontName ]] &&
        echo "${line[9]}"
done < ~/sample.xml
```

dylanaraps: Done! This can even parse cases like:

```
            <value
                name='FontName'
                type="string"
                data='Consolas'/>

```

```sh
while IFS=\> read -rd \< xml_data _; do
    [[ $xml_data == *[\"\']FontName[\"\']* ]] && {
        term_font=${xml_data##*=[\"\']}
        term_font=${term_font%%[\"\']/}
        break
    }
done < "$ce_conf"
```

It's fast too.

PR #1324, opened 2019-10-03
song: add comment about cmus mpris support

arisinfenix:
## Description

As of v2.8.0 cmus supports mpris2, so `get_song_dbus` could be used to query song info.
Until v2.8.0 has made it's way into the (major) distros add a comment as reminder. Perhaps debain stable is a good indicator for the switch?  :see_no_evil:


dylanaraps:
:+1:

PR #1326, opened 2019-10-07
local_ip: add interface selection for macOS

### LINKED ISSUE #1264: [Feature Request] mac os x - local_ip Interface enX
## Description

There is currently no way to define which interfaces should be used to get the local_ip information.

I have a macbook and the network adapter interface is en4 which is not detected by neofetch.

#### Neofetch version
6.0.0

iandrewt:
## Description

Adds option `--ip_interface` and array `local_ip_interface` to allow multiple local IPs to be shown

Resolves #1264 

## Features

Select as many interfaces as you want
Interfaces with no IP get ignored and error shown when run with `-v`

## Issues

Only supports macOS for now

## TODO

Add support for other OSes

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -382,6 +382,13 @@ public_ip_host="http://ident.me"
 # Flag:    --ip_timeout
 public_ip_timeout=2
 
+# Local IP interface (macOS)
+#
+# Default: 'en0'
+# Values:  'en0', 'en1'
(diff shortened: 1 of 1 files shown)


iandrewt:
wow, I deleted the branch trying to rename it :/

I'll open a new PR with the new branch

PR #1330, opened 2019-10-11
Fix macOS resolution detection

kode54:
## Description

## Features

Detects scale factor per screen by using a little math on the physical versus visible resolution.

## Issues

Not sure if this "UI Looks like" field didn't exist on older versions of macOS. In which case, this new code will fail spectacularly.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -2615,22 +2615,20 @@ get_resolution() {
 
             else
                 resolution="$(system_profiler SPDisplaysDataType |\
-                              awk '/Resolution:/ {printf $2"x"$4" @ "$6"Hz, "}')"
+                              awk '/Resolution:/ {printf $2" "$4} ; /UI Looks like:/ {print " "$4" "$6" "$8}' |\
+                              while read phys_x phys_y res_x res_y refresh; do \
+                                  scale_factor="$(expr $phys_x / $res_x)"; \
+                                  echo -n $res_x"x"$res_y"@"$scale_factor"x @ "$refresh"Hz, "; \
+                              done)"
             fi
 
-            if [[ -e "/Library/Preferences/com.apple.windowserver.plist" ]]; then
-                scale_factor="$(PlistBuddy -c "Print DisplayAnyUserSets:0:0:Resolution" \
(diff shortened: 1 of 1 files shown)


neofetch
```diff
@@ -2615,23 +2615,21 @@ get_resolution() {
 
             else
                 resolution="$(system_profiler SPDisplaysDataType |\
-                              awk '/Resolution:/ {printf $2"x"$4" @ "$6"Hz, "}')"
+                              awk '/Resolution:/ {printf $2" "$4} ; /UI Looks like:/ {print " "$4" "$6" "$8}' |\
+                              while read phys_x phys_y res_x res_y refresh; do \
+                                  scale_factor="$(expr $phys_x / $res_x)"; \
```
dylanaraps: `((scale_factor=phys_x / phys_y))` (No need for `; \\` either.

neofetch
```diff
@@ -2615,23 +2615,21 @@ get_resolution() {
...
+                                  echo -n $res_x"x"$res_y"@"$scale_factor"x @ "$refresh"Hz, "; \
```
dylanaraps: `printf %s "${res_x}x${res_y}@${scale_factor}x @ ${refresh}Hz,"` (No `; \` needed either.)

neofetch
```diff
@@ -2615,23 +2615,21 @@ get_resolution() {
...
+                              done)"
             fi
 
-            if [[ -e "/Library/Preferences/com.apple.windowserver.plist" ]]; then
-                scale_factor="$(PlistBuddy -c "Print DisplayAnyUserSets:0:0:Resolution" \
-                                /Library/Preferences/com.apple.windowserver.plist)"
-            else
-                scale_factor=""
-            fi
+            # If scale factor is redundant info
```
dylanaraps: `# @1x Scale factor is redundant.`

JadeMatrix:
FYI `UI Looks like` does not exist on 10.12

kode54:
I can commit again, or force push. Or I can just drop this, as it's more important that it continues to support old operating system versions.

PR #1335, opened 2019-10-19
Fix a irrational respond about Kernel Version in Windows

Zhboyu-BUAA:
## Description

In Windows, a third-party bash, like bash in MSYS, is needed to run neofetch. However, in MSYS, command 'uname -r', which should return the kernel version of OS, returns the version number of MSYS runtime itself. See screenshots below.
neofetch
pacman and uname

I think this result is illogical, since Windows 10 1903 should have a kernel version number of 10.0.18362, not some version number of some other software.

## Issues

Nothing witnessed as far.


dylanaraps:
It probably makes more sense to mimic the block above it:

```sh
    # Haiku uses 'uname -v' and not - 'uname -r'.
    [[ "$os" == Haiku ]] && {
        kernel=$(uname -v)
        return
    }
```

(Add a `return` and assign to `$kernel`).

Zhboyu-BUAA:
Ah I found out that I have misread the code. If actually merged, we may get a Kernel info of something like `MSYS-NT-10.0-18362 10.0.18362` with flag `--kernel_shorthand` set to `off` - maybe harder to make sense.
However I insist my opinion that Kernel info shown in this script means Kernel of the **Operating System**, not "Kernel" of the **GNU envionment (MSYS, MinGW, etc.)**.

dylanaraps:
Thanks :+1:

PR #1338, opened 2019-10-20
Add kitty to the list of available backends

Luflosi:
## Description

`kitty` wasn't in the list of backends, so I added it. Similar to #1091.

## Issues
Should I run `help2man` to generate a new `neofetch.1` man page?

## TODO
The options are in a different order in the many places they appear in. I think we should decide on one order and then stick to it everywhere.


dylanaraps:
Thanks for the PR! :+1:

> Should I run help2man to generate a new neofetch.1 man page?

If you don't mind, I'd appreciate it. :)

Luflosi:
Running it will add a few more missing things to the manpage. Is that ok or should I make a separate PR?

dylanaraps:
That's fine. :)

dylanaraps:
Thanks :)

Luflosi:
At first I ran `help2man` directly but then I discovered, that neofetch has a command-line option for that.

dylanaraps:
Yeah, it's a simple option to make things a little easier.

Luflosi:
It was very convenient. Without it I wouldn't have known the correct command-line options to use for help2man.

Issue #1342, opened 2019-10-25
Request: Support multiple profiles in Guix


spacelike:
The current method of counting per-user packages in Guix will only include the main profile. But Guix supports the ability to use any number of separate user profiles, so it would be a good idea to sum them all up as the "guix-user" package count.

A possible fix is to replace the current command for counting user packages on line 1374:
`guix package -I`
with something like:
```bash
for i in $(guix package --list-profiles); do
    # Check that the profile is in the users home (only needed if run as root), and filter out the 'guix pull' profile
    if [[ "$i" == "${HOME}"* ]] && [[ "$i" != "${HOME}/.config/guix/current" ]]; then
        guix package --profile="$i" -I
    fi
done
```
The above prints out a list of all packages (1 per line) in all the profiles owned by the current user.
But I'm not sure what the best way is to implement this in the neofetch script and make the "tot" function total it up.

[EDIT]
Also I might as well piggy back another Guix related minor issue which is probably just due to "GuixSD" being renamed to "Guix System". So now it defaults to using the Tux ascii art instead of the guix logo. An example OS line output in neofetch looks like:
`OS: Guix System 5cc96489ae20f3141edfa4d7de08f929991ad477 x86_64`

dylanaraps:
Closing as stale.

PR #1343, opened 2019-10-26
Fix Guix ascii not shown

arisinfenix:
## Description
According to #1342 Guix was renamed from GuixSD to Guix System and the ascii logo is not shown.

Reduced the matching to "Guix"* instead of "GuixSD"* to fix this. For consistency also renamed the small logo.


dylanaraps:
Thanks :+1:

PR #1354, opened 2019-11-28
Add option to arbitrarily position color columns

zjp:
## Description

This adds a command-line and/or configuration file flag to specify an arbitrary position for the color columns. It defaults to auto, which reproduces the default behavior of neofetch. It can also be set to an integer, N, which will print the columns starting at N spaces from the left side of the terminal. This is useful for users whose information is larger than their distro ascii or associated photo, and gives them a way to increase the symmetry of the output if they choose.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -600,6 +600,19 @@ block_width=3
 # Flag:     --block_height
 block_height=1
 
+# Color Alignment
+#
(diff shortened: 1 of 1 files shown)


dylanaraps:
LGTM :+1: 

Thanks.

PR #1360, opened 2019-12-07
small name changes

bgkillas:
neofetch --ascii_distro 
couldent display these without "" which annoyed me so i fixed hopefully i dident miss any


dylanaraps:
:+1:

PR #1365, opened 2019-12-10
de_version: Add Unity

arisinfenix:
## Description

Unity is still the default Desktop in Ubuntu 16.04 LTS which has support until April 2021.


dylanaraps:
:+1:

PR #1369, opened 2019-12-12
Detect Mutter on GNOME Wayland and Openbox theme on LXQt

arisinfenix:
## Description
* Added WM detection (Mutter) on GNOME Wayland
* Added Openbox theme detection on LXQt

Also:
* Removed `GNOME Shell` from a WM case as it's renamed to `Mutter` anyway.
* Changed the `awk` command in Openbox theme detection to be more robust.
  It failed when there were other lines (e.g. comments) between the `<theme>` and `<name>` lines in the config file. That was the case in Lubuntu default config for example.
* And finally refactored the Openbox theme detection to look nicer with the LXQt addition.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -1809,19 +1809,17 @@ get_wm_theme() {
         ;;
 
         "Openbox")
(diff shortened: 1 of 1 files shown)


dylanaraps:
Nice :+1:

PR #1374, opened 2019-12-20
Fix Konsole font on Tumbleweed

arisinfenix:
## Description

* On default Tumbleweed Plasma installation there is only qdbus-qt5. Add a wrapper `_qdbus`
to call `qdbus-qt5 `if available, otherwise `qdbus`.

* In Konsole 19.12.0 `KONSOLE_PROFILE_NAME` was replaced with dbus method `profile()`.  Use the `profile()` method if profile name is empty after grepping for `KONSOLE_PROFILE_NAME`.

* If there are more than one profile file in the config directory and getting the profile name fails for some reason it is possible that we print wrong font information. To avoid that return early if getting profile name fails.

* Remove some quotes for consistency.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -3125,7 +3125,7 @@ END
                         break
                     fi
                 done
-                [[ "$profile" ]] && break
+                [[ $profile ]] && break
             done
 
             [[ ! $profile ]] && return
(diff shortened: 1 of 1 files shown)


arisinfenix:
Well, I don't like the wrapper for qdbus. I will look further into this and hopefully find a better solution.
I'm going to close this and open another PR with just the changes needed for font detection in Konsole 19.12.0 and above.

PR #1375, opened 2019-12-22
Fixes and updates for Konsole font detection

arisinfenix:
## Description

In Konsole 19.12.0 `KONSOLE_PROFILE_NAME` was removed from `environment` and replaced with dbus method `profile()`.

* Added a line to use the new `profile()` method if `$profile` is empty after grepping for `KONSOLE_PROFILE_NAME`.

* To avoid printing wrong font info  return early if getting the profile name fails. 

* Removed some quotes for consistency.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -3112,6 +3112,7 @@ END
                     if ((child == "$(qdbus "$i" "$session" processId)")); then
                         profile="$(qdbus "$i" "$session" environment |\
                                    awk -F '=' '/KONSOLE_PROFILE_NAME/ {print $2}')"
+                        [[ ! $profile ]] && profile="$(qdbus "$i" "$session" profile)"
                         break
                     fi
                 done
--- a/neofetch
+++ b/neofetch
@@ -3116,7 +3116,7 @@ END
                         break
                     fi
                 done
-                [[ "$profile" ]] && break
+                [[ $profile ]] && break
             done
 
             [[ ! $profile ]] && return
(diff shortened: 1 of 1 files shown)


neofetch
```diff
@@ -3112,17 +3112,20 @@ END
                     if ((child == "$(qdbus "$i" "$session" processId)")); then
                         profile="$(qdbus "$i" "$session" environment |\
                                    awk -F '=' '/KONSOLE_PROFILE_NAME/ {print $2}')"
+                        [[ ! $profile ]] && profile="$(qdbus "$i" "$session" profile)"
```
dylanaraps: `[[ $profile ]] ||`

neofetch
```diff
@@ -3112,17 +3112,20 @@ END
...
                         break
                     fi
                 done
-                [[ "$profile" ]] && break
+                [[ $profile ]] && break
             done
 
+            [[ ! $profile ]] && return
```
```
[[ $profile ]] ||
```

arisinfenix:
Added a patch that solves the broken font detection on openSUSE.

It failed because on openSUSE (and maybe others too?) `qdbus` is not in `$PATH`. There is only `qdbus-qt5` whch is a symlink to the `qdbus` binary  in a directory outside of `$PATH`.

We can use `qtpaths --binaries-dir` and add the result to `$PATH` on success.

dylanaraps:
:+1:

PR #1377, opened 2019-12-23
added redox for some reason

bgkillas:
i was board so i ported over this from something even tho redox doesent have bash it uses dash i think and it doesent use those {c1} thingys but those colors looked worse so maybe this doesent comply to your contrib rules but eh it works

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -9658,22 +9658,22 @@ ssssssssssssso/-`      `-/osssssssssssss
 EOF
         ;;
         "Redox"*)
-            set_colors 7 7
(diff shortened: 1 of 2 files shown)


bgkillas:
image

bgkillas:
i got board and fixed it

bgkillas:
image

bgkillas:
hm conflicting files instead of doing this properly ima just redo the files

bgkillas:
welp idk why its saying my conflicting files are bad what ever

PR #1382, opened 2019-12-29
song: print empty tags correctly

arisinfenix:
## Description
Song info was incorrect whith empty tags.

For example empty 'album' tag.
`some artist - some title - Unknown Song`

 But it should be
`some artist - Unkown Album - some title`

## Features
Tested with all `non get_song_dbus players` except `muine`. Perhaps we should remove `muine`
entirely, it is a dead project and I was not able to install it for testing.

Tested with some `get_song_dbus players`, but should work with all.

I did not change the Apple players (spotify on MacOS and itues) because I cannot test.

Simplified the `cmus` query.


dylanaraps:
:+1:

PR #1388, opened 2020-01-19
Show memory in GiB and KiB

vankessel:
## Description

This PR implements the ability to display memory in GiB or KiB instead of MiB. The default is to use MiB, to preserve current behaviour. The precision for both used and total RAM can be set. These precisions both default to 1.

~I chose to use "gb" instead of "gib" in the option names because GiB just looks weird in snake case, and adding a single character will also require a whole bunch of whitespace realignment in the code around line 4673.~

I also noticed a single character whitespace misalignment and took the liberty of cleaning that up in this PR too.

## Features

Memory can be shown in GiB or KiB using ~`memory_use_gb="on"`~ `memory_unit=<kib|mib|gib>`
Precision for both used and total GiB can be set separately using `memory_used_gib_prec=<int>` and `memory_used_gib_prec=<int>`.

## Issues

Closes https://github.com/dylanaraps/neofetch/issues/1170

## TODO

N/A

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -4684,7 +4684,7 @@ TEXT FORMATTING:
 
 COLOR BLOCKS:
     --color_blocks on/off       Enable/Disable the color blocks
-    --col_offset auto/num      Left-padding of color blocks
+    --col_offset auto/num       Left-padding of color blocks
     --block_width num           Width of color blocks in spaces
     --block_height num          Height of color blocks in lines
     --block_range num num       Range of colors to print as blocks
--- a/neofetch
+++ b/neofetch
@@ -171,23 +171,23 @@ memory_unit="mib"
 #
 # Default: '1'
 # Values:  'int'
-# Flag:    --memory_used_gb_prec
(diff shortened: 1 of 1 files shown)


dylanaraps:
I would much rather a `--memory_unit` option which takes `mib`, `gib` and `kib`. :+1:

vankessel:
@dylanaraps Good call, that's much better. Here are the changes, let me know what you think.

bgkillas:
oooh i really like this small part of me wants bytes also but no real point

vankessel:
If you have less than a kilobyte of ram you probably have more important things to be doing with that system than using neofetch :stuck_out_tongue:

vankessel:
lmao who needs human readability. Unless you have less than a kilobyte of ram but then I suspect you should be doing other things with that system than using neofetch :stuck_out_tongue:

vankessel:
How's it look? @dylanaraps

dylanaraps:
I merged a slightly different implementation of this feature. Thanks!

PR #1397, opened 2020-01-28
Add info longtitle (FQDN)

xTrekStorex:
Add function get_longtitle() for info longtitle to display FQDN instead of short hostname.

## TODO
Maybe add config option for get_title() instead?


dylanaraps:
Done, thanks!

PR #1404, opened 2020-02-06
Regolith linux support

lawzava:
## Description

Support for https://regolith-linux.org

## Features
Added OS parsing, name and logo

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -973,15 +973,6 @@ get_distro() {
                 esac
                 distro=$(lsb_release "$lsb_flags")
 
-                # Temporary support for Regolith Linux
-                if [[ $DESKTOP_SESSION == "regolith" ]]; then
-                    case $distro_shorthand in
-                        on)   distro+="-REGOLITH" ;;
-                        tiny) distro+="-REGOLITH" ;;
-                        *)    distro=$(lsb_release -si)"-REGOLITH "$(lsb_release -sr) ;;
-                    esac
-                fi
-
             elif [[ -f /etc/os-release || \
                     -f /usr/lib/os-release || \
                     -f /etc/openwrt_release || \
@@ -8880,7 +8871,7 @@ d/hhhhhhhhhhhh${c3}`-/osyso+-`${c1}hhhhhhhhhhhh.h
 EOF
         ;;
 
-        "Ubuntu-REGOLITH"*)
+        "Regolith"*)
             set_colors 1
             read -rd '' ascii_data <<'EOF'
 ${c1}
--- a/neofetch
+++ b/neofetch
@@ -965,6 +965,12 @@ get_distro() {
                         distro="Proxmox VE ${distro%/*}"
                 esac
 
+            elif [[ $DESKTOP_SESSION == regolith ]]; then
+                case $distro_shorthand in
+                    on|tiny) distro="Regolith";;
+                    *) distro="Regolith ($(lsb_release -sic))"
+                esac
+
             elif type -p lsb_release >/dev/null; then
                 case $distro_shorthand in
                     on)   lsb_flags=-si ;;
@@ -8864,6 +8870,31 @@ d/hhhhhhhhhhhh${c3}`-/osyso+-`${c1}hhhhhhhhhhhh.h
       h+.hhhhhh${c6}+sssssss+${c1}hhhhhh`/yd
         dho:.hhh${c6}.:+++/.${c1}hhh`-+yd
             ddhso+++++osyhd
+EOF
+        ;;
+
+        "Regolith"*)
+            set_colors 1
+            read -rd '' ascii_data <<'EOF'
+${c1}
+                 ``....```              
+            `.:/++++++/::-.`            
+          -/+++++++:.`                  
+        -++++++++:`                     
+      `/++++++++-                       
+     `/++++++++.                    -/+/
+     /++++++++/             ``   .:+++:.
+    -+++++++++/          ./++++:+++/-`  
+    :+++++++++/         `+++++++/-`     
+    :++++++++++`      .-/+++++++`       
+   `:++++++++++/``.-/++++:-:::-`      ` 
+ `:+++++++++++++++++/:.`            ./` 
+:++/-:+++++++++/:-..              -/+.  
++++++++++/::-...:/+++/-..````..-/+++.   
+`......``.::/+++++++++++++++++++++/.    
+         -/+++++++++++++++++++++/.      
+           .:/+++++++++++++++/-`        
+              `.-:://////:-. 
 EOF
         ;;
 
--- a/neofetch
+++ b/neofetch
@@ -973,6 +973,15 @@ get_distro() {
                 esac
                 distro=$(lsb_release "$lsb_flags")
 
+                # Temporary support for Regolith Linux
+                if [[ $DESKTOP_SESSION == "regolith" ]]; then
+                    case $distro_shorthand in
+                        on)   distro+="-REGOLITH" ;;
+                        tiny) distro+="-REGOLITH" ;;
+                        *)    distro=$(lsb_release -si)"-REGOLITH "$(lsb_release -sr) ;;
+                    esac
+                fi
+
             elif [[ -f /etc/os-release || \
                     -f /usr/lib/os-release || \
                     -f /etc/openwrt_release || \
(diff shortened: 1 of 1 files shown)


neofetch
```diff
@@ -965,6 +965,12 @@ get_distro() {
                         distro="Proxmox VE ${distro%/*}"
                 esac
 
+            elif [[ $DESKTOP_SESSION == regolith ]]; then
+                case $distro_shorthand in
+                    on|tiny) distro="Regolith";;
+                    *) distro="Regolith ($(lsb_release -sic))"
+                esac
+
```
dylanaraps: There's no `os-release`?

dylanaraps:
It's not even a distribution according to the website?

> Regolith is a modern desktop environment ...

neofetch
lawzava: There is. But the distro is not seen from that file.

lawzava: i.e. not always seen (e.g. when installing on top of ubuntu)

lawzava:
Yes. Indeed. :thinking:

lawzava:
Then I guess only `DE: ` should be adjusted? :thinking:

dylanaraps:
Neofetch should already detect it unless `XDG_CURRENT_DESKTOP` reports GNOME (as is seen in Neofetch on the website)...

```sh
            if [[ $XDG_CURRENT_DESKTOP ]]; then
                de=${XDG_CURRENT_DESKTOP/X\-}
                de=${de/Budgie:GNOME/Budgie}
                de=${de/:Unity7:ubuntu}

            elif [[ $DESKTOP_SESSION ]]; then
                de=${DESKTOP_SESSION##*/}

            elif [[ $GNOME_DESKTOP_SESSION_ID ]]; then
                de=GNOME

            elif [[ $MATE_DESKTOP_SESSION_ID ]]; then
                de=MATE

            elif [[ $TDE_FULL_SESSION ]]; then
                de=Trinity
            fi
```

lawzava:
Yes, it reports GNOME :disappointed:

lawzava:
Hm.. How to handle such case? :thinking:

dylanaraps:
Regolith would also fail the `xdg-utils` detection for Deskstop Environment. I really don't think this is a Neofetch issue. The `$DESKTOP_SESSION` environment variable is set by the Display Manager / Login Manager and not the Desktop Environment which is why it is a fallback to `$XDG_CURRENT_DESKTOP`.

- https://cgit.freedesktop.org/xdg/xdg-utils/tree/scripts/xdg-utils-common.in#n270

There are a couple of options for Regolith:

- Set `XDG_CURRENT_DESKTOP` on launch. 
- Provide `$REGOLITH_DESKTOP_SESSION_ID` or another environment variable unique to Regolith.

As it stands right now, Regolith doesn't identify itself in any way as a Desktop. This is why Neofetch is correct here in Displaying GNOME as the DE and i3 as the WM.

lawzava:
Understood. Thank you for the input. Let me retract the PR for now.

dylanaraps:
I've added temporary support through a special case for now. :+1:

https://github.com/dylanaraps/neofetch/commit/eabdbcdb2bf2713e660a2e3f6c6289385d94e154

lawzava:
Nice! Thank you!.

Could I add somehow ASCII art to that? 

 e.g. Through `Ubuntu-REGOLITH` as in `Ubuntu-GNOME`?

dylanaraps:
Sure. :+1:

lawzava:
Would it be ok to add it here? https://github.com/dylanaraps/neofetch/commit/eabdbcdb2bf2713e660a2e3f6c6289385d94e154#diff-e863270127ca6116fd30e708cdc582fcR1635
`distro+="-REGOLITH"` ? @dylanaraps

lawzava:
Ignore. It doesn't export distro :man_facepalming:

neofetch
```diff
@@ -973,6 +973,15 @@ get_distro() {
...
                 distro=$(lsb_release "$lsb_flags")
 
+                # Temporary support for Regolith Linux
+                if [[ $DESKTOP_SESSION == "regolith" ]]; then
+                    case $distro_shorthand in
+                        on)   distro+="-REGOLITH" ;;
+                        tiny) distro+="-REGOLITH" ;;
+                        *)    distro=$(lsb_release -si)"-REGOLITH "$(lsb_release -sr) ;;
+                    esac
+                fi
```
lawzava: Would this workaround be ok? @dylanaraps

neofetch
```diff
@@ -973,6 +973,15 @@ get_distro() {
...
+
```
dylanaraps: This should be removed and only the ASCII logo should be kept. Regolith is only a DE right? Even this term is pushing it as it seems to be a configuration on top of regular GNOME and i3(?)

I'm happy to include the logo as well as temporary support for DE detection but I'm unconvinced as to adding it as a Linux distribution. Typically when none of the existing detection methods work for a distribution there are two reasons.

1) The distribution is based on another distribution but is only a light "skin" on top of it. No additional repositories are added, no identification files are updated and it is in reality what it is based on.
2) The distribution is unique in that it doesn't follow FreeDesktop/Linux Standards Base conventions for Linux distribution identification. 

I'm reluctant to add special cases for every single distribution which falls under (1) as the list would be endless, hard-coded and require maintenance. That being said, exceptions are made for (1) within reason. 

For Regolith to be considered a distribution by the existing detection methods a simple `/etc/os-release` file must be added to the distribution. It's that simple. The other alternative is to provide the `lsb-release` program with a Regolith configuration file. Also see: https://www.linux.org/docs/man5/os-release.html

If this isn't possible for Regolith, it points back to (1).

Here's an example (`/etc/os-release`):

```
NAME="KISS Linux"
PRETTY_NAME="KISS Linux"
ID=kiss
BUILD_ID=rolling
HOME_URL="https://kisslinux.github.io"
DOCUMENTATION_URL="https://kisslinux.github.io"
SUPPORT_URL="https://github.com/kisslinux/packages/issues"
BUG_REPORT_URL="https://github.com/kisslinux/packages/issues"
```

lawzava: Yes, that's ok, I completely agree with you. I just failed to find another way for the logo to show up if only `de` is specified :thinking:

dylanaraps: The DE specific logos aren't automatically used. They're enabled manually by the user through their configuration file or at runtime using `--ascii_distro name_here`. 

This is due to the fact that there is no way to distinguish the DE as a Linux distribution. Again related to (1) in my above post. Ubuntu-GNOME is no different to Ubuntu with GNOME installed to give an example. 

The Ubuntu-GNOME project (and other Ubuntu "flavors") make no effort to identify as anything other than Ubuntu... as they *are* Ubuntu. Neofetch *is* right here and the issue is external.

(Simply adding an `os-release` file is all it takes for detection to work.)

lawzava: Understood. That makes sense. :+1:

lawzava: Fixed

kgilmer: Hi @dylanaraps and @lawzava , thanks for taking the time to look into adding support for Regolith :smile: .  As you've discovered, Regolith makes no effort to distinguish itself from Ubuntu, which is intentional due to a Regolith design goal of being "brandless" at runtime.  I would classify Regolith as an unofficial, unsponsored, and unrecognized Ubuntu "respin".  If someone has the time and expertise to port other distributions I would support that but cannot do so given my limited time budget.  It's Ubuntu-only due to circumstance, not design.  However, it is a distinct desktop environment, based on gnome-flashback and i3.  It is neither easily containable by either "i3" or "gnome-flashback" but a hybrid of both.

I confess to not knowing much about the taxonomy of desktop environments, but if 'gnome' is the container for all gnome-based desktop environments, that seems like we're at the end of it.  Although to me, such a generalized classification isn't very useful.  gnome-flashback is very different in scope and capability than gnome-shell, not to mention MATE and other variants.

As the project lead, ideally I would like Regolith to be classified as a distinct DE (assuming that level of granularity is supported in software), but not as a distinct distribution.  (Regolith's distribution as ISO is intended mainly as a convenience to users that may be starting from scratch.)  I am happy to work with @lawzava to export the proper environment variables such that `neofetch` can consistently recognize it as such without any need for exceptional code paths.  From reviewing the dialog, it would seem this is the least-risk option as it's entirely new and is unlikely to cause regressions elsewhere:
> Provide $REGOLITH_DESKTOP_SESSION_ID or another environment variable unique to Regolith.

@dylanaraps is this the best way of identifying Regolith as a distinct DE or would you recommend something else?

dylanaraps: Thanks for getting in touch. That would be the best way, yes. :+1:

PR #1406, opened 2020-02-07
Added initial t2 Linux support

rxrbln:
I noticed our #t2sde, formerly knows as ROCK Linux was missing, so I added it ;-)

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -997,6 +997,12 @@ get_distro() {
                     *) distro="GoboLinux $(< /etc/GoboLinuxVersion)"
                 esac
 
+            elif [[ -f /etc/SDE-VERSION ]]; then
+                distro="$(< /etc/SDE-VERSION)"
+                case $distro_shorthand in
+                    on|tiny) distro="${distro% *}" ;;
+                esac
+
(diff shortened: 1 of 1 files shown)


neofetch
```diff
@@ -997,6 +997,12 @@ get_distro() {
                     *) distro="GoboLinux $(< /etc/GoboLinuxVersion)"
                 esac
 
+            elif [[ -f /etc/SDE-VERSION ]]; then
+                distro="$(< /etc/SDE-VERSION)"
+                case $distro_shorthand in
+                    on|tiny) distro="${distro% *}" ;;
+                esac
+
```
dylanaraps: There's no `/etc/os-release` file or `lsb-release` program available? None of the other detection methods work?

rxrbln: No, it is not based on another distribution, it did displayed plain Linux. I also added support for the mine packager, and a simple logo.

dylanaraps: Gotcha. Thanks for clarifying. :+1:

dylanaraps:
Thanks :+1:

PR #1417, opened 2020-03-01
Add Armbian Linux

igorpecovnik:
## Description

Display Armbian instead of Debian / Ubuntu

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -754,7 +754,7 @@ image_source="auto"
 # Values:  'auto', 'distro_name'
 # Flag:    --ascii_distro
 # NOTE: AIX, Alpine, Anarchy, Android, Antergos, antiX, AOSC,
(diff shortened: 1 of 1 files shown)


neofetch
```diff
@@ -950,6 +950,10 @@ get_distro() {
                     *) distro="Red Star OS $(awk -F'[^0-9*]' '$0=$2' /etc/redstar-release)"
                 esac
 
+            elif [[ -f /etc/armbian-release ]]; then
+                . /etc/armbian-release
+                distro="Armbian $DISTRIBUTION_CODENAME ($VERSION)"
+
```
dylanaraps: None of the existing detection methods work?

dylanaraps: `cat /etc/os-release` or `lsb-release -sir`?

igorpecovnik: For compatibility reasons /etc/os-release remains Debian or Ubuntu.

PR #1439, opened 2020-04-10
Add openmamba logo

arisinfenix:
## Description
Adds ascii logo for openmamba. https://openmamba.org
They use white/green for the logo in the boot screen. ;)

Also updated the man page.

mamba


arisinfenix:
Hm, seems there is something wrong with travis / shellcheck (download URL?).
No shellcheck issues running it locally on my machine.

PR #1442, opened 2020-04-16
neofetch: add support for OSH

Crestwave:
Not sure if it should use the current `$OIL_VERSION` if it exists and if `$BASH_VERSION` should only be set if the `$SHELL` is `bash`.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -34,6 +34,7 @@ version=7.0.0
 # but do not set the 'BASH_' shell variables (osh).
 bash_version=${BASH_VERSINFO[0]:-5}
 shopt -s eval_unsafe_arith &>/dev/null
+
 sys_locale=${LANG:-C}
 XDG_CONFIG_HOME=${XDG_CONFIG_HOME:-${HOME}/.config}
 PATH=$PATH:/usr/xpg4/bin:/usr/sbin:/sbin:/usr/etc:/usr/libexec
@@ -1589,6 +1590,7 @@ get_shell() {
         bash)
             [[ $BASH_VERSION ]] ||
                 BASH_VERSION=$("$SHELL" -c "printf %s \"\$BASH_VERSION\"")
+
             shell+=${BASH_VERSION/-*}
         ;;
 
--- a/neofetch
+++ b/neofetch
@@ -30,7 +30,14 @@
 
 version=7.0.0
 
-bash_version=${BASH_VERSION/.*}
+if [[ "$BASH_VERSION" ]]; then
+    bash_version=${BASH_VERSION/.*}
+else
+    BASH_VERSION=$(bash -c "printf %s \"\$BASH_VERSION\"")
+    bash_version=5
+    shopt -s eval_unsafe_arith
+fi
+
 sys_locale=${LANG:-C}
 XDG_CONFIG_HOME=${XDG_CONFIG_HOME:-${HOME}/.config}
 PATH=$PATH:/usr/xpg4/bin:/usr/sbin:/sbin:/usr/etc:/usr/libexec
@@ -1593,6 +1600,10 @@ get_shell() {
             shell=${shell/version}
         ;;
 
+        osh)
+            shell+=$("$SHELL" -c "printf %s \"\$OIL_VERSION\"")
+        ;;
+
         tcsh)
             shell+=$("$SHELL" -c "printf %s \$tcsh")
         ;;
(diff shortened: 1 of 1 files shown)


Crestwave:
Actually, `$bash_version` might be unnecessary, unless I'm missing something (if so, just discard that commit).

neofetch
```diff
@@ -30,7 +30,13 @@
 
 version=7.0.0
 
-bash_version=${BASH_VERSION/.*}
+if [[ "${BASH_VERSINFO[0]}" ]]; then
+    bash_version=${BASH_VERSINFO[0]}
+else
+    bash_version=5
+    shopt -s eval_unsafe_arith
+fi
```
dylanaraps, line 38: It's probably easier to just do `bash_version=${BASH_VERSINFO[0]:-5}` and then run the shopt with `&>/dev/null` (always).

neofetch
```diff
@@ -4261,7 +4279,11 @@ info() {
     [[ "$prin" ]] && return
 
     # Update the variable.
-    output="$(trim "${!2:-${!1}}")"
+    if [[ "$2" ]]; then
+        output="$(trim "${!2}")"
+    else
+        output="$(trim "${!1}")"
+    fi
```
dylanaraps, line 4286:
```sh
output=${!2}
output=$(trim "${output:-${!1}}")
```

neofetch
dylanaraps: Question, will it be difficult for us to remove the shopt entirely? What needs to be done?

neofetch
Crestwave: Oh, that's basically the same thing: trying to use a positional parameter as a variable name when it's unset. So it crashes as well

neofetch
```diff
@@ -30,7 +30,8 @@
 
 version=7.0.0
 
-bash_version=${BASH_VERSION/.*}
+bash_version=${BASH_VERSINFO[0]:-5}
+shopt -s eval_unsafe_arith &>/dev/null
```
dylanaraps, line 34:
```sh
# Fallback to a value of '5' for shells which support bash
# but do not set the 'BASH_' shell variables (osh). 
bash_version=${BASH_VERSINFO[0]:-5}
shopt -s eval_unsafe_arith &>/dev/null

```

neofetch
dylanaraps: Ah. Alrighty then.

neofetch
```diff
@@ -30,6 +30,8 @@
 
 version=7.0.0
 
+# Fallback to a value of '5' for shells which support bash
+# but do not set the 'BASH_' shell variables (osh).
 bash_version=${BASH_VERSINFO[0]:-5}
 shopt -s eval_unsafe_arith &>/dev/null
```
dylanaraps, line 36: Add a newline below shopt as well. :+1:

neofetch
```diff
@@ -1583,7 +1586,11 @@ get_shell() {
     [[ $shell_version != on ]] && return
 
     case ${shell_name:=${SHELL##*/}} in
-        bash) shell+=${BASH_VERSION/-*} ;;
+        bash)
+            [[ $BASH_VERSION ]] ||
+                BASH_VERSION=$("$SHELL" -c "printf %s \"\$BASH_VERSION\"")
```
dylanaraps, line 1591: Add a newline here.

neofetch
Crestwave: Note that I actually set `BASH_VERSION` later on. I could use `shell+=$BASH_VERSION; shell=${shell/-*}` instead

dylanaraps: :+1:

Crestwave: Nevermind, that doesn't make sense. At all.

Crestwave: Okay, I think what I was going for was:

```
if [[ $BASH_VERSION ]]; then
 shell+=$("$SHELL" -c "printf %s \"\$BASH_VERSION\"")
else
  shell+=$BASH_VERSION
fi
shell=${shell/-*}
```

Crestwave: Or even something like:

```
shell+=${BASH_VERSION:-$("SHELL" -c "printf %s \"\$BASH_VERSION\"")}
shell=${shell/-*}
```

But if you're fine with that, then cool

neofetch
Crestwave: I actually haven't had any issues without it, but I noticed that you do an `unset -v "gpus[0]"` on a certain condition in the script, which is invalid in OSH without `eval_unsafe_arith`.

Issue #1448, opened 2020-04-21
Extra newline character when calling neofetch


jtorreno:
## Description

Neofetch emits an extra newline character, which I'm led to believe is not intended behavior as per #910.
Issue discovered on latest and reproducible on master.

And of course, a terrible solution, for the glory of none other than Satan:
```
alias neofetch='function _neofetch() {
    (($# == 0)) && { neofetch | sed -e "40,$ d;s/?25l//g;s/?7l//g" } || neofetch $@
}; _neofetch'
```
Edit: Looks like removing the extra newline character suffixed on both [#L3750](https://github.com/dylanaraps/neofetch/blob/a61724f6b28b6285a003e403c184b5a5a2e155c9/neofetch#L3750) and [#L3751](https://github.com/dylanaraps/neofetch/blob/a61724f6b28b6285a003e403c184b5a5a2e155c9/neofetch#L3751) works so far as to do exactly what it sets out to amend.  

#### Neofetch version

```
% neofetch --version  
Neofetch 7.0.0
```

## Screenshot

![screenshot](https://www.dropbox.com/s/5j31eo0nyawcfk0/neofetch.png?raw=1)

jtorreno:
For what it's worth, `--color_blocks off` behaves as expected, emitting only a single newline.

PR #1449, opened 2020-04-21
Remove extra newlines from get_cols

jtorreno:
## Description

Fix as per #1448.

Blank lines after invocation are now consistent across distro ASCIIs and through either `--color_blocks` setting.


dylanaraps:
Thanks

PR #1454, opened 2020-05-03
Add Dahlia OS to Neofetch

nmcain:
This is confirmed to be working on our Linux based builds and our Zircon based builds. If you need me to add this to the list of distributions at the top of the file, I can do that as well.


dylanaraps:
Thanks :+1:

PR #1459, opened 2020-05-15
Add OS Elbrus support

jubnzv:
OS Elbrus is a Debian-based distribution developed by Russian company [MCST](https://en.wikipedia.org/wiki/MCST) for [Elbrus](https://en.wikipedia.org/wiki/Elbrus_(computer)) machines.

I found out that screenfetch already knows about us (see [PR](https://github.com/KittyKatt/screenFetch/pull/548)), so I think it would be nice to support OS Elbrus in neofetch too. :)

Here is a screenshot with Elbrus 1s+ machine:
image


neofetch
```diff
@@ -2387,6 +2393,10 @@ get_gpu() {
                         [[ -z "$(trim "$gpu")" ]] && gpu="Intel Integrated Graphics"
                     ;;
 
+                    *"MCST"*)
+                        gpu="${gpu/*MCST*MGA2*/MCST MGA2}"
```
jubnzv, line 2397: Otherwise it will return garbage with pci device number in old versions of distro.

neofetch
```diff
@@ -958,6 +958,12 @@ get_distro() {
                     *) distro="Siduction ($(lsb_release -sic))"
                 esac
 
+            elif [[ -f /etc/mcst_version ]]; then
+                case $distro_shorthand in
+                    on|tiny) distro="OS Elbrus" ;;
+                    *) distro="OS Elbrus $(< /etc/mcst_version)"
+                esac
+
```
dylanaraps, line 966: There's no `/etc/os-release` or `lsb_release` command?

jubnzv: No, they aren't.

nmcain:
Does neofetch even have full support for MCST CPUs?

jubnzv:
> Does neofetch even have full support for MCST CPUs?

What do you mean? neofetch can detect the OS type and print some system information, as in my screenshot above.

nmcain:
Ah. I was believing that Elbrus was a new architecture or something. It has
full support for the CPUs?

jubnzv:
> Ah. I was believing that Elbrus was a new architecture or something.

Sure. Elbrus has its own processor architecture.

neofetch will detect CPU type [reading /proc/cpuinfo](https://github.com/dylanaraps/neofetch/blob/0e448167c02f28603425c6913a33439d5d3a5e05/neofetch#L2046-L2102). For Elbrus this works the same way as for other Linux systems, because it runs custom distribution (OS Elbrus) based on Linux kernel.

PR #1472, opened 2020-06-05
Turn memory percent on

JavaSight:
## Description

Turns the memory display percent on by default.

## Features

May help people understand better.

## Issues

Hasn't been tested yet.

## TODO

(MAYBE) Better support for percent?


JavaSight:
Travis CI just passed.

Issue #1474, opened 2020-06-07
playerctl support


craffate:
## Description
Opening an issue as I'm unsure if [playerctl](https://github.com/altdesktop/playerctl) can be considered a music player since it's more of a MPRIS-enabled music player controller.

## Why?
I'm personally running [spotifyd](https://github.com/Spotifyd/spotifyd) as a systemd user service to play music and neofetch detects my music player to be "spotify" with `music_player="auto"` set.

I don't actually have spotify running so it reports wrong metadatas:
![20200607_14h19m00s_grim](https://user-images.githubusercontent.com/23284385/83968473-df82b700-a8c9-11ea-9527-0da05093fd3f.png)

A quick [patch](https://github.com/craffate/neofetch/commit/53f6f1c88727bdc3335dcd5e002743ed308f59cd) allows me to get the correct metadatas when setting `music_player="playerctl"`:
![20200607_14h25m47s_grim](https://user-images.githubusercontent.com/23284385/83968632-d5ad8380-a8ca-11ea-8d58-ed558868cafc.png)

It's working for my personal use but I was hesitant on making a PR.
[playerctl](https://github.com/altdesktop/playerctl) is **not** a music player, per-se.

craffate:
Made a PR: #1475

PR #1475, opened 2020-06-07
Add playerctl support

craffate:
## Description
Adds support for the [playerctl](https://github.com/altdesktop/playerctl) music player controller.

## Features
Adds support for [playerctl](https://github.com/altdesktop/playerctl).

## Issues
Doesn't get precedence over other music players when using `music_player="auto"`.


craffate:
_Fixes_ #1474.

Issue #1486, opened 2020-06-23
Various parts broken on macOS 11 (Big Sur) Developer Preview


Seshpenguin:
## Description

Various parts of neofetch do not work correctly on macOS Big Sur, including:

- OS (displays "unknown")
- Uptime
- WM
- Memory

This occurs on the master branch.

#### Neofetch version
7.0.0

## Screenshot
<img width="810" alt="Screen Shot 2020-06-23 at 6 16 57 PM" src="https://user-images.githubusercontent.com/6313946/85471269-33ee8d80-b57e-11ea-8be4-cdf1484a4062.png">

## Config file
Default from Homebrew

## Verbose log
https://pastebin.com/LFhb00Su

slinkouswasalreadytaken:
Can confirm this, though it's a bit different on my end: 
![Screen Shot 2020-06-29 at 2 30 55 PM](https://user-images.githubusercontent.com/53449377/86042632-6fbda300-ba15-11ea-8b3f-10bd768643e2.png)

omBratteng:
Same issue, and it opens the xquartz app and doesn't close it.

omBratteng:
@dylanaraps tested your commit on Big Sur Beta-2, and it works fine. And xquartz doesn't open.

gottabeleGIT:
Could you push a release to include this fix?

omBratteng:
@gottabeleGIT if you have installed it via brew, you can reinstall it but use the --HEAD flag to get the latest version. Or you could build it yourself

gottabeleGIT:
That is what I have done in the interim, but thanks anyway for the tip :)

gottabeleGIT:
Finally, seems to be resolved with release 7.1.0! Thanks @dylanaraps and @stevejkang :D

PR #1487, opened 2020-06-26
Add latest macOS support and fix issue

stevejkang:
## Description
Support for both "Mac OS X" and "macOS"
Support for macOS 10.16(11.0), Big Sur
Resolve issue #1486 

## Comment
The new macOS version, Big Sur, officially announced as 11.0, but according to `/System/Library/CoreServices/SystemVersion.plist`, its version is 10.16. So I made both works well.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -1087,7 +1087,7 @@ get_distro() {
             distro=${distro/NAME=}
         ;;
 
-        "Mac OS X")
+        "Mac OS X"|"macOS")
             case $osx_version in
                 10.4*)  codename="Mac OS X Tiger" ;;
                 10.5*)  codename="Mac OS X Leopard" ;;
@@ -1101,6 +1101,8 @@ get_distro() {
                 10.13*) codename="macOS High Sierra" ;;
                 10.14*) codename="macOS Mojave" ;;
                 10.15*) codename="macOS Catalina" ;;
+                10.16*) codename="macOS Big Sur" ;;
+                11.0*)  codename="macOS Big Sur" ;;
                 *)      codename=macOS ;;
             esac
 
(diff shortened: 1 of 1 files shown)


stevejkang:
After searching on google, Intel-based mac uses 10.16 and the ARM-based mac uses 11.0.

gottabeleGIT:
Actually it seems that Apple called it 10.16 internally so as to prevent leaks as it shows up as 11.0 on my Intel Mac in “About this Mac”. Apple will likely change the internal version in future beta releases.

stevejkang:
> > After searching on google, Intel-based mac uses 10.16 and the ARM-based mac uses 11.0.
> 
> Actually it seems that Apple called it `10.16` internally so as to prevent leaks as it shows up as `11.0` on my Intel Mac in `About this Mac`. Apple will likely change the internal version in future beta releases.
> 
> EDIT: Not sure if this is related, but the System folder now shows a `macOS` logo instead of an `X` logo, possibly further confirming that `10.16` is only temporary and should be changed to `11.0` before the GM.

Apple started using the name `macOS` from 10.12 Sierra. I am not sure if `Mac OS X` or `macOS` was used on the system exactly since version after 10.12, but I think that this project has to cover all environments. Also, 10.16 and 11.0 will be mixed until ARM Mac released and before GM. Most of the contents are not available like below, so I sent a patch-type PR.

> Before


Screen Shot 2020-07-10 at 10 26 39

> After

Screen Shot 2020-07-10 at 10 36 06

gottabeleGIT:
Sounds good, and makes sense
#
Although to clarify, I was referencing the /System folder icon:

Catalina:
Screen Shot 2020-07-10 at 8 19 13 AM

Big Sur:
Screen Shot 2020-07-10 at 8 18 20 AM

Historically, the /System folder had the icon of the "major" Mac OS version number it contained (e.g. Mac OS 9)
tagoreillycom20070531oreillyimages273713 png

So it would make sense for them (in my mind) to change it from X to something else if they were moving away from 10.x (they probably didn't use 11, possibly, again, to prevent leaks).

#

Either ways, I think the PR makes sense, for the purposes of getting `neofetch` to work during the betas and in the off chance 10.16 sticks around until the GM, so that's fine.

stevejkang:
Could you review this PR? @dylanaraps

gottabeleGIT:
So, FYI I just found this from the Big Sur Beta release notes known issues:
“The system ProductVersion is reported as 11.0 only on the Developer Transition Kit at this time. Existing Mac systems will temporarily report the value as 10.16.”

gottabeleGIT:
Screen Shot 2020-07-22 at 5 53 27 PM
Finally! 😄

Issue #1505, opened 2020-07-13
Show correct ubuntu flavor


philipzae:
There are different environment variables that can be used to correctly identify which ubuntu flavor is being used, primarily XDG_CONFIG_DIRS.

### Kubuntu
XDG_CONFIG_DIRS=/etc/xdg/**xdg-plasma**:/etc/xdg

### Ubuntu Mate
XDG_CONFIG_DIRS=/etc/xdg/**xdg-mate**:/etc/xdg
DESKTOP_SESSION=mate
XDG_SESSION_DESKTOP=mate
XDG_CURRENT_DESKTOP=MATE
XDG_DATA_DIRS=**/usr/share/mate**:/usr/share/xfce4:/usr/local/share:/usr/share:/var/lib/snapd/desktop:/usr/share
GDMSESSION=mate

### Xubuntu
XDG_CONFIG_DIRS=/etc/xdg/**xdg-xubuntu**:/etc/xdg:/etc/xdg
DESKTOP_SESSION=**xubuntu**
XDG_SESSION_DESKTOP=**xubuntu**
XDG_DATA_DIRS=**/usr/share/xubuntu**:/usr/share/xfce4:/usr/local/share:/usr/share:/var/lib/snapd/desktop:/usr/share
GDMSESSION=**xubuntu**

### Lubuntu
XDG_CONFIG_DIRS=/etc/xdg/**xdg-Lubuntu**:/etc/xdg:/etc/xdg
DESKTOP_SESSION=**Lubuntu**
XDG_DATA_DIRS=**/usr/share/Lubuntu**:/usr/local/share:/usr/share:/var/lib/snapd/desktop

arisinfenix:
Here is a patch based on your infos. I've tested it only on Kubuntu, since I use it on my netbook.
https://github.com/arisinfenix/neofetch/commit/f47b8b07f7483b7323669a6df1d7cad98766a9e3

@dylanaraps let me know if I should open a PR.

philipzae:
@arisinfenix thanks for the patch. I tested it in Xubuntu and Lubuntu and it worked.

It didn't work in Ubuntu Mate as you'd need to change your code to 'Ubuntu-MATE' rather than 'Ubuntu MATE'.

Please add to your patch detection of ubuntu budgie.
XDG_CONFIG_DIRS=/etc/xdg/**xdg-budgie-desktop**:/etc/xdg
DESKTOP_SESSION=**budgie-desktop**
[...]

PR #1507, opened 2020-07-17
Detect Ubuntu flavors

arisinfenix:
## Description
Added detection of ubuntu flavors. Solves #1505

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -1087,6 +1087,18 @@ get_distro() {
 
             distro=$(trim_quotes "$distro")
             distro=${distro/NAME=}
+
+            # Get Ubuntu flavor.
+            if [[ $distro == "Ubuntu"* ]]; then
+                case $XDG_CONFIG_DIRS in
+                    *"plasma"*)  distro=${distro/Ubuntu/Kubuntu} ;;
+                    *"mate"*)    distro=${distro/Ubuntu/Ubuntu MATE} ;;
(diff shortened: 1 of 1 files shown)


neofetch
```diff
@@ -1087,6 +1087,22 @@ get_distro() {
...
+                done
+
+                case $flavor in
+                    "plasma")  distro=${distro/Ubuntu/Kubuntu} ;;
+                    "mate")    distro=${distro/Ubuntu/Ubuntu MATE} ;;
+                    "xubuntu") distro=${distro/Ubuntu/Xubuntu} ;;
+                    "Lubuntu") distro=${distro/Ubuntu/Lubuntu} ;;
+                    "budgie")  distro=${distro/Ubuntu/Ubuntu Budgie} ;;
+                    "ubuntustudio") distro=${distro/Ubuntu/Ubuntu Studio} ;;
+                esac
```
dylanaraps, line 1104: Does something like this not work?

```
case $XDG_CONFIG_DIRS in
    *plasma*) ;;
    ...
esac
```

arisinfenix: Ah, yes that looks nicer and should also work. I will try it, thanks.

arisinfenix:
Just force pushed the changes and tested it with the flavors live CDs. Works fine.

arisinfenix:
Ah wait, I accidently removed the $distro == Ubuntu test.

arisinfenix:
Added back the test now.

dylanaraps:
Thanks!

PR #1517, opened 2020-08-03
Ubuntu + Flavors: Use more serious colors.

ItzSwirlz:
Don't use defualt colors-we have up to 256 values to test. Make
use of them and have fun!

Xubuntu is sadly a bit too dark-17 - 21 will work so it has to be a pick/choose,
so this may vary over time.

If any whitespaces/empty remove/new exact line edits, that is a whitespace remove.
Atom automatically does that so don't complain at me.

## Description

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -7624,7 +7624,7 @@ EOF
         ;;
 
         "Kubuntu"*)
-            set_colors 4 7 1
+            set_colors 33 7 1
             read -rd '' ascii_data <<'EOF'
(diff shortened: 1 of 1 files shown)


dylanaraps:
It's a requirement that we use 0-16 for colors.

arisinfenix:
Oh, didn't know that. So the colors in just merged #1518 should be reverted to 4, right?

ItzSwirlz:
Yeah.

PR #1522, opened 2020-08-04
Update Ubuntu Mate logo

arisinfenix:
This time just the unmodified output from www.text-image.com.
Perhaps not the best, but definitely an improvement. :)

mate


dylanaraps:
:+1: :+1:

PR #1528, opened 2020-08-07
song: Switch Exaile to get_song_dbus

arisinfenix:
## Description

Exaile 4.0.0 was released more than a year ago (06 Jun 2019). Since it is not packaged in debian stable/testing/sid at all, I think switching to MPRIS should be ok now.


dylanaraps:
Lovely

PR #1529, opened 2020-08-08
song: Switch Quodlibet to get_song_dbus

arisinfenix:
## Description
Quodlibet comes with mpris2 support, so we could use get_song_dbus.

## Issues
The only drawback I see is that the mpris plugin is disabled by default (at least on Ubuntu 20.04).

If you think it's not a good idea to make the switch just close this. :)

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -2768,6 +2768,7 @@ get_song() {
         "tauonmb"*)       get_song_dbus "tauon" ;;
         "olivia"*)        get_song_dbus "olivia" ;;
         "exaile"*)        get_song_dbus "exaile" ;;
+        "quodlibet"*)     get_song_dbus "quodlibet" ;;
         "netease-cloud-music"*)        get_song_dbus "netease-cloud-music" ;;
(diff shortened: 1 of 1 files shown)


dylanaraps:
I can't find any information in its source code to suggest that you can even turn this off/on?

arisinfenix:
I have not found the source code parts, but here is a screenshot.

According to https://github.com/quodlibet/quodlibet/blob/ed4d84575e11b976d33745d7afd007c3a1fbed1b/docs/packaging.rst
it needs `dbus-python` to work which is listed as optional runtime and plugin dependency. The current query method already needs it (for dbus interface). Or am I wrong here?

I'm fine with leaving it as is to be safe, just thought it cleans the song function a bit. ;)

Untitled

arisinfenix:
I'm going to close this for now. Perhaps it's better to have it work without any user actions needed. Also, the way it is it avoids bug reports of kind "Quodlibet dos not show song"... :)

PR #1530, opened 2020-08-08
improve makefile

Uniminin:


dylanaraps:
Open another pull request with an explanation of the changes.

Kreyren:
Wow.. This merge request has so many issues..

So i just provide this as a public service:

---

#### Practice for parsing environmental variables through a Makefile in a shell

You are using:

```Makefile
PROGRAM = neofetch

...

install:
	...
	@ if command -v ${PROGRAM} >/dev/null; then printf "%s\n" "${PROGRAM} has been installed on this system."; fi
...
```

Which has two issues:

1. You have to escape dollar sign if you expect `make` to parse it in the shell through Makefile:

```console
kreyren@leonid:~/test.uniminin$ cat Makefile 
all:
	@ export SOMEVAR="test"; echo "$SOMEVAR"
kreyren@leonid:~/test.uniminin$ make
OMEVAR
kreyren@leonid:~/test.uniminin$ cat Makefile 
all:
	@ export SOMEVAR="test";  echo "$$SOMEVAR"
kreyren@leonid:~/test.uniminin$ make
test
```

2. Your variable is not accessible to the called shell

```console
kreyren@leonid:~/test.uniminin$ cat Makefile 
SOMEVAR = test

all:
	@ echo "$$SOMEVAR"
kreyren@leonid:~/test.uniminin$ make
<blank>
```

To make it accessible you have to use:

```console
kreyren@leonid:~/test.uniminin$ cat Makefile 
SOMEVAR = test

all:
	@ echo "$(SOMEVAR)"
kreyren@leonid:~/test.uniminin$ make
test
```

---

#### Makefile will terminate if you provide 'false'

In your contribution:

```Makefile
...
	if command -v ${PROGRAM} >/dev/null; then  ...
...

```

This will terminate if the `if` statement returns anything other then `0`

```console
kreyren@leonid:~/test.uniminin$ cat Makefile 
all:
	@ false
	@ echo "Whatever"
kreyren@leonid:~/test.uniminin$ make
make: *** [Makefile:2: all] Error 1
```

Each `@` is basically running a **one-line** shell with `set -x` so if you want to sanitize the Makefile using `if` statement then you would do something alike:

```Makefile
kreyren:
	@ [ ! -f neofetch ] || rm neofetch
```

If file `neofetch` exists it will return `0` and continue

If file `neofetch` does not exists it will trigger `rm` (if `rm` removes the file it will return `0` else it will return 1~255 on a 64 bit system 8 octets wide.

@Uniminin Sent me a PM if you need more explanation so that i don't spam dylan with things that are not directly related to neofetch.

Uniminin:
got it!

Issue #1503, opened 2020-07-13
Terminal font detection fails


cinerea0:
## Description

Neofetch fails to detect my terminal font.

- [X] Does this issue still occur in the master branch? (**Required if issue**)

#### Neofetch version
7.0.0

## Screenshot
![neofetch](https://user-images.githubusercontent.com/49368915/87265753-1f355380-c4b3-11ea-88fe-9958b633c703.png)

## Config file
neofetch config file: https://gist.github.com/cinerea0/ec6e6c92d5ad21411ad33f801c6a5b9e
alacritty config file: https://gist.github.com/cinerea0/d2dcfeb1c6bfb2df4b2104cab947de82

## Verbose log
https://gist.github.com/cinerea0/56e8c737c42e313a48e2ddb4e920d4b8

dylanaraps:
Send your alacritty config file. :+1:

cinerea0:
I've listed it under the config file section.

arisinfenix:
@cinerea0 can you test https://github.com/arisinfenix/neofetch/commit/aac6f6530ef8940691390af9762a63b998359c9e ? It should solve the Issue.

PR #1535, opened 2020-08-17
term_font: fix alacritty

### LINKED ISSUE #1534: term_font not working in alacritty when using default config (suggested fix)
Just switched terminal to alacritty, I'm using the default config found [here](https://github.com/alacritty/alacritty/releases/download/v0.5.0/alacritty.yml)
I noticed that it wasn't displaying my terminal font so I did a bit of digging, I took a look at how you were getting the font and tested the command

     awk -F ':|#' '/normal:/ {getline; print}' ~/.config/alacritty/alacritty.yml
        # Font family
        black:   '#0c0301'

this is the relevent bit from my config

    # Font configuration
    font:
      # Normal (roman) font face
      normal:
        # Font family
        #
        # Default:
        #   - (macOS) Menlo
        #   - (Linux/BSD) monospace
        #   - (Windows) Consolas
        family: Iosevka Nerd Font

as you can see, it's finding the comment and then deleting it a couple of lines later leaving $term_font empty.
To fix this I changed line ***3218*** to this

            term_font="$(sed -e '/^\s*#.*$/d' "${confs[0]}" | awk -F ':|#' '/normal:/ {getline; print}')"

using sed to remove all the comments before piping into awk, I imagine you know a better solution though

EDIT: should have looked a little closer at the issues list before posting this, this is the same issue as #1503 - really should have posted this in there, sorry

arisinfenix:
## Description

Fixes #1534 and #1503 

This is the fix confirmed in #1534 plus ignoring comments that is needed to also solve #1503.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -3206,7 +3206,8 @@ get_term_font() {
 
             [[ -f "${confs[0]}" ]] || return
 
-            term_font="$(awk -F ':|#' '/normal:/ {getline; print}' "${confs[0]}")"
+            term_font="$(awk '/normal:/ {while (!/family:/ || /#/)
+                         {if (!getline) {exit}} print; exit}' "${confs[0]}")"
             term_font="${term_font/*family:}"
             term_font="${term_font/$'\n'*}"
             term_font="${term_font/\#*}"


arisinfenix:
Ah, wait. Now it fails with unchanged default config. I'm looking into it.

arisinfenix:
I changed ```{getline}``` to ```{if (!getline) {exit}}``` in the awk command to avoid looping forever in some cases. I think it's ready now.

Issue #1500, opened 2020-07-08
When using GDM WM identifies as Mutter regardless of the DE


AlanGriffiths:
## Description

I have multiple DEs installed and select one at login. Regardless of which I select, the WM is identified as Mutter.

#### Neofetch version

7.0.0 - but also happens on master

## Screenshot
```plain
$ neofetch
            .-/+oossssoo+/-.               alan@Octopull-PROX15-20 
        `:+ssssssssssssssssss+:`           ----------------------- 
      -+ssssssssssssssssssyyssss+-         OS: Ubuntu 20.04 LTS x86_64 
    .ossssssssssssssssssdMMMNysssso.       Host: PROX15 Standard 
   /ssssssssssshdmmNNmmyNMMMMhssssss/      Kernel: 5.4.0-40-generic 
  +ssssssssshmydMMMMMMMNddddyssssssss+     Uptime: 2 days, 38 mins 
 /sssssssshNMMMyhhyyyyhmNMMMNhssssssss/    Packages: 2505 (dpkg), 22 (snap) 
.ssssssssdMMMNhsssssssssshNMMMdssssssss.   Shell: bash 5.0.16 
+sssshhhyNMMNyssssssssssssyNMMMysssssss+   Resolution: 1920x1080 
ossyNMMMNyMMhsssssssssssssshmmmhssssssso   DE: egmde 
ossyNMMMNyMMhsssssssssssssshmmmhssssssso   WM: Mutter 
+sssshhhyNMMNyssssssssssssyNMMMysssssss+   WM Theme: Adwaita 
.ssssssssdMMMNhsssssssssshNMMMdssssssss.   Theme: Adwaita [GTK3] 
 /sssssssshNMMMyhhyyyyhdNMMMNhssssssss/    Icons: Adwaita [GTK3] 
  +sssssssssdmydMMMMMMMMddddyssssssss+     Terminal: gnome-terminal 
   /ssssssssssshdmNNNNmyNMMMMhssssss/      CPU: Intel i7-9750H (12) @ 4.500GHz 
    .ossssssssssssssssssdMMMNysssso.       GPU: Intel UHD Graphics 630 
      -+sssssssssssssssssyyyssss+-         GPU: NVIDIA GeForce GTX 1650 Mobile / Max-Q 
        `:+ssssssssssssssssss+:`           Memory: 1643MiB / 31973MiB 
            .-/+oossssoo+/-.
```                                                                   

[...]

PR #1539, opened 2020-08-19
Better handling of Wayland compositors

AlanGriffiths:
## Description

Better handling of Wayland desktops

## Features

Corrects detection of Wayland compositors when `WAYLAND_DISPLAY` is unset
Doesn't identify window managers belonging to other user session

## Issues

The logic for BSD is unchanged (I'm too lazy to set up a test system)

Fixes: #1500

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -1788,40 +1788,41 @@ get_wm() {
     # If function was run, stop here.
     ((wm_run == 1)) && return
 
-    case $kernel_name in
-        *OpenBSD*) ps_flags=(x -c) ;;
-        *)         ps_flags=(-e) ;;
-    esac
-
-    if [[ $WAYLAND_DISPLAY ]]; then
-        wm=$(ps "${ps_flags[@]}" | grep -m 1 -o -F \
-                           -e arcan \
-                           -e asc \
-                           -e clayland \
-                           -e dwc \
-                           -e fireplace \
-                           -e gnome-shell \
-                           -e greenfield \
-                           -e grefsen \
-                           -e kwin \
-                           -e lipstick \
-                           -e maynard \
-                           -e mazecompositor \
-                           -e motorcar \
-                           -e orbital \
-                           -e orbment \
-                           -e perceptia \
-                           -e rustland \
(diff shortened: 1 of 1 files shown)


arisinfenix:
```bash
fuser -v "${XDG_RUNTIME_DIR}/${WAYLAND_DISPLAY:-wayland-0}"
```
This does not work on openSUSE (Tumbleweed) without root privileges.

```bash
susi@linux-viez:~> echo $WAYLAND_DISPLAY
wayland-0
susi@linux-viez:~> echo $XDG_RUNTIME_DIR
/run/user/1000
susi@linux-viez:~> ls -l ${XDG_RUNTIME_DIR}/${WAYLAND_DISPLAY:-wayland-0}
srwxr-xr-x 1 susi users 0 19. Aug 15:39 /run/user/1000/wayland-0
susi@linux-viez:~> fuser -v ${XDG_RUNTIME_DIR}/${WAYLAND_DISPLAY:-wayland-0}
susi@linux-viez:~> sudo fuser -v ${XDG_RUNTIME_DIR}/${WAYLAND_DISPLAY:-wayland-0}
                     BEN.        PID ZUGR.  BEFEHL
/run/user/1000/wayland-0:
                     susi       2654 F.... kwin_wayland
```

arisinfenix:
I think this approach looks nice since it avoids grepping from a hardcoded list of compositors.

Sadly `lsof -t "${XDG_RUNTIME_DIR}/${WAYLAND_DISPLAY:-wayland-0}"` also needs root permissions to work on Tumbleweed.

AlanGriffiths:
> ...also needs root permissions

Disappointing. I assume it fails as non-root? I.e. something like this would detect if it works...
```plain
if wm_pid=$(lsof -t "${XDG_RUNTIME_DIR}/${WAYLAND_DISPLAY:-wayland-0}" 2>&1)
...
```

arisinfenix:
> I assume it fails as non-root?

Yup, no output at all as non-root.

dylanaraps:
This is great, thanks.

PR #1543, opened 2020-08-25
Strip color codes from ascii art line length calculations

Syphist:
## Description
This removes color and formatting codes from the line length calculation to allow for more control over custom ascii art. This would also fix issues I found when attempting to use the patch in #1220.

## Features
Allows for support for finer control of ascii art using escapes such as "[38;5;39m" for defining coloring and formatting in the terminal.

## Issues
None that I have found

## TODO
Nothing

## Fixed Issues
This so far only works if you directly paste the character for \033 or "" into the file. I was unable to get it to work for both.


Syphist:
I have made several more attempts to work around SC2001, but have had no success. I am wondering if it is too complex to do it with the search and replace in bash and may need to be done with sed. I will probably make further attempts at this, if anyone else has any ideas feel free to let me know too.

Syphist:
I have figured out the solution and have updated my code accordingly and have cleaned it up. I have tested both with the pasted character in the ascii file and the escape code and both work. This is also much faster than my initial sed implementation. I have updated the OP to reflect these changes as well as clarify a few things.

Z-8Bit:
+1 I agree, it's much needed this.

thenick775:
+1 this modification works well for me, have tested on Big Sur

Screen Shot 2021-12-27 at 9 54 00 PM

fvegaswp:
Just tested this with colored ascii generated with lolcat and works great.

jonaramos:
Tested with colored ascii generated from jp2a and ascii-image-converter. Thanks Syphist.
image

hykilpikonna (approved):

hykilpikonna:
Lol I actually ran into this issue while creating HyFetch. My solution was to pass in a separate `ascii_len` variable:

image

Thanks a lot!

hykilpikonna:
Works!

image

hykilpikonna:

This PR is [merged into hyfetch](https://github.com/hykilpikonna/hyfetch/commit/eeead8d7) since this repo (dylanaraps/neofetch) seems no longer maintained.

hykilpikonna:
I found a bug in the last line of `strip_escape_codes`:

```diff
-     eval "$2=\"${output}\""
+     eval "$2='${output}'"
```

If there exists $ signs in the ascii art, this eval statement will replace the $ with a variable content. You should use single quotes to prevent this.

**Before Fix:**

image

**After Fix:**

image

neofetch
```diff
@@ -4054,6 +4054,26 @@ image_backend() {
...
+            esac
+            continue
+        fi
+        if [[ "${char}" == $'\e' ]]; then
+            within_code=1
+            continue
+        fi
+        output+="${char}"
+    done
+    eval "$2=\"${output}\""
```
hykilpikonna:
```suggestion
    eval "$2='${output}'"
```

hykilpikonna (changes requested):

Syphist:
I have committed your fix. Thanks for finding a solution and testing further.

hykilpikonna:
Okay I found another bug lol

```diff
- while IFS=$'\n' read -r line; do
+ while IFS=$'\n' read line; do
```

The width is incorrectly calculated for ascii distros with backslashes in them (e.g. Bedrock) because the backslashes were not treated as escape characters in `read -r` while they are actually escaped when defining the ASCII:

image

So, if `read -r` is used, the escaped backslash (`\\`) will not be resolved as one character but two characters, making the ascii length longer. (https://github.com/hykilpikonna/hyfetch/issues/19)

**Before Fix:**

image


**After Fix:**

image

Syphist:
Thanks @hykilpikonna I fixed the issue on my fork as well just in case this ever gets merged upstream to neofetch or if someone bothers to use my fork.

PR #1573, opened 2020-10-04
Correct Shell, DE and WM on Windows

### LINKED ISSUE #1571: [Feature Request]: Correct Shell, DE and WM on Windows
## Description

Refer to this image: 

94721313-55c49100-0373-11eb-8095-f4ff1844ed90

1. I think in most use cases the default shell being used on Windows is either PowerShell or CMD. Neofetch still returns bash. That's understandable because underneath it's using Cygwin/MINGW/Git Bash. But there should be an option to display PowerShell or CMD, or if none of these are detected, default to bash. If the OS is detected as Windows, the [COMSPEC](https://en.wikipedia.org/wiki/COMSPEC) (command interpreter specification) environment variable, which is available in Git Bash (since the environment is inherited) should be used. COMSPEC is always available in any Windows system, and usually points to `cmd.exe` but users change it to `powershell.exe` or `pwsh.exe` as well. This is the equivalent of `$SHELL` unix variable. This can be used in the `get_shell()` function. 

2. The DE on Windows 10 is Fluent, and not Aero. Refer to [this](https://en.wikipedia.org/wiki/Desktop_environment#Examples_of_desktop_environments) Wikipedia article. TL,DR; Luna was the DE on Windows XP, Aero on Windows Vista and 7, Metro on Windows 8 and 8.1, and Fluent on Windows 10. The fix (my estimate) for this should be fairly simple, just adding some 'case' statements in the `get_de()` function. 

3. Windows uses `DWM.exe` as the WM, as mentioned [here](https://en.wikipedia.org/wiki/Desktop_Window_Manager). `Explorer.exe` is responsible for the desktop, the file manager, the taskbar, and most visible components of the Windows shell. Both these are commonly confused with each other, refer to [this](https://stackoverflow.com/a/5220133) StackOverflow answer for more details. 

4. The GPU output is being weird here, but PR #1540 already fixes that. 

I might send a PR solving the above (except 4.) if I'm able to wrap my head around this stuff. 

#### Neofetch version

7.1.0

rashil2000:
Fixes #1571.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -1911,7 +1911,7 @@ get_wm() {
                                      -e litestep)
 
                 [[ $wm == blackbox ]] && wm="bbLean (Blackbox)"
-                wm=${wm:+$wm, }Explorer
+                wm=${wm:+$wm, }DWM.exe
             ;;
 
             FreeMiNT)
--- a/neofetch
+++ b/neofetch
@@ -1720,7 +1720,8 @@ get_de() {
 
         Windows)
             case $distro in
-                "Windows 8"*|"Windows 10"*) de="Modern UI/Metro" ;;
+                *"Windows 10"*) de="Fluent" ;;
+                *"Windows 8"*) de="Metro" ;;
                 *) de=Aero
             esac
         ;;
--- a/neofetch
+++ b/neofetch
@@ -1637,59 +1637,78 @@ get_packages() {
 }
 
 get_shell() {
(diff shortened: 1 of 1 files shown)


dylanaraps:
Thanks for the PR. Fixed in https://github.com/dylanaraps/neofetch/commit/4b5f4e33beec42db4d3278ab92c7549656dc8202

Issue #1574, opened 2020-10-04
Improvement idea for multiprocessor systems, current output is unclear


Nalorokk:
## Description

Neofetch provides such output on dual xeon:

CPU: Intel Xeon E5405 (8) @ 2.000GHz

I think it would be a lot cleaner if it would be instead like this:

CPU: 2 x Intel Xeon E5405 (4) @ 2.000GHz

#### Neofetch version
7.1.0

[https://gist.githubusercontent.com/Nalorokk/25ab7f5969e4540d745d349204a9005f/raw/f45ef29957b73c610ae66067985d1a0e0d86292e/neofetchlog](neofetch verbose log)

hykilpikonna:
Just solved this in my fork [HyFetch](https://github.com/hykilpikonna/HyFetch). (See commit https://github.com/hykilpikonna/hyfetch/commit/c176e6ffaef6394969159485fa9b5851df4d36f6)

I only have a single-cpu system but I'm pretty sure that it would work on a multi-cpu system as well. You can test this out!

![image](https://user-images.githubusercontent.com/22280294/184272492-3aa59240-f57f-4b01-83b1-e188da0cd732.png)

PR #1585, opened 2020-10-22
Fix for Improvement idea for multiprocessor systems, current output is unclear #1574

swamytk:
## Description
Fix for Improvement idea for multiprocessor systems, current output is unclear #1574

## Features
Both CPU Physical and Logical cores are displayed for Linux. This requires "--cpu_cores hybrid" option to be passed.
Tested with shellcheck
Supported only for Linux

## TODO
man page to be updated - Please let me know how to generate man page with the help of help2man

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -275,13 +275,14 @@ cpu_speed="on"
 # Display CPU cores in output
 #
 # Default: 'logical'
-# Values:  'logical', 'physical', 'off'
+# Values:  'logical', 'physical', 'hybrid', 'off'
 # Flag:    --cpu_cores
-# Support: 'physical' doesn't work on BSD.
+# Support: 'physical' doesn't work on BSD. 'hybrid' works on Linux only.
 #
 # Example:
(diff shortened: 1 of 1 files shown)


hykilpikonna:
This doesn't solve #1574...

#1574 is talking about multiple-socket systems (like one computer with two physical cpus). This PR will only display logical cores vs physical cores.

For example, my system has 1 E5-2696 v3 CPU with 18 physical cores and 36 logical cores, this version of neofetch will show `18 x E5-2696 v3 (2)` which doesn't make sense.

image

The correct solution would be to use the socket field in `lscpu`

Like, `lscpu | awk -F': ' '/Socket\(s\)/ {print $2; exit}'` would return 2 on a dual-cpu system

Or counting cpuinfo also works: `grep "physical id" /proc/cpuinfo | sort -u | wc -l`

hykilpikonna:
This would be better:

image
image
image

PR #1592, opened 2020-11-03
feat(backend): add viu image backend

oknozor:
## Description

This add [viu](https://github.com/atanunq/viu) image backend. 
I added this because Überzug was not working for me (alacritty + sway/wayland). 

## Feature 

Image can now be rendered under wayland (low definition but still). 

## TODO

Still need to document this in the wiki (how to I do it ?)

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -749,7 +749,8 @@ disk_display="off"
 #
 # Default:  'ascii'
 # Values:   'ascii', 'caca', 'catimg', 'chafa', 'jp2a', 'iterm2', 'off',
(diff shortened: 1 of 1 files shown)


neofetch
```diff
@@ -4398,6 +4400,12 @@ display_image() {
             "$image"
         ;;
 
+        "viu")
+            viu \
+                -t -w ""$((width / font_width))"" -h "$((height / font_height))" \
```
dylanaraps, line 4405: `""$((width / font_width))""` should be `"$((width / font_width))"`

oknozor: fixed

PR #1601, opened 2020-11-10
Adds the BigLinux logo to neofetch

ezileicorreia:
## Description

BigLinux is a Brazilian distribution, based on Ubuntu, focused on creating facilities for the user.


neofetch
```diff
@@ -5400,6 +5400,7 @@ get_distro_ascii() {
     #
     # $ascii_distro is the same as $distro.
     case $(trim "$ascii_distro") in
+
```
dylanaraps, line 5403: ???

ezileicorreia: Details!

PR #1602, opened 2020-11-10
Add HoleOS's ASCII image

BenjaminStonawski:
## Description

HoleOS (HoleOS Sarti) is an Arch-based distro that is still under development. 

## Features

- HoleOS brings a user-friendly interface for Arch Linux for people who want to use Arch but are unfamiliar with the command line
- There are so many package managers such as pacman, yay, and git.
- It gets patches regularly.

## TODO

We would like you to add our ASCII image to Neofetch.
The ASCII image: https://github.com/BenjaminStonawski/HoleOS/blob/master/Neofetch/holeascii.txt


dylanaraps:
This PR contains more than your changes. Please resubmit it.

PR #1614, opened 2020-11-24
Refactor ueberzug

silentnoodlemaster:
## Description

This PR refactors the way Überzug is used, as the old way is deprecated


Rashad-707:
The image disappear when I want to enter a command

PR #1617, opened 2020-11-26
NixOS - squash logo, add new package location.

kanashimia:
Logo was too tall before, normalized based on ubuntu and manjaro logo.

There is additional directory where user "packages" can be located - when you you set `home-manager.useUserPackages = true;` 
Even tho `nix-store -q` accepts multiple paths - if one of them doesen't exist it fails without any stdout output, so i had to wrap in a function.


Luflosi:
There is a mistake in the code for macOS. I think it should be `manager=nix-user   && tot nix-user-pkgs` instead of `manager=nix-user   && tot nix-store -qR nix-user-pkgs`.

kanashimia:
> There is a mistake in the code for macOS. I think it should be `manager=nix-user && tot nix-user-pkgs` instead of `manager=nix-user && tot nix-store -qR nix-user-pkgs`.

You are right, thanks for noticing!

Issue #482, opened 2016-12-03
Identify the boot loader and payload


picsi:
It would be nice to have the boot loader identified: BIOS, UEFI, EFI, Coreboot, Libreboot...

Also the payload: GRUB, SeaBIOS, depthcharge.

See related item: https://github.com/KittyKatt/screenFetch/issues/345

dylanaraps:
This feature is very unlikely to be added to Neofetch as I fail to see how we're going to implement it for every Operating System we support. I foresee this involving a hardcoded list of Bootloaders + bootloader config files and I don't want to go down that route. 

I'm closing this issue. If anyone comes up with any ideas then feel free to open a new issue or comment in this one.

PR #1645, opened 2020-12-30
Added Bootloader version support for Linux/Mac

renegadevi:
## Description

Shows Bootloader name and version. 

I found an earlier [issue](https://github.com/dylanaraps/neofetch/issues/482) about payload and bootloader info, but that's too much. Initially just me and some others just wanted the name and version for OpenCore.

## Features

Initial support for MacOS/Opencore and Linux/GRUB.

## Issues
- Relying on piping to awk
The  guidelines says to avoid that, but at the same time we don't want a cluttered output.

## TODO

Support more variations.


neofetch
```diff
@@ -2721,6 +2722,22 @@ get_memory() {
     esac
 }
 
+get_bootloader() {
+    case $os in
+        "Linux")
+            if [[ -d "/boot/grub" ]]; then
```
arisinfenix, line 2728: This will fail on openSUSE (and maybe others too), since there the directory is named `/boot/grub2`.

neofetch
```diff
@@ -2721,6 +2722,22 @@ get_memory() {
...
+                bootloader="GRUB ($(grub-install --version | awk '{printf $3}'))"
```
arisinfenix, line 2729: Same here, the command is `grub2-install`, and it needs root privileges on openSUSE.

renegadevi:
all these Linux inconsistencies does make it a little harder to deal with. With this now in mind, I wouldn't be surprised that other distros also has similar issues with this to workaround. @arisinfenix

PR #1658, opened 2021-01-15
added support for public ipv4 and ipv6

usmcfiredog:
## Description

I have noticed that when the option for Public IP is selected that sometimes it will return IPv4 and other time it will show IPv6. I have separated out the public IP info to be either IPv4 or IPv6

## Features

Choose between public IPv4 and IPv6 to be displayed

## Issues
n/a

## TODO
n/a

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -81,7 +81,8 @@ print_info() {
     # info "Song" song
     # [[ "$player" ]] && prin "Music Player" "$player"
     # info "Local IP" local_ip
-    # info "Public IP" public_ip
+    # info "Public IPv4" public_ipv4
+    # info "Public IPv6" public_ipv6
     # info "Users" users
     # info "Locale" locale  # This only works on glibc systems.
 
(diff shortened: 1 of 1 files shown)


neofetch
```diff
@@ -3805,23 +3820,43 @@ get_local_ip() {
     esac
 }
 
-get_public_ip() {
+get_public_ipv4() {
+    if type -p dig >/dev/null; then
+        public_ipv4="$(dig +time=1 +tries=1 +short myip.opendns.com @resolver1.opendns.com)"
```
rasa, line 3825: This is the same dig command as for IPv4. How will it return the IPv6 address in one call and the IPv4 address in anither? Same question for drill.

Also I would list curl and wget before dig and drill, as 
`dig +time=1 +tries=1 +short myip.opendns.com @resolver1.opendns.com`
fails for me in Windows/WSL

Also, `get_public_ipv4` and `get_public_ipv6` are have common logic. Perhaps move it to `_get_public_ip` and call it with params?

usmcfiredog: They are pulling from different URLs. Right now while at home sometimes the URL will return IPv4 and sometimes it will return IPv6. So each time I run it I get a different IP address. The logic is the same because there is no need to change it except for the URL and by having 2 separate options it allows a user to choose either/or or both.

This is just what I found by using the default site of ident.me from the config. I just followed the logic based on a certain URL. I am open for other ideas also.

Issue #1294, opened 2019-07-30
Incorrect package count.


Phate6660:
## Description

So, I'm currently on Gentoo. Using Neofetch would give me the correct package count. Then I decided to hijack my Gentoo install with Bedrock as I have been curious about it for a while. However, after doing so, my package count is significantly higher than it's supposed to be. I currently have 720 packages installed, however Neofetch says I have 2880 packages installed.

#### Neofetch version

6.0.0

## Screenshot

![2019-07-29-222434_1366x768_scrot](https://user-images.githubusercontent.com/44369810/62095890-4f1f2300-b271-11e9-8921-68ba7bc990c2.png)

## Config file

https://privatebin.net/?9d17274c908b35be#BwDEviuxxAXpKwnUJk9iLykzSDpSdd7Ggx5KUJGTY3o

## Verbose log

https://privatebin.net/?8a73bfc16c883472#DwMdAtfPx4ekb8CJdxv9iKRX1JHTapa7TwstxzvHa3Ny

dylanaraps:
This isn't a bug as you are running neofetch as "bedrock" and not an individual strata.

nezuSTDdos:
@dylanaraps Someone on Discord just informed me that this problem is also present within seperate stratas. I'll be attaching an image that someone called Elder from the Bedrock discord provided.
![unknown](https://user-images.githubusercontent.com/56595252/87557380-237e8e00-c6b8-11ea-91b2-67a9c127c37a.png)

paradigm:
_For the most part_ it works as expected for me.

Here's a quick and dirty script I wrote to sanity check things:

```
#!/bin/sh

echo "Neofetch reported system emerge package count:"
neofetch | awk -vRS=' ' '
	/(emerge)/ {
		print x;exit
	}
	{
		x=$0
	}
'

echo "Neofetch reported individual strata emerge package counts:"
for s in $(brl list -i); do
	echo "stratum=$s="
	strat -r $s ./neofetch 2>/dev/null
done | awk -vRS=' ' '
	/stratum=/ {
		split($0, a, "=")
		s=a[2]
	}
	/(emerge)/ {
		print s":\t"x
		t+=x
	}
	{
		x=$0
	}
	END {
		print "------"
		print "total:\t"t
	}
' | column -t
```

On my system it outputs:

```
Neofetch reported system emerge package count:
1637
Neofetch reported individual strata emerge package counts:
brl-build-cross-ppc:  226
[...]

Issue #1671, opened 2021-01-28
Bedrock Linux neofetch bug (on gentoo)


MuhammedTaha02:
## Description
Although the package number is 997, it appears as 3988 in neofetch.

## Neofetch version
Neofetch version: 7.1.0

## Screenshot
![image](https://user-images.githubusercontent.com/62564400/106178050-f603b880-61a1-11eb-961f-b45c7f447408.png)

## Config file
Default

## Verbose log
https://paste.centos.org/view/8f867f0c
Alternative:
https://pastebin.pl/view/1f833fd5

MuhammedTaha02:
drwxr-xr-x 23 root root 4096 Jan 28 19:06 gentoo/
lrwxrwxrwx  1 root root    6 Jan 28 19:05 hijacked -> gentoo/
lrwxrwxrwx  1 root root   23 Nov 10 14:48 init -> /bedrock/run/init-alias/
lrwxrwxrwx  1 root root   21 Nov 10 14:48 local -> ../cross/.local-alias/

Neofetch counts the packages in these 4 folders repeatedly and the value appears to be 4 times

paradigm:
The proper fix here might be to filter out symlinks from that bit of code.  That will add a lot of code/complexity compared to the current one-liners in there.  I'd have to play with it for a bit to find a clean enough solution for Dylan's tastes.

Your proposed PR is not a good general fix, as:

- That explicitly breaks counting for other strata.
- The `hijacked` alias is a default, but not actually required.  You can `brl remove` it.  The Bedrock system I'm using to type this message doesn't have a `hijacked` alias.

If your goal is _only_ to get the hijacked stratum's package count, you can use `strat -r hijacked neofetch`.

MuhammedTaha02:
> The proper fix here might be to filter out symlinks from that bit of code. That will add a lot of code/complexity compared to the current one-liners in there. I'd have to play with it for a bit to find a clean enough solution for Dylan's tastes.
> 
> Your proposed PR is not a good general fix, as:
> 
> * That explicitly breaks counting for other strata.
> * The `hijacked` alias is a default, but not actually required.  You can `brl remove` it.  The Bedrock system I'm using to type this message doesn't have a `hijacked` alias.
> 
> If your goal is _only_ to get the hijacked stratum's package count, you can use `strat -r hijacked neofetch`.

[...]

PR #1675, opened 2021-01-30
Bedrock Linux fixes

paradigm:
Neofetch is Bedrock-aware by default, but ignores Bedrock and restricts itself to the local layer of the Bedrock system when run in Bedrock's "restricted" mode.  Previously, neofetch checked for restricted mode by looking at the $PATH, which can fail in some workflows.  Instead, check for $BEDROCK_RESTRICT, which is more robust.

Neofetch previously counted Bedrock stratum aliases when counting packages for some package managers.  This resulted in over-counting packages.  This commit restricts neofetch to only counting actual strata, not aliases.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -971,7 +971,7 @@ get_distro() {
 
     case $os in
         Linux|BSD|MINIX)
-            if [[ -f /bedrock/etc/bedrock-release && $PATH == */bedrock/cross/* ]]; then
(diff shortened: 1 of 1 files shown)


paradigm:
This fixes https://github.com/dylanaraps/neofetch/issues/1294 and https://github.com/dylanaraps/neofetch/issues/1671

Let me know if any changes are needed to upstream this.

paradigm:
Very happy to see you back, Dylan :)

PR #1683, opened 2021-02-17
Added a 'NF_' prefix to env vars in makefile, to prevent clashing with shell env vars.

dflock:
## Description

Added a 'NF_' prefix to env vars in makefile, to prevent clashing with shell env vars.


dylanaraps:
This is intentional.

PR #1684, opened 2021-02-19
Change Windows' DE name to "Windows Explorer"

AdrUlb:
## Description
The Windows Desktop Environment is part of Windows Explorer which would make it the DE on Windows

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -1730,9 +1730,7 @@ get_de() {
     case $os in
         "Mac OS X"|"macOS") de=Aqua ;;
 
-        Windows)
-            de="Windows Explorer"
-        ;;
+        Windows) de="Windows Explorer" ;;
 
         FreeMiNT)
             freemint_wm=(/proc/*)
--- a/neofetch
+++ b/neofetch
@@ -1731,19 +1731,7 @@ get_de() {
         "Mac OS X"|"macOS") de=Aqua ;;
 
         Windows)
-            case $distro in
-                *"Windows 10"*)
-                    de=Fluent
-                ;;
-
-                *"Windows 8"*)
-                    de=Metro
-                ;;
-
-                *)
-                    de=Aero
-                ;;
-            esac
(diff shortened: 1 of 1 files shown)


rashil2000:
#1571

AdrUlb:
WM is actually correctly reporting as DWM.exe now, all that's left to fix is the DE

rashil2000:
The DEs are correct. Try visiting the references in the issue I linked.

AdrUlb:
If you actually read through the article carefully it states that the Windows shell supports various *Visual Styles* and then lists them of, which are the Aero theme on Vista and 7, the Metro theme on Windows 8 and the Fluent theme on Windows 10, it does not state that these are the names of the DEs on the respective operating systems.

rashil2000:
The thing is, the concept of DEs is a Unix thing; it doesn't really fit well with Windows based systems. 

However, for the sake of distinction between the different versions, it makes sense to list out whatever visual styles the Windows Explorer provides.

Replacing them all with a single term would make it too generic.

AdrUlb:
There are custom desktop environments available for Windows, such as the Cairo Desktop Environment, they are not exclusively a UNIX thing. Luna, Aero, Metro and Fluent are themes. They should be displayed as themes or not at all. If this is really needed for distinction purposes I'd suggest both displaying the themes Luna, Aero, Metro and Fluent under the theme section and perhaps appending the Windows version to the Windows Explorer text in parantheses.

I however have decided to close this PR because every time I comment on it a new comment appears stating subjective reasons to keep objectively wrong terminology: "Windows Explorer" (or "Windows shell") is the default (and not the only) Desktop Environment, the concept of different Desktop Environments did emerge with UNIX but it is NOT a UNIX-only idea and Luna, Aero, Metro and Fluent are themes and not DEs and should either be displayed as such or not at all. What you are advocating for is the wrong solution to the distinction problem.

PR #1686, opened 2021-02-20
Debian: Update distro colors

baco:
Change the c2 color for Debian distro for a 256-base magenta color. This
color for the distro logo resembles the original color of Debian
logotype.


arisinfenix:
I guess this won't be accepted. See https://github.com/dylanaraps/neofetch/pull/1517#issuecomment-668115245

PR #1691, opened 2021-02-28
update

fairyos:
## Description


fairyos:
Like to add my Distro to neofetch

PR #1696, opened 2021-03-06
I use Arch Linux btw

FedericoAntoniazzi:
## Description

Neofetch is widely used in r/unixporn and the sentence "I use arch btw" is a meme so I added the possibility to show "Arch Linux btw" in the OS section.

## Features
Shows "Arch Linux btw" in the OS section only if `--i-use-arch-btw` (or the shorthand `--btw`) is not set to "off".
> **NOTE:** It works only in Arch Linux

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -138,6 +138,17 @@ distro_shorthand="off"
 # off: 'Arch Linux'
 os_arch="on"
 
+# Show btw in case you use Arch Linux
+#
+# Default: 'on'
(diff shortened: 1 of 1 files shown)


C0V1D-19:
# Not the hero we asked for....but rather the hero we needed
> I also use arch btw

thunderstorm010:
i don't use arch, but i think this should be added

PR #1707, opened 2021-03-22
add ZFS pool stats

db48x:
## Description

ZFS pools are reported just like disks, so I opted to have it follow
the disk options (disk_percent, disk_display, etc) rather than adding
new ones just for zpools.

## Features

`neofetch zpools`

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -76,6 +76,7 @@ print_info() {
 
     # info "GPU Driver" gpu_driver  # Linux/macOS only
     # info "Disk" disk
+    # info "ZFS" zpools     # if you have ZFS pools
     # info "Battery" battery
     # info "Font" font
     # info "Song" song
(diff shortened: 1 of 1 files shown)


db48x:
Neofetch with zpools listed

coletonodonnell:
This seems to report `zpool list` which in my case would yield the raw capacity of `29.1T` for one of my pools, but the actual usable capacity is found with `zfs list`. Running `zfs list` yields the usable capacity under Raidz1 as `19.8T`. I think finding a percentage of the raw capacity is less than desirable as you can never actually hit 100% in most cases as it doesn't reflect the actual usable capacity. From the [zpool manpage](https://linux.die.net/man/8/zpool):

```
     The space usage properties	report actual physical space available to the
     storage pool. The physical	space can be different from the	total amount
     of	space that any contained datasets can actually use. The	amount of
     space used	in a raidz configuration depends on the	characteristics	of the
     data being	written.  In addition, ZFS reserves some space for internal
     accounting	that the zfs(8)	command	takes into account, but	the zpool(8)
     command does not. For non-full pools of a reasonable size,	these effects
     should be invisible. For small pools, or pools that are close to being
     completely	full, these discrepancies may become more noticeable.
```

Carlosgrr:
I am here just to tell that this PR was useful to me, even if the dev decided to not merge it.

PR #1716, opened 2021-03-30
Add pkg_info listing for macOS

DesantBucie:
## Add pkg_info listing for MacOS
I have been using pkgsrc and I saw there are no packages printed on mac.
image


DesantBucie:
Nvm, when using pkgin it will be printed double...

Issue #1719, opened 2021-04-03
Git version counts slowly then reports packages available to apt (should report number of installed deb packages)


enigma9o7:
![Screenshot from 2021-04-03 12-01-27](https://user-images.githubusercontent.com/30579733/113488667-5b12ad80-9474-11eb-9033-4d99ef284f66.png)

On the left is neofetch from ubuntu focal repository.  On the right is neofetch from github today.

When I run focal version, its quick and I believe accurate packages.  When I run from github, it stops at the counting packages for about 15 seconds, then displays that huge number.

Unrelated and only noticed when I was making screenshot, it shows higher memory too.  Notice the one on the left aligns with lxtask, htop, whereas the one on the right must be including cached or something?

enigma9o7:
I figured out how to resolve this.  Change the "has apt" line to:
` has apt        && tot dpkg-query -f '.\n' -W`

PR #1720, opened 2021-04-03
Use dpkg to count installed packages for apt users

enigma9o7:
Resolves https://github.com/dylanaraps/neofetch/issues/1719


dylanaraps:
Equivalent change has been merged from elsewhere. Care to elaborate on the memory changes?

PR #1282, opened 2019-07-13
Add text alignment

zjp:
Let's try this again with a clean git log -- sorry about that. 

## Description

Adds the option to align the output to the space after the end of the separator string ":", "->", etc by padding all arguments to printf by the same amount regardless of length 

## Features
--align_output 'left', 'right', 'off'
left: Don't right-justify labels like "OS", but align what gets listed after
right: Align labels to make their contents aligned
off: Default behavior of neofetch

## Issues

1. Addresses #1241 and #895 

Screenshot 2019-07-12 at 16 46 11
Screenshot 2019-07-12 at 16 46 28
Screenshot 2019-07-12 at 16 46 49

PR #1734, opened 2021-04-20
Add alignment (replaces #1282)

zjp:
Continuation of #1282, which was accidentally closed due to a bad rebase.


dylanaraps:
I'm not pleased with this implementation (or any of the others proposed by myself or others in the past) though this is one of the better ones.

Thanks for the PR.

PR #1743, opened 2021-05-01
Add Rocky Linux

lr94:
## Description

Add Rocky Linux logo


lr94:
I just noticed that while I was submitting this pull request, another one was sent.
Maybe both arts can be accepted (mine is more similar in size to rhel and centos, but some distros have more than one version so a smaller one could be of use).

The maintainers will decide which pull request to merge

lr94:
Added small version of the logo by klebdotio

PR #1787, opened 2021-06-19
gpu: Add cache for `lspci` calls

Drarig29:
## Description

On a laptop, the first call to `lspci` can take 2 seconds when the dedicated GPU is sleeping. So for example, when I'm on battery neofetch hangs a long time when gathering the GPU and/or GPU Driver infos.

## Features

I noticed caching was only done for Mac OS X (https://github.com/dylanaraps/neofetch/pull/518) so I tried to do something similar for Linux.

I added a `cache_arr` function to cache the `gpus` array because I didn't manage to cache the `gpu_cmd` output. Everything was on a single line when I tried to `echo` or `printf` into a file.

Anyway, now with my laptop running on battery, neofetch dropped from 2 seconds to 500 ms.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -3933,8 +3933,16 @@ get_locale() {
 get_gpu_driver() {
     case $os in
         "Linux")
-            gpu_driver="$(lspci -nnk | awk -F ': ' \
-                          '/Display|3D|VGA/{nr[NR+2]}; NR in nr {printf $2 ", "; exit}')"
+            if [[ -f "${cache_dir}/neofetch/gpu_driver" ]]; then
+                source "${cache_dir}/neofetch/gpu_driver"
+
+            else
+                gpu_driver="$(lspci -nnk | awk -F ': ' \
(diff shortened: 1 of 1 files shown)


Drarig29:
Up

Drarig29:
@dylanaraps this is working since I'm using it on my fork. Why don't you consider merging this?

Do you need something else? :)

dylanaraps:
I have no plans to introduce a cache for other platforms (and would like to remove the macOS cache as well).
You can very easily introduce caching yourself (for any information) in the config file.

Thanks for the PR.

Drarig29:
@dylanaraps sorry but could you elaborate on how to cache something via the config file? I didn't find anything about that in the Wiki. Caching this information is very important to me.

dylanaraps:
The configuration file is a shell script. You can run the detection functions (`get_`) and cache them there. You also have the option to simply print hard-coded information (if these values will never change).

See also: https://github.com/dylanaraps/neofetch/wiki/Customizing-Info

PR #1790, opened 2021-06-20
Extract st terminal font from Xresources

AtifChy:
## Description
Extracts font set for st terminal from xresources. st terminal needs to patched with [the xresource patch](https://st.suckless.org/patches/xresources/)

## Features
xresources patch support
font2 patch support


neofetch
```diff
@@ -3549,16 +3549,26 @@ END
...
+                    xrdb=$(xrdb -query)
+                    term_font=$(grep -im 1 -e "^${term/d}"'\**\.*font:' -e '^\*font:' <<< "$xrdb")
+                    term_font=${term_font/*"*font:"}
+                    term_font=${term_font/*".font:"}
+                    term_font=${term_font/*"*.font:"}
+                    term_font=$(trim "$term_font")
+                fi
+
+                if [[ -z $term_font && -n "$binary" ]]; then
+                    term_font=$(strings "$binary" | tac | \
```
dylanaraps, line 3562: Why is the `tac` call needed? 

Also, it's better to use something more portable like awk: 

```sh
awk '{a[i++]=$0} END {for (j=i-1; j>=0;) print a[j--] }'
```

neofetch
```diff
@@ -3549,16 +3549,26 @@ END
                 # like a font definition. NOTE: There is a slight limitation in this approach.
                 # Technically "Font Name" is a valid font. As it doesn't specify any font options
                 # though it is hard to match it correctly amongst the rest of the noise.
-                [[ -n "$binary" ]] &&
-                    term_font=$(
-                        strings "$binary" |
+                if [[ -n $(strings "$binary" | grep resource) ]]; then
```
dylanaraps, line 3552: Can we save `strings "$binary"` to a variable so we don't call it twice in some cases?

neofetch
```diff
@@ -3549,16 +3549,26 @@ END
...
+                    xrdb=$(xrdb -query)
+                    term_font=$(grep -im 1 -e "^${term/d}"'\**\.*font:' -e '^\*font:' <<< "$xrdb")
+                    term_font=${term_font/*"*font:"}
+                    term_font=${term_font/*".font:"}
+                    term_font=${term_font/*"*.font:"}
+                    term_font=$(trim "$term_font")
+                fi
+
+                if [[ -z $term_font && -n "$binary" ]]; then
```
dylanaraps, line 3561: Why do we check if `$binary` is non-null here? (`strings` above runs with no check).

neofetch
dylanaraps, line 3552: Grep call should use `-F` as we're sending a plain string.

dylanaraps: This whole thing would be better with:

```sh
st_bin=$(strings "$binary")

if [[ $st_bin == *resource* ]]; then
    # ...
fi 

# ...

File: neofetch (hunk 3 above)
Comment (AtifChy) [reply neofetch:3561]: 3561-3568 is for font2 patch

File: neofetch [hunk 4]
```diff
@@ -3549,16 +3549,26 @@ END
...
                 # though it is hard to match it correctly amongst the rest of the noise.
-                [[ -n "$binary" ]] &&
-                    term_font=$(
-                        strings "$binary" |
+                if [[ -n $(strings "$binary" | grep resource) ]]; then
+                    xrdb=$(xrdb -query)
+                    term_font=$(grep -im 1 -e "^${term/d}"'\**\.*font:' -e '^\*font:' <<< "$xrdb")
+                    term_font=${term_font/*"*font:"}
+                    term_font=${term_font/*".font:"}
+                    term_font=${term_font/*"*.font:"}
```
dylanaraps, line 3557: Does this work to replace all of these lines?

```
term_font=${term_font/*[\*.]"font:"}
```

neofetch
```diff
@@ -3549,16 +3549,26 @@ END
                 # like a font definition. NOTE: There is a slight limitation in this approach.
                 # Technically "Font Name" is a valid font. As it doesn't specify any font options
                 # though it is hard to match it correctly amongst the rest of the noise.
-                [[ -n "$binary" ]] &&
-                    term_font=$(
-                        strings "$binary" |
+                if [[ -n $(strings "$binary" | grep resource) ]]; then
+                    xrdb=$(xrdb -query)
```
dylanaraps, line 3553: `xrdb` is only used once (following line). Why not just do:

```
term_font=$(xrdb -query | grep -im 1 -e "^${term/d}"'\**\.*font:' -e '^\*font:')
```

neofetch
dylanaraps: Yes, that's not what I asked.

PR #1795, opened 2021-06-24
Better kitty terminal font detection

kidonng:
## Description

The current way of getting kitty's terminal font won't work on v0.21.0 and later due to the removal of `--debug-config`. However, the author provided a more precise way to query the current config. More info can be read at kovidgoyal/kitty#3756.

This change is backwards compatible (and tested) with older versions of kitty.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -3428,11 +3428,9 @@ END
         ;;
 
         "kitty"*)
-            kitty_config="$(kitty --debug-config)"
-            [[ "$kitty_config" != *font_family* ]] && return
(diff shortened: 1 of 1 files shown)


redxtech:
Looks like it works perfectly! Looking forward to this, as I just switched to kitty myself.

dylanaraps:
This breaks older versions no? Can we support both?

kidonng:
This change supports (and has been  tested on) older versions of kitty.

PR #1830, opened 2021-07-18
Replace non-posix use of 'hostname' command with 'uname -n'

unah77:
## Description

This PR replaces the use of the 'hostname' (not accepted by POSIX) command with 'uname -n' (accepted by POSIX).

## Issues

The 'hostname' command is not specified by the POSIX standard, thus limiting the portability of the program.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -1376,10 +1376,10 @@ get_title() {
 
     case $title_fqdn in
         on) hostname="$(uname -n)" ;;
-        *)  hostname="${HOSTNAME:-$(
+        *)  hostname="${HOSTNAME:-"$(
                 hostname="$(uname -n)"
                 printf '%s\n' "${hostname%%.*}"
-            )}" ;;
+            )"}" ;;
     esac
 
     title=${title_color}${bold}${user}${at_color}@${title_color}${bold}${hostname}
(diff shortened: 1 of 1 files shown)


neofetch
```diff
@@ -1375,8 +1375,11 @@ get_title() {
     user=${USER:-$(id -un || printf %s "${HOME/*\/}")}
 
     case $title_fqdn in
-        on) hostname=$(hostname -f) ;;
-        *)  hostname=${HOSTNAME:-$(hostname)} ;;
+        on) hostname="$(uname -n)" ;;
+        *)  hostname="${HOSTNAME:-"$(
+                hostname="$(uname -n)"
+                printf '%s\n' "${hostname%%.*}"
+            )"}" ;;
```
dylanaraps, line 1382: You need to keep `hostname -f` as `uname -n` cannot print the FQDN. `uname -n` is fine for non-FQDN output though.

unah77:
Oh okay, `uname -n` prints the FQDN on OpenBSD so I assumed that was the standard behaviour.

dylanaraps:
See: https://serverfault.com/questions/367492/how-is-the-fqdn-determined

unah77:
Interesting, thanks!

unah77:
Readded 'hostname -f'.

PR #1837, opened 2021-07-22
Add Slax logo. Closes #1836

### LINKED ISSUE #1836: Support for os: Slax
Slax is a Live CD operating system, and while you wouldn't normally use the neofetch in an OS typically used for maintenance, all of the code is working properly (the newer versions are based on Debian) and it isn't **that** unheard of (I think). Also I personally wouldn't consider another ASCII art a bloat

This is their current logo:
https://logos-download.com/wp-content/uploads/2019/07/Slax_Logo.png

arisinfenix:
## Description

Added ascii logo for Slax as proposed in #1836

## Features
Screenshot_20210722_100201

## Issues
The logo will only show up when using `--ascii_distro slax` since they do not have a custom `/etc/os-release` file.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -10461,6 +10461,27 @@ ${c2}
 EOF
             ;;
 
+        "Slax"*)
+            set_colors 2
+            read -rd '' ascii_data <<'EOF'
+${c1}            .cMMc.
+      .,,.cMMMMMMm.      ..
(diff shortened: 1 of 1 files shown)


dylanaraps:
If there's no identification file for the distribution then it really is just Debian. 
If (or until) Slax identifies itself as Slax, I will not add support.

Thanks for the PR.

PR #1841, opened 2021-07-25
speed up get_packages by adding shortcut in `has` function

SleepyBag:
I don't know if anyone else is suffering from the same issue. On my WSL 2, it takes 4 seconds and more to run neofetch. Most of the time is consumed by `get_pacakges`.

After investigation into the code. I think it unnecessary to try reaching all the possible package managers -- it would be a large bunch of disk operation.

So I add a shortcut in the `has` function -- once the variable `$manager` is set, all the following check of `has` would return false. That is, it will just take the first package manager it sees as the answer.

This simple modification makes significant improvement on my machine:

Before modification:
```
./neofetch  0.32s user 1.11s system 33% cpu 4.298 total
```
It takes 4.298 seconds in total.

After modification:
```
./neofetch  0.27s user 0.20s system 51% cpu 0.900 total
```
It takes 0.900 seconds in total.


dylanaraps:
This breaks the whole point of why the code works the way it does, *displaying packages from all installed package managers*. 
Have you looked at a verbose log to see what is taking so much time? ie, the real issue here.

SleepyBag:
> displaying packages from all installed package managers.

It seems that I did make a mistake.

> Have you looked at a verbose log to see what is taking so much time? ie, the real issue here.

Actually, the `has` function takes most of time. On my machine, a failed  `has` check takes 0.1 seconds, which is very painful as we have so many package managers to check one by one.

I think it will be good to let users make a shortcut, or skip this feature.

dylanaraps:
Open an issue and fill out the template. I want to find out what's really causing this.

SleepyBag:
I proposed another way keeping the current design while speeding up `get_pakcage`. Created a new PR #1857

PR #1857, opened 2021-08-03
Speed up `get_packages`

SleepyBag:
## Description

Speed up the performance of `get_packages` by redesign the flow.

Currently, `get_package` runs unacceptably slow on WSL. The reason is its call to `type`. Each fail `type` call can consume 0.1s on a WSL machine. There are too many package managers to check and in total the checks consume too much time.

So, in the new flow, all the package managers are checked at once by list all executables in $PATH directories and then grep all the files. Thus we get a hit file list of possible package managers. Then each `has` call can just run `type` for package managers who show in the hit list.

On WSL, the upgraded version is much faster than the original one. 

Before modification:

```
real    0m2.513s
user    0m0.043s
sys     0m0.464s
```

After modification:

```
real    0m0.414s
user    0m0.151s
sys     0m0.053s
```

## TODO

This branch is only tested on a WSL machine. The influence on other platforms is unknown yet.


hykilpikonna:
Optimization is great! But neofetch isn't something that people run a lot, so I think compatibility is more important than performance here

Issue #1812, opened 2021-07-04
Fix Haiku OS ASCII logo


C47F15H:
## Screenshot
![image](https://user-images.githubusercontent.com/80169337/124376984-2b6b7f80-dcc7-11eb-8120-ed4107c18281.png)
Original logo ([source](https://www.haiku-os.org/images/haiku_logo_white.svg))
<img src="https://www.haiku-os.org/images/haiku_logo_white.svg" />

C47F15H:
Is anyone working on it?

dylanaraps:
You can work on it!

C47F15H:
This is what i have come up with
```
      MM         MM
      MM         MM
      MM      . cmmmm-
      MM  .cMMMLMMM'
"'---cMMMMMMMMM"*lM
      M"*:Mc`"`` MM
      MM         MM
      ""         ""
```

torculus:
@C47F15H, I made a slightly upscaled version of that which you can see here. Most distro asciis are 38-40 pixels wide
![haiku-neofetch](https://user-images.githubusercontent.com/20175597/128496727-4ccb8d68-e8b4-4a76-8a44-4158e8fd211e.png)

C47F15H:
looks nice, but can the leaf colour be replaced with light green?

torculus:
Yeah, you'd just use ascii color 10 instead of 2
Here's what it looks like
![neofetch-haiku](https://user-images.githubusercontent.com/20175597/128516693-69bd0d67-1918-4e13-b866-3ff5226e1c47.png)

C47F15H:
nice, but can you send that part of the neofetch with the c{x}'s?

torculus:
Sorry, I mixed up the two issues I'm working on. Here's the plaintext for the Haiku logo:



 ${c1}           MMMM              MMMM
            MMMM              MMMM
            MMMM              MMMM
            MMMM              MMMM
            MMMM${c2}        .**\ /**MMMM*-
 ${c1}           MMMM${c2}   .MMMMMMMMvMMMMM<
 #*         ,*MMMvMMMMMMMMMMMMMMM*
   '------cMMMMMMMMMMMMMMMMM^%.${c1}MMM
            MMMM${c2}%MMMM^MMMMM/  ${c1}MMMM
            MMMM              MMMM
            MMMM              MMMM
            """"              """"

torculus:
Sorry, I mixed up the two issues I'm working on

[...]

Issue #1757, opened 2021-05-24
New Distribution Request: Bodhi Linux


HexGremlin:
## Description

New Distribution Request: Bodhi Linux

Logo: [https://image.spreadshirtmedia.com/image-server/v1/compositions/T757A1PA1578PT10X11Y23D11654563FS9999/views/3,width=650,height=650,appearanceId=1,backgroundColor=fff/the-bodhi-linux-logo.jpg](url)

![image](https://user-images.githubusercontent.com/70352769/119332475-8a79b580-bc80-11eb-8a2d-7e155c79e3c6.png)

torculus:
If you want to skip the hard work, you're more than welcome to steal from my [repo](https://github.com/torculus/Unix-rices/tree/main/neofetch). It also includes work for illumos (general) and Genode

torculus:
@HexGremlin, here's what I came up with. What do you think? If you're all good, I can initiate a pull request to merge
![bodhi](https://user-images.githubusercontent.com/20175597/128497349-18ea2188-7a32-43fc-8d83-514e78193025.png)

dylanaraps:
That looks great. Send a PR whenever you are ready.

HexGremlin:
That looks awesome! 😁🥳

PR #1862, opened 2021-08-06
fix issue #1812

torculus:
## Description


torculus:
Sorry, this was actually in reference to issue #1757

Issue #949, opened 2018-04-11
Theme information is reported incorrectly


fearphage:
## Description

The theme information is inaccurate. However if you set the following environment variable, it displays the correct theme information.

* `GIO_EXTRA_MODULES=/usr/lib/x86_64-linux-gnu/gio/modules/`

#### Neofetch version
`3.4.0`

## Screenshot

```shell
➜  neofetch-3.4.0 ./neofetch --version | head -n 1
Neofetch 3.4.0
➜  neofetch-3.4.0 ./neofetch -vv 2> neofetch.log                        
            .-/+oossssoo+/-.               phred@honeypot
        `:+ssssssssssssssssss+:`           ----------------------------
      -+ssssssssssssssssssyyssss+-         OS: Ubuntu 17.10 x86_64
    .ossssssssssssssssssdMMMNysssso.       Host: XPS 15 9560
   /ssssssssssshdmmNNmmyNMMMMhssssss/      Kernel: 4.13.0-38-generic
  +ssssssssshmydMMMMMMMNddddyssssssss+     Uptime: 3 days, 21 hours, 1 min
 /sssssssshNMMMyhhyyyyhmNMMMNhssssssss/    Packages: 2583
.ssssssssdMMMNhsssssssssshNMMMdssssssss.   Shell: zsh 5.2
+sssshhhyNMMNyssssssssssssyNMMMysssssss+   Resolution: 3840x2160
ossyNMMMNyMMhsssssssssssssshmmmhssssssso   DE: Unity:Unity7:ubuntu
ossyNMMMNyMMhsssssssssssssshmmmhssssssso   WM: Compiz
+sssshhhyNMMNyssssssssssssyNMMMysssssss+   WM Theme: Adwaita
.ssssssssdMMMNhsssssssssshNMMMdssssssss.   Theme: Adwaita [GTK2/3]
 /sssssssshNMMMyhhyyyyhdNMMMNhssssssss/    Icons: Adwaita [GTK2/3]
  +sssssssssdmydMMMMMMMMddddyssssssss+     CPU: Intel i7-7700HQ (8) @ 3.800GHz
   /ssssssssssshdmNNNNmyNMMMMhssssss/      GPU: NVIDIA GeForce GTX 1050 Mobile
    .ossssssssssssssssssdMMMNysssso.       GPU: Intel HD Graphics 630
      -+sssssssssssssssssyyyssss+-         Memory: 31099MiB / 32018MiB
[...]

PR #1873, opened 2021-08-13
Avoid overwriting GIO_EXTRA_MODULES env var

lilyinstarlight:
## Description

The fix for #949 introduced a regression in environments where `GIO_EXTRA_MODULES` is already set (and different from the provided value of `/usr/lib/x86_64-linux-gnu/gio/modules/`). This PR checks if the variable is already set before overwriting it.

I was personally experiencing an issue with incorrect themes being reported on NixOS that seemed to be directly caused by overwriting the existing `GIO_EXTRA_MODULES` environment variable. Using this patch fixed the issue.

I'm not sure if the issue @fearphage had in #949 was due to the variable being unset entirely or set to something incorrect, so this should probably be tested on other systems (especially Ubuntu, judging by #949) to ensure there is no regression. Alternatively, just appending `/usr/lib/x86_64-linux-gnu/gio/modules/` to the end of the variable if it already exists would possibly work, but I've not tested this.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -46,7 +46,7 @@ LC_ALL=C
 LANG=C
 
 # Fix issues with gsettings.
-export GIO_EXTRA_MODULES=/usr/lib/x86_64-linux-gnu/gio/modules/
+[[ -z $GIO_EXTRA_MODULES ]] && export GIO_EXTRA_MODULES=/usr/lib/x86_64-linux-gnu/gio/modules/
 
 # Neofetch default config.
 read -rd '' config <<'EOF'


fearphage:
FYI I'm on Ubuntu 20.04 and the variable is unset.

```shell
➜  meh.gQki5Z7 if [ -z ${GIO_EXTRA_MODULES+x} ]; then echo 'variable is unset'; else echo "value is: >$GIO_EXTRA_MODULES<"; fi
variable is unset
```

lilyinstarlight:
> FYI I'm on Ubuntu 20.04 and the variable is unset.

Thank you @fearphage! If that's the case, then this change should continue working on Ubuntu with no regression.

rhoriguchi:
@dylanaraps could you please merge this. Will fix a downstream issue https://github.com/NixOS/nixpkgs/issues/108903.

hykilpikonna (approved):

hykilpikonna:

This PR is [merged into hyfetch](https://github.com/hykilpikonna/hyfetch/commit/19d15500) since this repo (dylanaraps/neofetch) seems no longer maintained.

PR #1878, opened 2021-08-17
Add support for nushell

nathom:
## Description

Currently, neofetch looks like this for nushell users

image

This fixes the issue so that it now looks like this

image

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -1740,8 +1740,8 @@ get_shell() {
         ;;
 	
         nu)
-            shell="${shell:0:2}shell "
-            shell+=$("$SHELL" --version | grep -o '\d\.\d\d\.\d[^-]' | head -n 1 )
+            shell+=$("$SHELL" -c "version | get version")
+            shell=${shell/ $shell_name}
         ;;
 
 
--- a/neofetch
+++ b/neofetch
@@ -1738,6 +1738,12 @@ get_shell() {
             shell=${shell/ Yet another shell}
             shell=${shell/Copyright*}
         ;;
+	
+        nu)
+            shell="${shell:0:2}shell "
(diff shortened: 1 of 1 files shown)


dylanaraps:
What does the full `nu --version` output look like?

kjhoerr:
I'd like to see this fixed! Although this PR doesn't seem to fix this for me either. Grep on my system doesn't match on `\d`, so it doesn't output the version.

Here is the full output:

    > nu --version
    ───┬────────────────────┬────────────────────────────────────────────────────────────────────────────────────
     # │      Column0       │                                      Column1                                       
    ───┼────────────────────┼────────────────────────────────────────────────────────────────────────────────────
     0 │ version            │ 0.37.0                                                                             
     1 │ build_os           │ macos-aarch64                                                                      
     2 │ rust_version       │ rustc 1.55.0                                                                       
     3 │ cargo_version      │ cargo 1.55.0                                                                       
     4 │ pkg_version        │ 0.37.0                                                                             
     5 │ build_time         │ 2021-09-14 19:45:30 +00:00                                                         
     6 │ build_rust_channel │ release                                                                            
     7 │ features           │ clipboard-cli, ctrlc, dataframe, default, rustyline, term, trash, uuid, which, zip 
     8 │ installed_plugins  │                                                                                    
    ───┴────────────────────┴────────────────────────────────────────────────────────────────────────────────────

It would likely be best for nu itself to process the version from the table, since it works with this format natively. It should be as simple as `nu -c 'version | get version'`. I can test or submit the change if need be.

kjhoerr:
Just to verify, the changes look great and work great for me!

dylanaraps:
Thanks!

Issue #1848, opened 2021-07-27
Cannot distinguish between WSL on Windows 10 vs Windows 11


dmadisetti:
## Description

**neofetch shows "on Windows 10" when on Windows 11**

Not sure how to do this aside from grepping Windows build numbers, but this seems like abit of a hack. Didn't see anything when searching through issues, so thought I would open here.

- [x]  Does this issue still occur in the master branch? (Required if issue)

#### Neofetch version

Neofetch 7.1.0 (newest source, top) + 2.0.2 (debian 9 stock, bottom)

#### Logs

https://gist.github.com/dmadisetti/7ec45a39e611ea4953c1029d2179c44c

## Screenshot

![image](https://user-images.githubusercontent.com/2689338/127213190-ac881bc8-7e99-4e28-a176-9b2b77c922fe.png)

dylanaraps:
What is the issue exactly?

dmadisetti:
neofetch shows "on Windows 10" when on Windows 11

dylanaraps:
Send me a verbose log.

dmadisetti:
https://gist.github.com/dmadisetti/7ec45a39e611ea4953c1029d2179c44c

dylanaraps:
Yeah. We need a way to detect windows 10 vs 11 while inside WSL.

dmadisetti:
in WSL `wmic.exe os get Version`
```
Version
10.0.22000
```

Windows 11 is still in preview mode, so this might not be worth examining until official release.

MrVeink:
Windows 11 currently acts as an Windows 10 update. Until public release, you won't most likely have a way on detecting it.

dylanaraps:
Let me know if the issue still occurs. Do not have Windows/WSL to test against.

dmadisetti:
Close enough?
<img width="491" alt="image" src="https://user-images.githubusercontent.com/2689338/144783804-04256728-6b74-45c8-b633-23a5369a4368.png">
Lower is the old script, `neofetch.1` was the newest from `wget`ing github

Logs: https://gist.github.com/dmadisetti/4153cb71b09362bbf63386117b88a0bc

dmadisetti:
Yeah, to access windows binaries `.exe` has to be specified https://gist.github.com/dmadisetti/4153cb71b09362bbf63386117b88a0bc#file-neofetch-1848-L28

[...]

PR #1818, opened 2021-07-07
Change Windows 11 logo

CatFoodEnthusiast:

PR #1889, opened 2021-08-26
Fix Windows 11 detection #1848

kipters:
## Description

Replace hardcoded "on Windows 10" with the output of `wmic os get Caption` called via PowerShell interop when running on WSL, fixing #1848 

It also correctly shows Fluent as DE when running natively on Windows 11 instead of Aero.

## Issues

It now shows "Microsoft Windows" instead of just "Windows" because that's what wmic returns.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -1102,10 +1102,11 @@ get_distro() {
             fi
 
             if [[ $(< /proc/version) == *Microsoft* || $kernel_version == *Microsoft* ]]; then
+                caption="$(powershell.exe -c wmic os get Caption | sed -n '2 p')"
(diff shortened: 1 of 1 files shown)


Jakubk15:
Also it will be good to fix the Windows 7 logo when running neofetch via scoop. 
image

Rexogamer:
(#1981 seems to have the relevant art - I'd recommend adding it or a variation of it in this PR)

hykilpikonna:
Duplicates:
#2109 (Merged into HyFetch)
#2014
#2059
#1818

PR #1906, opened 2021-09-10
added pkgin to BSD case

Zayac-The-Engineer:
## Description

Added a pkgin to Package manager programs under BSD


Zayac-The-Engineer:
On NetBSD neofetch only showed pkg_* packages with pkg_info. Now it also shows pkgin packages.

PR #1922, opened 2021-09-29
Added BugTrack asciiart

DarkCodeOrg:
## Description


hykilpikonna:
I can't find anything about a distro named BugTrack, but there is Bugtraq though. Can you provide a link to the BugTrack distro?

PR #1925, opened 2021-10-06
New OS added: CBL-Mariner

Demoodite:
## Description

I've added support for the new distro: CBL-Mariner


hykilpikonna:
CBL Mariner does have its own logo though:

image

PR #1927, opened 2021-10-08
improve/add model name for arm linux cpus

theofficialgman:
## Description

arm cpus currently fallback to the standard cpu detection

this allows to get the friendly name (Cortex A-57 for example)

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -2191,6 +2191,10 @@ get_cpu() {
                     [[ -z "$cpu" ]] && cpu="$(awk -F':' '/family/ {printf $2; exit}' "$cpu_file")"
                 ;;
 
+                "aarch64" | "armv7l" | "armv6l")
+                    cpu="$(lscpu | awk  '/Model name:/ {print $3}')"
+                ;;
+
                 *)
                     cpu="$(awk -F '\\s*: | @' \
                             '/model name|Hardware|Processor|^cpu model|chip type|^cpu type/ {


theofficialgman:
I've make this PR as well https://github.com/dylanaraps/neofetch/pull/1928, with both integrated you can see some before and after pictures:
Before:
InkedScreenshot_from_2021-10-08_12-31-31
After:
InkedScreenshot_from_2021-10-08_14-24-32

theofficialgman:
@dylanaraps anything I can do to get these changes merged?

dylanaraps:
Look at the surrounding code and how it makes use of /proc/cpuinfo. 
Does the ARM model information not exist in this file?

theofficialgman:
it does not, and thats the problem.
I did it this way because it guarantees to get the info necessary

Miepee:
Just wanted to say that this also works on Android.
Before / master:
image
After / PR:
image

Likely to be out of scope for this PR, but I've also seen an alternative way to get the CPU name on Android. OEMs can opt-in so one can view the CPU in the system settings, in which case it's possible to get it via `getprop`. However, there doesn't seem to be any standardization as to which field contains the actual CPU name and everyone seems to implement it under a different property name.

theofficialgman:
> @dylanaraps any comment on this and my other PR? #1928

@dylanaraps can this be merged? seems to be a nice change for arm cpus on linux/android. as well as my other PR fallback for integrated non-pcie gpus

hykilpikonna:
> Look at the surrounding code and how it makes use of /proc/cpuinfo. Does the ARM model information not exist in this file?

For most Android systems, the vendor patches `/proc/cpuinfo` so that it contains the processor model as plain text:

image

On these Android systems, `lscpu` will not work, showing `-` as the model name.

image

However, for Linux installed on ARM devices, `/proc/cpuinfo` is untouched, and the Hardware section will only contain the binary code for the processor, therefore it needs to be decoded with `lscpu` based on a lookup table, something like:

image

So my solution (https://github.com/hykilpikonna/hyfetch/commit/42d6aa490afa171cc92224b0f0b963e56e2afd94) is to check whether the system is Android, use `/proc/cpuinfo` if it is, and use `lscpu` if it's not.

image

Check #2139 for an updated pull request.

PR #1934, opened 2021-10-21
Add VNux GNU/Linux logo

th1nhhdk:
ảnh


th1nhhdk:
@dylanaraps Yes i know, but the logo is too complex to convert to only 6 colors

PR #1940, opened 2021-10-29
Added code to display battery percentage and resolution in termux on android devices

tathastu871:
## Description

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -2968,6 +2968,8 @@ get_resolution() {
 
                     [[ $resolution ]] && break
                 done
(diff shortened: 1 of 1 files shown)


tathastu871:
Currently code for resolution in termux requires adb to be installed on termux and connected over tcpip.
In future if termux gets a feature to create screenshot, resolution can be obtained by taking screenshot and obtaining size in pixels of that screenshot using imagemagick or other tool, because the resolution of screenshot image is always equal to screen resolution

PR #1953, opened 2021-11-10
Update copyright year and fixed POP_OS spelling mistake

Gcat101:
Self-explanatory.

POP_OS misspeling was in the "usage" function.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -8,7 +8,7 @@
 #
 # The MIT License (MIT)
 #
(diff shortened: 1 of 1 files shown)


dylanaraps:
The "spelling mistake" is there because it is the "value" given when using distribution related configuration options.
Thanks for the PR.

Gcat101:
But in distros with small logos it says "POP_OS" instead of "popos"

Issue #1952, opened 2021-11-10
Output issues with neofetch 7.1.0 on Fedora


Skaldebane:
## Description

Hi! Today I updated neofetch (through dnf), I like the new icon, but...
![Screenshot from 2021-11-10 07-08-20](https://user-images.githubusercontent.com/35589610/141059827-4de474cc-329b-4af3-9a50-5073f2719c7b.png)
What's all this spacing here?

(cmd: `neofetch --speed_shorthand on --cpu_temp C --memory_display infobar --disable packages`)

#### Neofetch version

This is neofetch 7.1.0.
But again, why do I get these errors when requesting the version?
![Screenshot from 2021-11-10 07-16-51](https://user-images.githubusercontent.com/35589610/141060264-2a1aecc3-26c9-4b35-a5d7-8ce41f0dac25.png)

## Config file

[config.conf file](https://pastebin.ubuntu.com/p/khpdc84RKk/)

## Verbose log

[neofetchlog file](https://pastebin.ubuntu.com/p/Qqy593xMKz/)

Skaldebane:
This also seems to happen differently if I run the same command normally (note that the cmd above is being run in .bashrc):
![Screenshot from 2021-11-10 07-31-08](https://user-images.githubusercontent.com/35589610/141062011-80cab169-e22a-4673-ac96-436407cf3730.png)

dylanaraps:
The errors are from running in verbose mode. These are normal.
Does the issue occur in master?

dylanaraps:
I cannot reproduce using the provided commands.

Skaldebane:
I'll try the master version and reply right away.

And also note that this doesn't happen when I run that command normally, only when it's run through .bashrc.

Skaldebane:
I installed the master version in a `toolbox` Fedora container, still the same issue.
It shows this "istrom" inside the memory indicator `infobar`, plus some line breaks (the line breaks count depends on the situation: .bashrc > normal > verbose (=0)).

Main version downloaded from dnf (through .bashrc):
![Screenshot from 2021-11-10 07-08-20](https://user-images.githubusercontent.com/35589610/141611911-37f1113b-f0ef-4975-8b75-5f27bc5a182e.png)
(Notice how the line breaks try to fit the terminal height?)

-------
Master version, inside a toolbox container:
[...]

PR #1996, opened 2021-12-16
set_text_color: fix color code for multi-digit ones

dakr:
Distributions which call set_color with a multi-digit color code end up
in the default case of the switch case statement in set_text_colors for
bar_color_total. This results in setting the color code to 'distroXY'
within the color function.

E.g. Fedroa calls 'set_color 12 7', which results in printing the color
code '\e[38;5;distrom'.

Fix this by checking whether bar_color_total equals 'distro' and in case
default to the color code in $1.

Signed-off-by: Danilo Krummrich <danilokrummrich@dk-develop.de>

## Description


hykilpikonna:
The same issue is mentioned in #1952, and is fixed by #2084 using a different method

image

PR #2051, opened 2022-02-13
Typo in comment for col_offset

thehefi:
typo in comment for col_offset

## Description


hykilpikonna:
The option `col_offset` is the offset for the color block at the end of the neofetch message so I don't think there are anything wrong with calling it "Color Offset"
image

PR #2078, opened 2022-03-23
Fixed distro detection on CentOS 6

verfriemelt-dot-org:
## Description

CentOS 6 has multiple symlinks back to /etc/centos-release

    $ ls /etc/*-release
    -rw-r--r-- 1 root root 28 Jun 26  2018 /etc/centos-release
    lrwxrwxrwx 1 root root 14 Jul  6  2018 /etc/redhat-release -> centos-release
    lrwxrwxrwx 1 root root 14 Jul  6  2018 /etc/system-release -> centos-release

which caused the distro information to be repeated 3 times. This change
will only take the first file available.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -1082,7 +1082,8 @@ get_distro() {
 
             else
                 for release_file in /etc/*-release; do
-                    distro+=$(< "$release_file")
+                    distro=$(< "$release_file")
+                    break
                 done
 
                 if [[ -z $distro ]]; then


hykilpikonna:
This would likely break distro detection on other distros... for example, in my Arch Linux, even though there are three release files, these files have different content so it's necessary to load all of them

Maybe the better solution is to transform links to their original file paths and filter out duplicates... but IDK how to do that in bash

image

PR #2095, opened 2022-04-09
Fix hostname detection for obscuring FQDN.

### LINKED ISSUE #2094: title_fqdn switch does nothing
`hostname` and `hostname -f` return the same value on most modern GNU/Linux systems with an actual FQDN applied.  Here's an example without redacting anything but my PS1.

```
$ hostname > /tmp/hostname
$ hostname -f > /tmp/full-hostname
$ diff /tmp/hostname /tmp/full-hostname 
$ 
```
As you can see, the outputs of the two commands are identical and non-empty.  The environment variable `$HOSTNAME` also evaluates to the same string.

```
$ hostname > /tmp/hostname
$ echo $HOSTNAME > /tmp/env-hostname
$ diff /tmp/hostname /tmp/env-hostname
$
```

A better way to get the short hostname and ensure no domain information leaks is to do the following:

`hostname=$(hostname | cut -d. -f1)`

That guarantees that no domain information leaks by trimming all but the leftmost element in a hostname.

tidux:
## Description
POSIX or not, the current FQDN detection doesn't work on modern Linux systems.  This fixes it as tested on FC35.

## Features
Stops leaking my hostname information.

## Issues
resolves https://github.com/dylanaraps/neofetch/issues/2094

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -1384,7 +1384,7 @@ get_title() {
 
     case $title_fqdn in
         on) hostname=$(hostname -f) ;;
-        *)  hostname=${HOSTNAME:-$(hostname)} ;;
+        *)  hostname=$(hostname | cut -d. -f1);;
     esac
 
     title=${title_color}${bold}${user}${at_color}@${title_color}${bold}${hostname}


hykilpikonna (approved):

hykilpikonna:

This PR is [merged into hyfetch](https://github.com/hykilpikonna/hyfetch/commit/09ea0c33) since this repo (dylanaraps/neofetch) seems no longer maintained.

hykilpikonna:
Just realized a problem... not all linux distros come with the `inetutils` that provides the `hostname` command. Maybe we should fall back to using `$HOSTNAME` if `hostname` command is not found.

hykilpikonna:
I changed it so that it's compatible with systems without the `hostname` command. I also replaced `cut` with parameter expansion

image

PR #2113, opened 2022-05-12
Updated Ubuntu logo

jibel:
## Description

This is an update of the Ubuntu logo to the new tag format and logo (https://design.ubuntu.com/brand/). Due to the resolution I used a square format instead of a real tag which would make the circle too small.

Here is the result with the default gnome-terminal profile:

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -11029,28 +11029,28 @@ EOF
         ;;
 
         "Ubuntu"* | "i3buntu"*)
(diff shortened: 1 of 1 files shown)


cheesepufffff:
Will it be in neofetch soon?

hykilpikonna:
We should vote between this and dylanaraps#2125. I personally think dylanaraps#2125 looks better:

image

PR #2139, opened 2022-06-25
Fix support for ARM, RISC-V, LoongArch detection

saeziae:
## Description

Use lscpu to get name of ARM CPU, tested on HiSilicon and Phytium machines.

cpuinfo shows ARM CPU implementer and model as binary information stored at cpuid register, e.g. `0x48` means HiSilicon, lscpu uses a lookup table to decode that, utilising lscpu could avoid maintaining the lookup table ourselves.

image

image

### Update 20220731

- added detection for Android, which shows CPU as string in cpuinfo.
- added Loongarch (tested on Loongson 3A5000), RISC-V (to show the uarch, tested on StarFive).

### Update 20220909

- Moved LoongArch and RISC-V to fallback detection, making it more universal.

P.S.:

small bugfix: fixed the bug of that the scripts deleting postfixes not matching those CPUs saying `with Radeon Graphics`

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -2206,6 +2206,10 @@ get_cpu() {
                     [[ -z "$cpu" ]] && cpu="$(awk -F':' '/family/ {printf $2; exit}' "$cpu_file")"
                 ;;
 
+                "arm"* | "aarch64")
+                    cpu="$(lscpu | grep "Vendor ID" | sed "s/.*: *//g") $(lscpu | grep "Model name" | sed "s/.*: *//g")"
+                ;;
+		
                 *)
                     cpu="$(awk -F '\\s*: | @' \
                             '/model name|Hardware|Processor|^cpu model|chip type|^cpu type/ {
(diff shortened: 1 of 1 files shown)


Icenowy:
图片
On a big.LITTLE machine (RK3399)

The CPU frequency is also wrong (it shows only the little core freq, not the big core one)

hykilpikonna:
圈子好小 qwq

hykilpikonna:
合并到 [HyFetch](https://github.com/hykilpikonna/HyFetch) 里面啦。会加入到下一个版本的 `neowofetch` 里面w

（看 neofetch 半年没接过 PR 了我要借机 take over！

satmandu:
This is working on a Raspberry Pi.

image

hykilpikonna:
Found a bug: This method fails to detect cpu for riscv on QEMU.

It can be solved by moving the fallback cpu detection outside the switch-case block. (i.e. if no cpu is detected with platform-specific methods, try using the `/proc/cpuinfo` method)

image

aaronfranke:
What is the status of this PR? I would like to make a PR to improve neofetch on RISC-V, but I don't want to conflict with this work.

PR #2152, opened 2022-07-13
created much cleaner void ascii art

rejahtavi:
## Description
The existing Void logo looks very messy, perhaps it was auto-generated from an image.
I made a much cleaner version by hand.


hykilpikonna:
This looks cool

PR #2154, opened 2022-07-15
Added Support for Uwuntu

fish-in-the-sea:
## Description

## Features
Added Support for Uwuntu


hykilpikonna:
uwu

hykilpikonna (approved):

eruizc-dev:
Now that [UwUntu](https://uwuntuos.site/) is a real OS, it would be great to merge this!

I've tested it and it even works by default but it paints next prompt white for some reason, do you have any idea why this could be?

image

eruizc-dev:
NVM, there was an issue with my terminal or the way I installed it

image

eruizc-dev (approved):

PR #2166, opened 2022-07-27
Added some Termux support(Detect termux and detect fonts in use by termux )

dontknowhy:
## Description

Add Termux detection by using $TERMUX_VERSION
Add Termux's font detection by reading font file in /data/data/com.termux/files/home/.termux/font.ttf

## Features
now supports Termux
## TODO
I don't know if I can detect the rows and columns of the terminal in Termux.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -3598,6 +3598,9 @@ END
                          "$termite_config")"
         ;;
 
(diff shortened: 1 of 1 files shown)


dontknowhy:
(

PR #2150, opened 2022-07-05
Add macOS 13 to version name list

kidonng:
## Description

Reference: https://www.apple.com/macos/macos-ventura-preview/

PR #2178, opened 2022-08-15
Show macOS Ventura name

bomberfish:
## Description

Show the full name of macOS Ventura instead of just "macOS 13.0"


hykilpikonna:
This is a duplicate of #2150

PR #2187, opened 2022-08-27
changed Windows 11 logo

yusufkaanocak:
did not liked the current windows 11 ascii logo and changed it to a better version in my opinion.


Ubuntufanboy:
This should be merged. Looks good to me

hykilpikonna:
Maybe we should have a vote on this, I personally feel like the original logo looks cleaner.

Or another option is to add this ascii art as a variant, so the default wouldn't be changed but people can use this variant by specifying `ascii_distro` in either the command line or the config.

**Original**:

image

**This branch**:

image

slackingfred:
Yeah, the original Windows 11 flag must be the dullest ASCII art ever...

PR #2212, opened 2022-10-10
add show_package_total option

PauloGDPeixoto:
## Description

I found that the package count was degrading the performance of neofetch in my machines, so I added a parameter to optionally disable this particular output.

## Features
Add --show_package_total flag to enable/disable the package count display in neofetch.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -197,6 +197,13 @@ memory_unit="mib"
 # off:  '908'
 package_managers="on"
 
+# Show/Hide total package count
+# Default: 'true'
+# Values:  'true', 'false'
+# Flag:    --show_package_total
+#
+show_package_total=true
+
 
 # Shell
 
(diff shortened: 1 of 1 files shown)


jinasanami:
Wow, it works well on Univalent GNU/Linux!

jinasanami:
Oops, I forgot to attach the image what I turned off the option.
VirtualBox_plasma_13_10_2022_12_45_19

neofetch
```diff
@@ -1507,203 +1514,205 @@ get_uptime() {
...
-    if ((packages == 0)); then
-        unset packages
-
-    elif [[ $package_managers == on ]]; then
-        printf -v packages '%s, ' "${managers[@]}"
-        packages=${packages%,*}
-
-    elif [[ $package_managers == tiny ]]; then
-        packages+=" (${manager_string%,*})"
+    if [[ $show_package_total == "on" ]]; then
```
hykilpikonna, line 1517: It would be much better to return early if `$show_package_total == "off"` rather than re-indent the entire function

hykilpikonna:
This feature already exists in neofetch. You can use `--disable <infoname>` to disable the detection of a specific item. In this case, you can use `--disable packages`

image

You can also turn off package manager detection by editing the `~/.config/neofetch/config.conf` (specifically, comment out the `info "Packages" packages` line)

PR #2225, opened 2022-10-31
Enhancing --memory_unit functionality

TriMoon:
## Description

Improving https://github.com/dylanaraps/neofetch/commit/0435dcd0cd44bb22afa9b986f15742cc05de7b20

## Features

- Implementing https://github.com/dylanaraps/neofetch/issues/1170#issuecomment-455576889 to avoid awk usage.
- Adding ability to configure precision of output using `mem_precision` which defaults to `2`.
- Added `tib` to accommodate TiB mentioned in https://github.com/dylanaraps/neofetch/issues/1170#issuecomment-894705941

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -172,7 +172,7 @@ memory_percent="off"
 # Change memory output unit.
 #
 # Default: 'mib'
-# Values:  'kib', 'mib', 'gib'
+# Values:  'kib', 'mib', 'gib', 'tib'
 # Flag:    --memory_unit
 #
 # Example:
--- a/neofetch
+++ b/neofetch
@@ -2812,22 +2812,22 @@ get_memory() {
             mem_label=TiB
             memory_unit_divider=$((1024 * 1024))
             printf -v mem_used "%'.*f" \
-                        ${mem_precision} \
-                        $(($mem_used / $memory_unit_divider)).$(($mem_used % $memory_unit_divider))
+                        "${mem_precision}" \
+                        $((mem_used / memory_unit_divider)).$((mem_used % memory_unit_divider))
             printf -v mem_total "%'.*f" \
-                        ${mem_precision} \
-                        $(($mem_total / $memory_unit_divider)).$(($mem_total % $memory_unit_divider))
+                        "${mem_precision}" \
+                        $((mem_total / memory_unit_divider)).$((mem_total % memory_unit_divider))
         ;;
 
         gib)
             mem_label=GiB
             memory_unit_divider=1024
             printf -v mem_used "%'.*f" \
-                        ${mem_precision} \
-                        $(($mem_used / $memory_unit_divider)).$(($mem_used % $memory_unit_divider))
+                        "${mem_precision}" \
+                        $((mem_used / memory_unit_divider)).$((mem_used % memory_unit_divider))
             printf -v mem_total "%'.*f" \
-                        ${mem_precision} \
-                        $(($mem_total / $memory_unit_divider)).$(($mem_total % $memory_unit_divider))
+                        "${mem_precision}" \
+                        $((mem_total / memory_unit_divider)).$((mem_total % memory_unit_divider))
         ;;
 
         kib)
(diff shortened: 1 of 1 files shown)


hykilpikonna (approved):

hykilpikonna:
There are a lot of shellcheck warnings...

image

TriMoon:
> There are a lot of shellcheck warnings...

I'm editing online, will check/address those warnings later :wink:

hykilpikonna:
> I'm editing online, will check/address those warnings later 😉

I just fixed it for you:

```bash
    case $memory_unit in
        tib)
            mem_label=TiB
            memory_unit_divider=$((1024 * 1024))
        ;;

        gib)
            mem_label=GiB
            memory_unit_divider=1024
        ;;

        kib)
            mem_used=$((mem_used * 1024))
            mem_total=$((mem_total * 1024))
            mem_label=KiB
        ;;
    esac

    case $memory_unit in
        tib|gib)
            printf -v mem_used "%'.*f" \
                        "${mem_precision:-2}" \
                        "$((mem_used / memory_unit_divider)).$((mem_used % memory_unit_divider))"
            printf -v mem_total "%'.*f" \
                        "${mem_precision:-2}" \
                        "$((mem_total / memory_unit_divider)).$((mem_total % memory_unit_divider))"
        ;;
    esac
```

But I can't push to your branch though

TriMoon:
Hahaha that's a way to do it also ofcourse, by creating an extra block....
Will think about it, let me first finish my addition of the precision documentation :wink: 

I know i should have started with a "Draft" pull request but ahh well heheheh

hykilpikonna:
Looking into this, I actually found another problem, which is that memory bars only work when the memory unit is MiB (the bar function takes in MiB values even though they have been converted).

Can be fixed by keeping track of an original mb value:

image

TriMoon:
@hykilpikonna, i just noticed the current non-patched version does wrong arithmetic in it's attempt to convert `KB` into `MiB`, https://github.com/dylanaraps/neofetch/blob/0435dcd0cd44bb22afa9b986f15742cc05de7b20/neofetch#L2538-L2649

It does integer calculation instead of floating point...
fe.  `1000 KB` should convert to `0.9765625 MiB`, but current calculation does `$((1000 / 1024)) = 0` in multiple places in above function.

I already know i will use the same kind of `printf` coding i used in here to do the necessary floating-point calculations in bash, probably will make it a function...

So I'm going to first correct that before going further, as it will change more stuff.
_(in due time when i get more time again)_

TriMoon:
> Looking into this, I actually found another problem, which is that memory bars only work when the memory unit is MiB (the bar function takes in MiB values even though they have been converted).
> 
> Can be fixed by keeping track of an original mb value:

That won't be needed when we keep the original values in `Bytes` so we can display in more units as MiB-only like the current code does...

TriMoon:
> @hykilpikonna  But I can't push to your branch though

Invited as collaborator :wink:

hykilpikonna:
> Invited as collaborator 😉

Thanks! I just pushed my commit fixing the memory progress bar issue

> So I'm going to first correct that before going further, as it will change more stuff. (in due time when i get more time again)

Yea, I think creating a function for floating-point calculations would be nice, and we can keep the raw value in bytes or kib to reduce the number of calculations and conversions. Let me know when it's ready to merge!

TriMoon:
> @hykilpikonna Let me know when it's ready to merge!

These changes, with your fix could be merged first, because the floating point calculation functionality would be something completely different and extra to this MR :wink: 
Will create a new one especially for that in due time.

hykilpikonna:
Okay, just [merged into hyfetch](https://github.com/hykilpikonna/hyfetch/commit/cde4914c)!

> [HyFetch](https://github.com/hykilpikonna/hyfetch) is a fork of neofetch with LGBTQ pride flags, but the repo also maintains an updated version of the original neofetch, addressing many pull requests that are not merged in the original repo.
>
> Read the ["Running Updated Original Neofetch" section](https://github.com/hykilpikonna/hyfetch#running-updated-original-neofetch) for more info!

PR #2084, opened 2022-03-28
Fixes #1952

skaldebane:
## Description
This pull request replaces the old broken code (see #1952) with a simpler, working one for `bar_color_total`, just like `bar_color_elapsed`. Let me know if the old code was done that way for a specific purpose.

## Issues
Fixes #1952

PR #2227, opened 2022-10-31
Fixes the broken progress bars that result from particular distro colors

### LINKED ISSUE #2226: progress bars for disk and memory usage are broken
## Description

The first part of the bar is printed, followed by zero or more blank lines, followed by the string “istrom”, the rest of the bar and the rest of the info.

- [✓] Does this issue still occur in the master branch? (**Required if issue**)

#### Neofetch version

Neofetch 7.1.0

## Screenshot

Screenshot from 2022-10-31 03-02-27

## Config file

Omitted; it’s nothing unusual.

## Verbose log

I’ll save you the effort. Here’s the relevant bit:

```
+ case ${bar_color_total}${1} in
++ color distro
++ case $1 in
++ printf '\e[38;5;%bm' distro
+ bar_color_total='\033[38;5;distrom'
```

You can see that it is substituting the string “distro” into the escape sequence for the color. It was supposed to pick a number, but failed. It turns out that the new Fedora logo uses color 12, which is two digits. The patterns in `set_text_colors` only match single–digit numbers, so presumably the same bug will happen on a fair number of distros.

db48x:
A distro like Feodra that uses a two–digit color wouldn’t match either of the first two patterns, and would end up trying to use the string “distro” as a color value.

Fixes #2226.

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -4717,7 +4717,7 @@ set_text_colors() {
 
     case ${bar_color_total}${1} in
         distro[736]) bar_color_total=$(color "$1") ;;
-        distro[0-9]) bar_color_total=$(color "$2") ;;
+        distro*)     bar_color_total=$(color "$2") ;;
         *)           bar_color_total=$(color "$bar_color_total") ;;
     esac
 }


hykilpikonna:
Duplicate of https://github.com/dylanaraps/neofetch/pull/2084

Thanks anyways.

PR #2231, opened 2022-11-07
Proof-of-concept device tree realization

### LINKED ISSUE #2230: [Feature Request] Parsing 'device tree' instead of 'cpuinfo' if present
## Description

Some devices and emulators do not display the processor model in `/proc/cpuinfo`, but have it in `/proc/device-tree/model`/`/sys/firmware/devicetree/base/model` or in compiled `/sys/firmware/fdt`, so it would be appropriate to parse it instead of displaying an empty line.

More information about device tree:
* https://www.devicetree.org/specifications/
* https://www.kernel.org/doc/Documentation/devicetree/
* https://elinux.org/Device_Tree_Usage
* https://github.com/jviki/dtree

- [x] Does this issue still occur in the master branch? (**Required if issue**)

000exploit:
## Description
Not intended for daily use on every machine, but now it can not only parse `cpuinfo` in Linux.

## Features
* Device tree
## Issues
Fixes #2230

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -2189,29 +2189,36 @@ get_wm_theme() {
 get_cpu() {
     case $os in
         "Linux" | "MINIX" | "Windows")
-            # Get CPU name.
+            dt="/proc/device-tree"
             cpu_file="/proc/cpuinfo"
 
-            case $kernel_machine in
-                "frv" | "hppa" | "m68k" | "openrisc" | "or"* | "powerpc" | "ppc"* | "sparc"*)
-                    cpu="$(awk -F':' '/^cpu\t|^CPU/ {printf $2; exit}' "$cpu_file")"
-                ;;
-
-                "s390"*)
-                    cpu="$(awk -F'=' '/machine/ {print $4; exit}' "$cpu_file")"
-                ;;
-
-                "ia64" | "m32r")
-                    cpu="$(awk -F':' '/model/ {print $2; exit}' "$cpu_file")"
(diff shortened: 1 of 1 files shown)


neofetch
```diff
@@ -2189,29 +2189,36 @@ get_wm_theme() {
...
+                   ;;
+    
+                   *)
+                       cpu="$(awk -F '\\s*: | @' \
+                               '/model name|Hardware|Processor|^cpu model|chip type|^cpu type/ {
+                               cpu=$2; if ($1 == "Hardware") exit } END { print cpu }' "$cpu_file")"
+                   ;;
+               esac
+        else
+                cpu="$(cat $dt/model)"
```
hykilpikonna, line 2220: Weird... `/proc/device-tree` doesn't exist for 5/6 of my devices on very new kernels, and for one that does exist, the model file returns the name of the SoC `OrangePi 3 LTS` instead of the CPU name. What's the requirement for `/proc/device-tree` to exist?

image

image

image

image

image

image

000exploit: > What's the requirement for `/proc/device-tree` to exist?

The main requirement is to have device tree, of course. It's missing on most x86(_64) machines, and also not all device trees have the processor name in the model property, but some Android devices have. Also, for some RISC-V devices Linux kernel can use the `compatible` property as processor model in the `uarch` entry in `/proc/cpuinfo`.

I marked this PR as a draft, because it's just PoC, and we need more research for that.

000exploit: In addition, it's possible to parse `compatible` property and show processor codename in kernel, but it's not always true and in some cases will not work. For example, `compatible` property contains specific components for different SoCs or only one entry without the SoC codename and so.

PR #2243, opened 2022-11-27
Fix "with Radeon Graphics" filter not working on some APUs

ara-mariandre:
## Description

On my machine (Fedora 37; AMD Ryzen 5 4600G) the "with Radeon Graphics" bit is not properly filtered out since there is nothing between the words Radeon and Graphics in its cpuinfo


hykilpikonna (approved):

hykilpikonna:
This is already fixed in https://github.com/dylanaraps/neofetch/pull/2139. Thanks anyways

PR #2260, opened 2022-12-10
packaging: Make it available through the baz plugin manager

ar1ja:
https://github.com/TruncatedDinosour/baz

Signed-off-by: Ari Archer <ari.web.xyz@gmail.com>

## Description

## Features

- Package `neofetch` for the `baz` plugin manager for BASH

### DIFF
--- a/baz.env
+++ b/baz.env
@@ -1,0 +1,8 @@
+export BAZP_NAME='neofetch'
+export BAZP_LICENSE='MIT'
+export BAZP_VER=710
+export BAZP_AUTHOR='Dylan Araps'
+export BAZP_SRC='.'
+export BAZP_METHOD='git'
+export BAZP_SOURCE='https://github.com/dylanaraps/neofetch'
+export BAZP_DESCRIPTION='A command-line system information tool written in bash 3.2+'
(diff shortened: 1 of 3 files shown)


ar1ja:
Note: You *can* move the `hooks` dir to some other nested dir by changing `BAZ_SRC`, for example if `BAZ_SRC=hello`, you can move `hooks` to `hello/hooks`

Note: To test you can `baz install git https://ari-web.xyz/gh/neofetch`, will install this pull's fork

Note: If you merge it would be fun if you added the `baz-plugin` tag to neofetch :) but more than optional, just nice, sorry if this seems rude

Note: The shebangs in hooks dont matter, they're `source` hooks, meaning theyll get `source`d, i just changed them for consistency

Note: Fun fact: `neofetch` is also installable with [basher](https://www.basher.it/)

hykilpikonna:
It might be better to separate packaging scripts from the source repo... 

1. If every package manager needs some special files added to the source repo, then the source repo would be filled with files that are irrelevant to the code or functionality of the project. 
2. If you need to include package scripts, you cannot package something without the approval of the developer (to edit/merge into the source repo)

It might be better to keep every package script in one repo that you control, like [Homebrew](https://github.com/Homebrew/brew) and [nixpkgs](https://github.com/NixOS/nixpkgs)

ar1ja:
> 

:+1:

PR #2273, opened 2022-12-31
Fall back to host name if model name is not found.

tfpf:
## Description
On Debian GNU/Linux 11 (bullseye) on WSL, the device model is not shown, since none of the files `get_model` searches for are present. One might consider falling back to the host name in that case.


hykilpikonna:
But the hostname doesn't always contain model information though... Also, the hostname is already displayed beside the username, so I don't think it would make sense to repeat it in the model field

tfpf:
Didn't think of that.:sweat_smile: I always thought it is the same as the device name.

PR #2276, opened 2023-01-04
🔧 [fix] battery bar elapsed maths incorrect #2275

### LINKED ISSUE #2275: battery bar elapsed maths incorrect
## Description

Battery at 94% charged (so 6% used). barinfo shows 94% elapsed when it should show 6% elapsed.

line 3844 to 3846 should say `$((100-${battery/\%*}))` instead of `${battery/\%*}`

- [x] Does this issue still occur in the master branch? (**Required if issue**)

#### Neofetch version
7.1.0

## Screenshot
incorrect actual behaviour:
actual

expected behaviour:
expected

## Config file
[config.conf](https://github.com/dylanaraps/neofetch/files/10342975/config.conf.txt)

## Verbose log
[neofetchlog](https://github.com/dylanaraps/neofetch/files/10342959/neofetchlog.txt)

1. Run `neofetch -vv 2> neofetchlog`
2. Upload the contents of `neofetchlog` to pastebin, gist or equivalent.

mightbesimon:
Description

Battery at 94% charged (so 6% used). barinfo shows 94% elapsed when it should show 6% elapsed. line 3844 to 3846 should say `$((100-${battery/\%*}))` instead of `${battery/\%*}`

incorrect actual behaviour:
actual

expected behaviour:
expected

## Issues
fixes: #2275

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -3841,9 +3841,9 @@ get_battery() {
     [[ "$battery_state" ]] && battery+=" Charging"
 
     case $battery_display in
-        "bar")     battery="$(bar "${battery/\%*}" 100)" ;;
-        "infobar") battery="${battery} $(bar "${battery/\%*}" 100)" ;;
-        "barinfo") battery="$(bar "${battery/\%*}" 100)${info_color} ${battery}" ;;
+        "bar")     battery="$(bar "$((100-${battery/\%*}))" 100)" ;;
+        "infobar") battery="${battery} $(bar "$((100-${battery/\%*}))" 100)" ;;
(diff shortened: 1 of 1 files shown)


mightbesimon:
This PR fixes #2276

hykilpikonna (approved):

hykilpikonna:
Maybe it would be better to make this an option of how it should be displayed, like `--battery_bar_used`. 

With this pr, the remaining battery would be "draining" from left to right, which is kind of counterintuitive but also matches with the CPU and RAM progress bars (i.e. the green bar is the remaining). So I guess there is a valid argument on both sides of the choice.

PR #2313, opened 2023-04-18
Added new distro Astra Linux

js-denton:
## Description
Added support new disto Astra Linux

## Features
* Support Fly DE
* Improvment detect host model
HP ProDesk 490 G3 MT Business PC - bette than - HP 805F
ASUSTeK COMPUTER INC. X201EP - better than X201EP
* Little refactoring
image

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -28,7 +28,7 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 
-version=7.1.0
+version=7.1.1
 
 # Fallback to a value of '5' for shells which support bash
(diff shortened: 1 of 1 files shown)


neofetch
```diff
@@ -28,7 +28,7 @@
 # OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 # SOFTWARE.
 
-version=7.1.0
+version=7.1.1
```
hykilpikonna, line 31: Please don't change the version here. Otherwise everything looks good!

hykilpikonna (approved):

neofetch
js-denton: Thanks, I'll take it into account next time

TitaniumHocker (approved):

PR #2335, opened 2023-06-26
Update neofetch

MoonBurst:
adding the -v flag to the wm check, fixing the bug where swayidle and swaybg commonly used in other systems read as sway being the running WM.


hellidox:
did you test this? this will not work.

MoonBurst:
> did you test this? this will not work.

Yeah, it removed swayidle from being listed as Sway WM for me.

PR #2377, opened 2023-09-08
Fixed #2376

### LINKED ISSUE #2376: Add ANSI 8-bit Support
## Description

I'm doing an ascii art project, and I noticed that it would be awesome to have a profile picture in Neofetch, but there's a problem: for some reason, Neofetch treats 8-bit colors as actual characters. <br>
This means that if I have, for instance, `^[[38;231;0;255m hi mom` it will trhow the neofetch system information about 12 characters to the right, instead of the 1 expected. <br>

I know this is hard af to explain, and my poorly written text doesn't help, so I have an image to give as an example:
image

As you can see the information went far away, this is because it sees the ansi color as a char. For example, in the first line, we have
[38;2;111;68;85m╏[38;2;132;95;113m0[38;2;101;75;104m╏ <br>

I have no idea how hard it is to fix this issue, but it shouldn't be that hard, as I was kind of able to fix it with simple tweaks.
image

<b>Now, you can argue that if I'm using chars to make colored art, why don't I use images in the first place? <br>
And if you say that, I just gotta tell you to stop, this way of thinking hurts my feelings and f u</b>

Furlanetto-Dev:
## Description
Fixed #2376

## Issues
I know I shouldn't be using echo or sed, but that's just easier

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -4069,7 +4069,7 @@ print_ascii() {
     # Calculate size of ascii file in line length / line count.
     while IFS=$'\n' read -r line; do
         line=${line//\\\\/\\}
-        line=${line//█/ }
+        line=$(echo -e "$line" | sed -E 's/█//g; s/\x1B\[[0-9;]*[a-zA-Z]//g')
         ((++lines,${#line}>ascii_len)) && ascii_len="${#line}"
     done <<< "${ascii_data//\$\{??\}}"
 


hykilpikonna:
Thanks. Though this is already fixed in https://github.com/dylanaraps/neofetch/pull/1543

Furlanetto-Dev:
> I have figured out the solution and have updated my code accordingly and have cleaned it up. I have tested both with the pasted character in the ascii file and the escape code and both work. This is also much faster than my initial sed implementation. I have updated the OP to reflect these changes as well as clarify a few things.



> Thanks. Though this is already fixed in #1543

dude so why the fuck do didn't the neofetch devs  make it to main

Furlanetto-Dev:
I'm sorry if I was rude, but for real, why it wasn't been patched yet????

PR #2385, opened 2023-09-27
Fix LMDE ASCII art.

vndmtrx:
## Description

Fix the first line of the LMDE Ascii Art, which is not aligned with the rest of the logo.


hykilpikonna:
Thanks, but this has already been addressed in https://github.com/dylanaraps/neofetch/pull/2063

PR #2395, opened 2023-10-12
Resolution: actually support wayland protocol on Linux

CarterLi:
Format `${scaled_width}x${scaled_height} @${scale_factor}x @ ${refresh_rate} Hz` is used to match the format of macOS

Note: current implementation requires `wayland-info`, which is usually provided by package [`wayland-utils`](https://gitlab.freedesktop.org/wayland/wayland-utils)

Ref: https://github.com/hykilpikonna/hyfetch/issues/196#issuecomment-1757046939

image

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -3079,7 +3079,34 @@ get_resolution() {
         ;;
 
         *)
-            if type -p xrandr >/dev/null && [[ $DISPLAY && -z $WAYLAND_DISPLAY ]]; then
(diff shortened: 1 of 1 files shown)


CarterLi:
While the issue in Windows may be fixable too, it doesn't worth the effert, because neofetch in Windows is fundamentally broken. Try [fastfetch](https://github.com/ScoopInstaller/Main/blob/master/bucket/fastfetch.json) instead.

hykilpikonna (approved):

PR #2397, opened 2023-10-12
Resolution: support multi-display resolutions detection on Windows

CarterLi:
Format: `${scaled_width}x${scaled_height} @${scale_factor}x`

Refresh rates are missing. It's possible but I don't bother to do it. For full functional and 100x faster detection, try fastfetch instead.


CarterLi:
image

PR #2416, opened 2023-11-26
EshanizedOS

eshanized:
## Description

## Issues
Making my own distro


neofetch
```diff
@@ -820,7 +820,7 @@ image_source="auto"
 #       Slackware, SunOS, LinuxLite, OpenSUSE, Raspbian,
 #       postmarketOS, and Void have a smaller logo variant.
 #       Use '{distro name}_small' to use the small variants.
-ascii_distro="auto"
+ascii_distro="EshanizedOS"
```
hykilpikonna, line 823: Please don't change the default ascii_distro in the default config. If you change this to EshanizedOS, all distros will show up as EshanizedOS, which is not good.

If you just want to test it, you can use the `--ascii_distro` argument

neofetch
```diff
@@ -7346,6 +7346,41 @@ dMm     `/++/-``/yNNh+/sdNMNddMm-    mMd
       /dMMh+.              .+hMMd/
         -odMMNhso//////oshNMMdo-
            `:+yhmNMMMMNmhy+:`
+EOF
+        ;;
+
+        "EshanizedOS"*)
+            set_colors 1 7
+            read -rd '' ascii_data <<'EOF'
+${c1}            .:-==++++++++++++++++++++++-.
```
hykilpikonna, line 7355: Please pad with spaces to make the first line align.

image_2023-12-01_14-40-55

hykilpikonna (changes requested):

PR #2419, opened 2023-12-07
Resolution: don't display disabled resolutions

CarterLi:
Before:

image

After:

image

This issue affects both tty and wayland.

Note this method reports preferred / maximum resolution instead of current resolution as I said before. I highly recommand https://github.com/dylanaraps/neofetch/pull/2395 if you use wayland


CarterLi:
AD: fastfetch uses [libdrm](https://gitlab.freedesktop.org/mesa/drm) to fetch current resolution and refresh rate, which is not available in `/sys/class/drm`

image

PR #2449, opened 2024-03-04
WIP: attempt to detect DE on remote sessions

mmuman:
## Description

When connecting with ssh to a machine that has an X11 session running, try to find which DISPLAY we can query.

## Features

It only detects the DE for now.

## Issues

The `KDE_SESSION_VERSION` property doesn't seem to exist on my session, so I had to check another one which doesn't mention the version.

## TODO

Reuse the found `DISPLAY` to detect other components?

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -1833,7 +1833,7 @@ get_de() {
             # the desktop variables are sometimes also set to the
             # window manager name. This checks to see if WM == DE
             # and discards the DE value.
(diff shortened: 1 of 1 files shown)


mmuman:
I noticed [this video](https://www.youtube.com/watch?v=rZFyi-BsdZA) mentioning neofetch didn't detect desktops remotely via ssh, so I had a try at it.

Issue #2454, opened 2024-03-15
[Feature] Display device MAC Address


trymeouteh:
Output the device MAC Address when using the _neofetch_ in the terminal

thomastubby:
Just submitted pull request with this feature included.

PR #2464, opened 2024-04-07
added MAC custom info

thomastubby:
## Description

## Features

added function that will pull and display MAC address, mentioned in issue #2454

### DIFF
--- a/neofetch
+++ b/neofetch
@@ -3923,6 +3923,11 @@ get_public_ip() {
     fi
 }
 
+get_hwaddr() {
+    hwaddr="$(ip route get 1 | awk '{print $5}')"
+    hwaddr="$(ip link show dev "$hwaddr" | awk '/ether/ {print $2}')"
+}
+
 get_users() {
     users="$(who | awk '!seen[$1]++ {printf $1 ", "}')"
     users="${users%\,*}"


neofetch
```diff
@@ -3923,6 +3923,11 @@ get_public_ip() {
     fi
 }
 
+get_hwaddr() {
+    hwaddr="$(ip route get 1 | awk '{print $5}')"
+    hwaddr="$(ip link show dev "$hwaddr" | awk '/ether/ {print $2}')"
```
waterkip, line 3928: A random comment/nit.

You can nest `$()` blocks, eg:

```
hwaddr="$(ip link show dev "$(ip route get 1 | awk '{print $5}')" | awk '/ether/ {print $2}')"
```
This code surprised me a bit. My initial thought was that you would override the previous value.

thomastubby: I have followed previously set conventions for this project. Please see the 'get_local_ip' function.

waterkip: They do different things there. They do bashisms to remove some parts of the string.  When they do something similar to what you are doing, they do this:

```
interface="$(route get 1 | awk -F': ' '/interface/ {printf $2; exit}')"
local_ip="$(ipconfig getifaddr "$interface")"
```

thomastubby: Sure, I could make different variables but why would I when this works? Its easy to read and it works well enough. I do appreciate the insights though and will try to incorporate them in future feature implementations.

